PR #3, opened 2023-01-27
Add HA redis cluster CFN

lgarvey:
No. of replica nodes, engine and instance size configurable per env.
Automatic failover
TLS transport
Cluster can be scaled down to 1 node for non prod environments

Remaining:
incorporate into bootstrap script


shehzadashiq (approved):
LGTM

PR #4, opened 2023-02-10
Add env level addons command

lgarvey:
Secret formats standardised
Where possible full connection strings are stored in SSM, e.g rediss://theurl:6379 instead of just "theurl"
TLS everywhere
t-shirt sizes specified in storage-plans.yaml


shehzadashiq (approved):
LGTM

PR #13, opened 2023-04-21
Modify bootstrap generation and add generated config

lgarvey:
The bootstrap script has been modified to reflect changes we have made to the paas. These are:

1) remove the command - the docker start command is handled in the procfile/paketo 
2) remove backing services - these are now managed in storage.yaml at env level 
3) change bootstrap image path
4) remove logic to retain prod domains and replace non prod domains copilot equivalent 
5) remove domain element - we're now retaining the non prod domains


ejayesh (approved):
lgtm

PR #16, opened 2023-04-27
Bugfix: Allow all YAML extensions (`yml` and `yaml`) in `get-ci-conf.py`

yusufsheiqh:

PR #17, opened 2023-04-28
Update migration tools

yusufsheiqh:
## Context

- Updates all scripts found in the `migration-tools` directory to be folder agnostic; can be run from anywhere.
  - Each script will now automatically generate the `*-config.yml` files as output.
- Updated documentation; understand what each script does and what are the requirements.
- Update all configuration files in `bootstrap-config` directory to match GOV.UK PaaS data more accurately.
  - Add in missing configuration files; see https://github.com/uktrade/copilot-tools/pull/16 for reference.
- All file extensions now follow the agreed `*.yml` format.


migration-tools/main.py
```diff
@@ -1,3 +1,5 @@
+# Todo: I think we can just delete this file?
```
WillGibson, line 1: What do you think @yusufsheiqh?

yusufsheiqh: This can be deleted. I wonder if this was created to combine the 4 scripts into one where we only need to run this `main.py`. Could be something we can look into for a future PR.

WillGibson: Yes, would be good to wrap it all up into a single command, but whilst it's not doing anything useful, it's just cruft 🗑️

WillGibson: Also, wrapping into one command maybe not worth the effort as this stuff will be binned once we are all off Gov.UK PaaS.

WillGibson (approved):

bootstrap-config/data-workspace-apps-copilot.yml
```diff
@@ -55,47 +54,8 @@ services:
...
-    DATAWORKSPACE_BASE_URL: DATAWORKSPACE_BASE_URL
-    EXPORT_WINS_DATASET_ID: EXPORT_WINS_DATASET_ID
-    EXPORT_WINS_SOURCE_TABLE_ID: EXPORT_WINS_SOURCE_TABLE_ID
-    FLASK_ENV: FLASK_ENV
-    HAWK_CLIENT_ID: HAWK_CLIENT_ID
-    HAWK_CLIENT_KEY: HAWK_CLIENT_KEY
-    PAGINATION_SIZE: PAGINATION_SIZE
-    RUN_SCHEDULER: RUN_SCHEDULER
-    STUB_SOURCE_DATA: STUB_SOURCE_DATA
+  secrets: {}
```
lgarvey, line 58: I'm wondering why quite a few config files are losing their secrets?  🤔

bootstrap-config/data-workspace-apps-copilot.yml
```diff
@@ -113,26 +73,8 @@ services:
...
-    DATAFLOW_BASE_URL: DATAFLOW_BASE_URL
-    DATAFLOW_HAWK_ID: DATAFLOW_HAWK_ID
-    DATAFLOW_HAWK_KEY: DATAFLOW_HAWK_KEY
-    DSS_PAGINATION_SIZE: DSS_PAGINATION_SIZE
-    FLASK_ENV: FLASK_ENV
-    INPUTS_SOURCE_FOLDER: INPUTS_SOURCE_FOLDER
-    S3_BUCKET_URL: S3_BUCKET_URL
-    SENTRY_DSN: SENTRY_DSN
-    SENTRY_ENVIRONMENT: SENTRY_ENVIRONMENT
+  secrets: {}
```
lgarvey, line 77: Odd 🤔

lgarvey (commented):
This looks good.  The only query is re's the secrets in quite a few config files being replaced with 

```
secret: {}
```

migration-tools/main.py
lgarvey: Yeh, that's my feeling about these scripts.  We could optimise them heavily, but ultimately is it worth it?

bootstrap-config/data-workspace-apps-copilot.yml
yusufsheiqh: Good catch. Updated the `build-copilot-config.py` script accordingly in https://github.com/uktrade/copilot-tools/pull/17/commits/91e0a56b5f458479b6ebee36de430ee34bd792b5.

bootstrap-config/data-workspace-apps-copilot.yml
yusufsheiqh: Resolved in https://github.com/uktrade/copilot-tools/pull/17/commits/8d8a41c3b591670596c29d2cd72ee4a66b4dffec.

PR #19, opened 2023-05-11
GPM-472 Add Paketo `buildspec` configuration

yusufsheiqh:
## Context

- Add `buildspec` configuration file for CodeBuild project.
- Automates pulling Docker image from DockerHub to push to public ECR.

### Question(s)

Opinions/thoughts on putting the `buildspec` file in a `paketo` directory rather than in the root directory?

### DIFF
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -1,0 +1,33 @@
+version: 0.2
+
+phases:
+  install:
+    runtime-versions:
+        docker: 18
+    commands:
+      - nohup /usr/local/bin/dockerd --host=unix:///var/run/docker.sock --host=tcp://127.0.0.1:2375 --storage-driver=overlay2&
+      - timeout 15 sh -c "until docker info; do echo .; sleep 1; done"
+  pre_build:
+    commands:
+    - echo Logging in to AWS ECR....
+    - aws --version
+    - $(aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/uktrade)
+    - IMAGE_TAG=$(curl -s 'https://hub.docker.com/v2/repositories/paketobuildpacks/builder/tags?name=-full&page_size=1' | jq -r '.results[].name')
+    - REPOSITORY_URI=public.ecr.aws/uktrade/paketobuildpacks/builder
+  build:
+    commands:
+    - echo Pull started on `date`
+    - echo Pulling the Docker image...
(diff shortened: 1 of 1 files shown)


ejayesh (commented):
Looks good, just need to include run and lifecycle images too

buildspec.yml
```diff
@@ -0,0 +1,27 @@
...
+phases:
+  install:
+    commands:
+      - nohup /usr/local/bin/dockerd --host=unix:///var/run/docker.sock --host=tcp://127.0.0.1:2375 --storage-driver=overlay2&
+      - timeout 15 sh -c "until docker info; do echo .; sleep 1; done"
+  pre_build:
+    commands:
+    - echo Logging in to AWS ECR....
+    - aws --version
+    - aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/uktrade
```
WillGibson, line 12: Can we set up an IAM role giving CodeBuild the required access instead of needing to log in?

If yes, I would be happy to see this happen in a follow up pull request.

ejayesh: @WillGibson you still need a role to give you access, but you need to use the get-login command to authenticate.

ejayesh (approved):
lgtm.  approved

buildspec.yml
```diff
@@ -0,0 +1,36 @@
...
+    - BUILDER_IMAGE_TAG=$(curl -s 'https://hub.docker.com/v2/repositories/paketobuildpacks/builder/tags?name=-full&page_size=1' | jq -r '.results[].name')
```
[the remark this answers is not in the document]
yusufsheiqh: Nope - `page_size` is set to `1` which means that there will only ever be a single result.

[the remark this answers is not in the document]
yusufsheiqh: > page_size | integer Number of items to get per page. Defaults to 10. Max of 100.

More info here: https://docs.docker.com/docker-hub/api/latest/#tag/repositories/paths/~1v2~1namespaces~1%7Bnamespace%7D~1repositories~1%7Brepository%7D~1tags/get

buildspec.yml
WillGibson: I don't think I understand how logging in to AWS ECR helps us with Docker Hub rate limiting, maybe our AWS ECR has a paid Docker Hub subscription or something?

WillGibson (approved):

PR #23, opened 2023-05-23
Update health check success codes for redirect status codes

yusufsheiqh:
## Context

- Allow `301` and `302` success codes as well for health check in the `manifest.yml` file for a service, e.g. `web`.


ejayesh (approved):
approved

PR #21, opened 2023-05-22
Add support for s3

lgarvey:
This PR adds support for s3 bucket creation, and granting services access to existing s3 buckets using an IAM role policy.

The schema validation will be fixed in a follow up PR.

The `storage.yml` structure has changed slightly to make it more flexible.  Environments are now listed under an environments key.  This makes it easy to config keys that allow across envs, e.g 

```
mybucket:
    type: s3

    readonly: true   # applies to all environments

    services:
        - web
        - s3proxy

    environments:
        default:
            plan: small

       prod: 
           plan: medium-ha
```

The default key remains under environments.

Further PRs will improve validation/schema checking and add some unit tests.

PR #24, opened 2023-05-23
Updated volume-size key for opensearch and rename s3-policy

shehzadashiq:
Volumesize key does not work with copilot-bootstrap so renaming to volume-size in storage-plans.yaml
Rename external-s3 to s3-policy

### DIFF
--- a/storage-plans.yaml
+++ b/storage-plans.yaml
@@ -64,14 +64,14 @@ redis:
 opensearch:
   # 1 dedicated VM, 1 CPU per VM, 4GB RAM per VM, 80GB disk space.
   tiny:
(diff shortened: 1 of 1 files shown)


storage-plans.yaml
```diff
@@ -186,4 +186,4 @@ aurora-pg: {}
 
 s3: {}
 
-external-s3: {}
+s3-policy: {}
```
yusufsheiqh, line 189: This change cannot be merged till https://github.com/uktrade/copilot-tools/pull/21 is merged. Should this change be made in a subsequent PR or do we wait for the linked PR to merged first?

PR #26, opened 2023-05-25
Remove r53 records

ejayesh:
This change is for when we have multiple environments.

### DIFF
--- a/domain-bootstrap.py
+++ b/domain-bootstrap.py
@@ -14,7 +14,6 @@ from schema import Optional, Schema, SchemaError
 
 # To do
 # -----
-# Need to change this to a class to make it more cleaner (maybe)
 # Need to seperate Dev and Prod in this app - related to prod domain not being found if specified dev profile for the domain
 # Check user has logged into the aws accounts before scanning the accounts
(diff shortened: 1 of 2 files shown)


domain-bootstrap.py
```diff
@@ -417,10 +419,11 @@ def assign_domain(app, domain_profile, svc):
     vars = result_json['variables']
 
     for var in vars:
-        if var['name'] == 'COPILOT_LB_DNS':
+        if var['environment'] == env and var['name'] == 'COPILOT_LB_DNS':
```
yusufsheiqh, line 422: Might be worth making sure we are pulling this value for/from the correct container.

```suggestion
        if var['environment'] == env and var['name'] == 'COPILOT_LB_DNS' and var['container'] == svc:
```

ejayesh: good pick up.  LB can only be attached to a web svc, but when there are multiple services, the copliot command will grab the non web services in its list.

I will add this too.

yusufsheiqh: Thank you kindly. Resolved in https://github.com/uktrade/copilot-tools/pull/26/commits/83ef14001f02b1e2b4fd947c652a500a147d3272.

PR #27, opened 2023-05-25
Improve storage command

lgarvey:
Do deeper validation of the storage.yml doc using jsonschema
Add additional checks that aren't easy to do in jsonschema - namely validating envs and svcs exist
Rename generate-storage to make-cloudformation in prep for handling WAF and s3 IP filter policies 
Initially breaking up the make-cloudformation function to move towards more easily testable code
Generated some unit tests to validate the json schema logic. 
Move to pip-tools


ejayesh (approved):
lgtm

copilot-bootstrap.py
```diff
@@ -409,76 +391,144 @@ def instructions(config_file):
...
+
+    return normalised_config
+
+
 @cli.command()
 @click.argument("storage-config-file", type=click.Path(exists=True))
 @click.argument("output", type=click.Path(exists=True), default=".")
 @click.option('--overwrite', is_flag=True, show_default=True, default=True, help='Overwrite existing cloudformation? Defaults to True')
-def generate_storage(storage_config_file, output, overwrite):
+def make_cloudformation(storage_config_file, output, overwrite):
```
yusufsheiqh, line 468: Should we be specific about CloudFormation whether it's a stack, template, etc?

yusufsheiqh (approved):

copilot-bootstrap.py
lgarvey: Good question. I wonder what is the most descriptive from a developer perspective? We expect them to run this command anytime they make changes to `storage.yaml`, and anytime they add a new service or environment. 

'make-config' seems a bit too vague

@WillGibson what do you think?

PR #28, opened 2023-05-30
Update template instructions

yusufsheiqh:
## Context

- Update CLI output for storage instructions.
- Remove redundant instructions from CLI output when generating Copilot configuration files (`make_config`).
  - Not used and conflates CLI output.
- Add link to GitHub docs for CLI output in `make_config` method.

### DIFF
--- a/copilot-bootstrap.py
+++ b/copilot-bootstrap.py
@@ -94,12 +94,11 @@ storage_schema = Schema({
 
 
 def _mkdir(base, path):
-
     if (base / path).exists():
-        return f"directory {path} exists; doing nothing"
+        return f"Directory {path} exists; doing nothing"
 
     (base / path).mkdir(parents=True)
-    return f"directory {path} created"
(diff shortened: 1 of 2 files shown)


shehzadashiq (commented):
It looks good to me however I don't agree with the part where the instructions that were previously being rendered being replaced solely by the GitHub link.
 You can keep both and say latest documentation is on the GitHub page.

shehzadashiq:
> > It looks good to me however I don't agree with the part where the instructions that were previously being rendered being replaced solely by the GitHub link. You can keep both and say latest documentation is on the GitHub page.
> 
> @shehzadashiq Why not? The main reason to remove this was to reduce conflation of CLI output as the GitHub documentation was being followed already, so moar instructions from the CLI output seems to be unnecessary and also outdated.

In my opinion it is easier to see what is on the command line sometimes instead of having to open another window. You can put the caveat that the latest documentation is on GitHub.

PR #29, opened 2023-05-31
refactor and exception handling for dns_cli.py

ejayesh:
No unit tests added yet.
Refactored and added exception handling to dns_cli.py.  (Will do same to codebuild_cli.py in seperate PR)


dns_cli.py
```diff
@@ -265,224 +226,286 @@ def create_hosted_zone(client, domain, start_domain, base_domain):
...
-    # breakpoint()
-    # pp.pprint(response)
 
     # add records to hosted zone to validate certificate
+    cert_arn = create_cert(acm_client, domain_client, domain, base_len)
 
     return cert_arn
 
 
+def check_aws_conn(aws_profile):
```
yusufsheiqh, line 278: This is nice - we definitely needed this.

dns_cli.py
```diff
@@ -265,224 +226,286 @@ def create_hosted_zone(client, domain, start_domain, base_domain):
...
+    alias_client = session.client('iam')
+    account_name = alias_client.list_account_aliases()['AccountAliases']
+    print(f"Logged in with AWS Domain account: {account_name[0]}/{sts.get_caller_identity()['Account']}\n\
+          User: {sts.get_caller_identity()['UserId']}")
+    return session
+
+
+def check_response(response):
+    if response['ResponseMetadata']['HTTPStatusCode'] != 200:
+        print("Uknown response error from AWS")
```
yusufsheiqh, line 313: Minor typo here.

```suggestion
        print("Unknown response error from AWS")
```

dns_cli.py
```diff
@@ -265,224 +226,286 @@ def create_hosted_zone(client, domain, start_domain, base_domain):
...
 
         parts.pop(0)
 
     else:
-        print(f"No hosted zone found for {domain}")
+        print(f"No hosted zone found for {domain_name}")
         return
 
 
+def run_atest():
```
yusufsheiqh, line 505: Better to move tests to their own directory and/or own files.

yusufsheiqh (approved):

dns_cli.py
ejayesh: I forgot to take out run_atest.  Will remove

dns_cli.py
ejayesh: good point, I will add the status code

dns_cli.py
yusufsheiqh: Thank you! Resolved in https://github.com/uktrade/copilot-tools/pull/29/commits/47f03694754c5665e60347821bc51297092e9e7d.

lgarvey (approved):
Looks good.  

Only minor thing, we should probably use click.echo instead of print for consistency.  The nice thing is we can apply colour formatting.  Not essential, so feel free to ignore for now.

https://click.palletsprojects.com/en/7.x/utils/#ansi-colors

PR #30, opened 2023-05-31
add rds templates

gabelton:
Created svc and env postgres rds addon template, taking [this answer](https://github.com/aws/copilot-cli/issues/1178#issuecomment-963408231) as basis and combining with parts of our existing aurora postgres template.

### DIFF
--- a/copilot-bootstrap.py
+++ b/copilot-bootstrap.py
@@ -42,7 +42,7 @@ config_schema = Schema({
             Optional("backing-services"): [
                 {
                     "name": str,
-                    "type": lambda s: s in ("s3", "external-s3", "postgres", "redis", "opensearch",),
+                    "type": lambda s: s in ("s3", "external-s3", "postgres", "postgres-rds", "redis", "opensearch",),
                     Optional("paas-description"): str,
                     Optional("paas-instance"): str,
                     Optional("notes"): str,
@@ -64,7 +64,7 @@ config_schema = Schema({
 
 storage_schema = Schema({
     str: {
-         "type": lambda s: s in ("s3", "external-s3", "postgres", "redis", "opensearch",),
+         "type": lambda s: s in ("s3", "external-s3", "postgres", "postgres-rds", "redis", "opensearch",),
          str: {
             Optional("plan"): str,
             # s3
@@ -148,6 +148,7 @@ def setup_templates():
             "public-manifest": templateEnv.get_template("svc/manifest-public.yml"),
             "backend-manifest": templateEnv.get_template("svc/manifest-backend.yml"),
             "opensearch": templateEnv.get_template("svc/addons/opensearch.yml"),
+            "postgres-rds": templateEnv.get_template("svc/addons/postgres-rds.yml"),
             "postgres": templateEnv.get_template("svc/addons/postgres.yml"),
             "redis": templateEnv.get_template("svc/addons/redis.yml"),
             "s3": templateEnv.get_template("svc/addons/s3.yml"),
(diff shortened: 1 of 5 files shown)


copilot-bootstrap.py
```diff
@@ -48,7 +48,7 @@
             Optional("backing-services"): [
                 {
                     "name": str,
-                    "type": lambda s: s in ("s3", "external-s3", "postgres", "redis", "opensearch",),
+                    "type": lambda s: s in ("s3", "external-s3", "aurora-pg", "rds-pg", "redis", "opensearch",),
```
yusufsheiqh, line 51: I think this is not up to date with the `main` branch, is it? I believe we now use `s3-policy` rather than `external-s3`.

copilot-bootstrap.py
```diff
@@ -129,16 +129,16 @@ def setup_templates():
             "public-manifest": templateEnv.get_template("svc/manifest-public.yml"),
             "backend-manifest": templateEnv.get_template("svc/manifest-backend.yml"),
             "opensearch": templateEnv.get_template("svc/addons/opensearch.yml"),
-            "postgres": templateEnv.get_template("svc/addons/postgres.yml"),
+            "rds-pg": templateEnv.get_template("svc/addons/rds-pg.yml"),
```
yusufsheiqh, line 132: We will still need the other `aurora` Postgres template.

```suggestion
            "aurora-pg": templateEnv.get_template("svc/addons/aurora-pg.yml"),
            "rds-pg": templateEnv.get_template("svc/addons/rds-pg.yml"),
```

yusufsheiqh (commented):
I think it might be worthwhile to also add a `storage-example.yml` file which can be referenced in the Platform Documentation.

lgarvey:
@yusufsheiqh Adding a `storage-example.yaml` file is a great idea. I'll create a PR and add it now.

lgarvey (approved):
Looks good!

yusufsheiqh (approved):
Great work! Let's get this merged. 🚀

buildspec-tests.yml
```diff
@@ -0,0 +1,18 @@
...
+      - pip install -r requirements-test.txt
+  
+  build:
+    commands:
+      - echo Running Tests
+      - pytest
+  
+  post_build:
+    commands:
+      - echo Tests Complete
```
WillGibson, line 18: Is there any actual value in this post build command?

copilot-bootstrap.py
```diff
@@ -48,7 +48,7 @@
             Optional("backing-services"): [
                 {
                     "name": str,
-                    "type": lambda s: s in ("s3", "external-s3", "postgres", "redis", "opensearch",),
+                    "type": lambda s: s in ("s3", "s3-policy", "aurora-pg", "rds-pg", "redis", "opensearch",),
```
WillGibson, line 51: To reduce potential for ambiguity/cognitive load, is there any reason we cannot use the full "postgres" word in these instead of "pg"?

templates/svc/addons/rds-pg.yml
```diff
@@ -0,0 +1,105 @@
+Parameters:
```
WillGibson, line 1: I am fully confused about when we do a service addon vs when we do an environment addon. 

@lgarvey Am I correct in thinking that this database will be destroyed if we delete the service?

lgarvey: A service level addon is bound to the service, and will be nuked along with the service.  When we started work we only had service level addons.  Then they introduced env level addons.  

There's actually no facility to add a service level add on with `storage.yml` except for the s3 & s3-policy types which have a `services` key to add an identity policy in the service's addons directory.
 
We could remove most of the service level stuff to avoid confusion.

buildspec-tests.yml
yusufsheiqh: Probably not, since CodeBuild has its own `STATUS` anyway.

templates/svc/addons/rds-pg.yml
yusufsheiqh: Aren't these templates very similar? Can we make them moar generic and not separate them out into `env` and `svc` directories?

Is there a difference in the templates that I am missing?

WillGibson: There's definitely something to do like that @yusufsheiqh . As well as the duplication angle, I think we want to be guiding people away from service addons to avoid exasperation when people accidentally delete a data store or something.

PR #33, opened 2023-06-01
Add example storage file

lgarvey:


storage-example.yaml
yusufsheiqh, line 1: I believe we are going with `.yml` as the convention instead of `.yaml`.

yusufsheiqh (approved):
Nice! If the filename could be renamed as per my other comment, that would be great.

storage-example.yaml
lgarvey: Ah, of course!

gabelton (approved):

storage-example.yaml
yusufsheiqh: Resolved in https://github.com/uktrade/copilot-tools/pull/33/commits/ad8cbacd21c169184cd9c4f41fc242ade2547187.

storage-example.yml
```diff
@@ -0,0 +1,115 @@
...
+
+      plan: some-other-plan  # Any of the keys defined in a storage plan can also be overridden
+      volume-size: 200
+
+
+##### s3 bucket example:
+
+my-s3-bucket:
+  # type: s3 will create a bucket in each environment.
+  # Services listed in the `services` key will be granted access to the bucket.
```
WillGibson, line 26: I think these two are not needed as covered in the comments below.
```suggestion
```

storage-example.yml
```diff
@@ -0,0 +1,115 @@
...
+  type: s3   # creates an s3 bucket in each environment and gives the listed services permissions to access the bucket
+
+  readonly: true  # services are granted read only access to the bucket
+
+  services:  # services that require access to the bucket.
+    - "web"
+    - "web-celery"
+
+  environments:
+    # NOTE: bucket names must be unique and need to be defined for each environment
```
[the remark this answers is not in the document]
yusufsheiqh: This is a `.yml` file so not sure if it will parse markdown.

storage-example.yml
```diff
@@ -0,0 +1,115 @@
...
+
+
+##### s3 bucket example:
+
+my-s3-bucket:
+  # type: s3 will create a bucket in each environment.
+  # Services listed in the `services` key will be granted access to the bucket.
+  # currently no other features are supported such as bucket versioning, etc.  If you need additional features, you can use the s3-policy type to give a service access to an existing bucket.
+
+  type: s3   # creates an s3 bucket in each environment and gives the listed services permissions to access the bucket
```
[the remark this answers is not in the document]
lgarvey: It's a combination of an env level and a service level add on.  The bucket creation stuff is added to the environment, then an associated policy needs adding to the service addon to give the service access to the bucket.

The s3-policy option skips the env level s3 bucket creation and just adds the iam policy to the service addon.  We'll get situations where people say they need exotic bucket configs, and we could let them do this outside of copilot and then use the s3-policy option to grant access to the bucket.

storage-example.yml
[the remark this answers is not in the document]
WillGibson: Oh, of course!

storage-example.yml
```diff
@@ -0,0 +1,113 @@
...
+
+  readonly: true  # services are granted read only access to the bucket
+
+  services:  # services that require access to the bucket.
+    - "web"
+    - "web-celery"
+
+  environments:
+    # NOTE: [bucket names must be unique across all AWS accounts in the AWS (Standard Regions) Partition](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html) and need to be defined for each environment
```
WillGibson, line 36: @yusufsheiqh made a good point about this not being markdown. Sorry...
```suggestion
    # NOTE: bucket names must be unique across all AWS accounts in the AWS (Standard Regions) Partition (see https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html) and need to be defined for each environment
```

Will approve now anyway as this is not a deal breaker.

WillGibson (approved):

PR #37, opened 2023-06-02
Add `pre-commit` hooks

yusufsheiqh:
## Context

- Add `pre-commit` hooks:
  - `black` as code formatter.
  - `pre-commit-hooks` for replacing line endings and detecting private keys. 
  - `pytest` for running tests.
- Apply initial `pre-commit` changes for all files.
  - The `pre-commit` hooks are only run on the files that are changed.

## Additional information

JIRA ticket: https://uktrade.atlassian.net/browse/GPM-534

### DIFF
--- a/commands/codebuild_cli.py
+++ b/commands/codebuild_cli.py
@@ -10,12 +10,8 @@ DEFAULT_CI_BUILDER = "public.ecr.aws/uktrade/ci-image-builder"
 
 
 def import_pat(pat, client):
-
     response = client.import_source_credentials(
-        token=pat,
-        serverType='GITHUB',
-        authType='PERSONAL_ACCESS_TOKEN',
-        shouldOverwrite=True
+        token=pat, serverType="GITHUB", authType="PERSONAL_ACCESS_TOKEN", shouldOverwrite=True
     )
     check_response(response)
     print("PAT successfully added")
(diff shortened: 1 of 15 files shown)


commands/codebuild_cli.py
```diff
@@ -10,12 +10,8 @@
 
 
 def import_pat(pat, client):
-
     response = client.import_source_credentials(
-        token=pat,
-        serverType='GITHUB',
-        authType='PERSONAL_ACCESS_TOKEN',
-        shouldOverwrite=True
+        token=pat, serverType="GITHUB", authType="PERSONAL_ACCESS_TOKEN", shouldOverwrite=True
```
WillGibson, line 14: What if we wanted it on multiple lines because we find that easier to grok? It will just get automatically put onto one line?

yusufsheiqh: You can add the comment `# fmt: off` at the top of the block and then turn the formatter back on at the end of the block with `# fmt: on`.

```suggestion
        # fmt: off
        token=pat,
        serverType="GITHUB",
        authType="PERSONAL_ACCESS_TOKEN",
        shouldOverwrite=True
        # fmt: on
```

yusufsheiqh: Here's a direct link to the docs: https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#code-style

lgarvey (approved):
Smashing 👍 🚀

lgarvey (approved):
Smashing 👍 🚀

commands/codebuild_cli.py
yusufsheiqh: Yes, I think so. But, if you really want to keep it on multiple lines and avoid the `fmt: on/off` comments, then you need to append a trailing comma, and it will not automatically put it onto one line. The docs also recommend this approach.

```suggestion
        token=pat,
        serverType="GITHUB",
        authType="PERSONAL_ACCESS_TOKEN",
        shouldOverwrite=True,
```

Reference: https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#trailing-commas

PR #38, opened 2023-06-02
Update `buildspec` configuration

yusufsheiqh:
## Context

- Refactor: Remove `post_build` phase in `buildspec` configuration.
- Upgrade Python runtime version in `buildspec` configuration.
- Rename `requirements-test` -> `dev`.
- Add `requirements` folder.

## Additional information

JIRA ticket - https://uktrade.atlassian.net/browse/GPM-538


buildspec-tests.yml
```diff
@@ -3,16 +3,12 @@ version: 0.2
 phases:
   install:
     runtime-versions:
-      python: 3.9
+      python: 3.11
     commands:
       - echo Installing dependencies...
-      - pip install -r requirements-test.txt
+      - pip install -r requirements/dev.txt
```
gabelton, line 9: What's the thinking behind the rename?

gabelton (approved):

buildspec-tests.yml
yusufsheiqh: These requirements are mainly for developers and isn't only for testing purposes. The `pre-commit` hook is a dev dependency, for example.

lgarvey: I suggest we move to poetry anyway. What do you think?

yusufsheiqh: Yes - we have a JIRA ticket to address `poetry`: https://uktrade.atlassian.net/browse/GPM-530

PR #44, opened 2023-06-06
Add WAF config

lgarvey:
The WAF acl arns live in each `copilot/environment/{name}/manifest.yml` file Includes tests

TODO: look at chaining commands so that there's a single entrypoint, e.g `make-cloudformation`


yusufsheiqh (approved):
Great work! 🚀

commands/copilot_cli.py
```diff
@@ -9,11 +9,21 @@
...
+from .utils import camel_case, ensure_cwd_is_repo_root, mkdir, mkfile, SSM_BASE_PATH, setup_templates
 
 
 BASE_DIR = Path(__file__).parent.parent
 
+WAF_ACL_ARN_KEY = "waf-acl-arn"
+
+
+def list_copilot_local_environments():
+    return [path.parent.parts[-1] for path in Path("./copilot/environments/").glob("*/manifest.yml")]
```
WillGibson, line 21: I think I just learnt how to return the last element of an array in Python 😄

commands/copilot_cli.py
```diff
@@ -9,11 +9,21 @@
...
+
+
+def list_copilot_local_services():
+    return [path.parent.parts[-1] for path in Path("./copilot/").glob("*/manifest.yml")]
```
WillGibson, line 25: Would this include things like `environments` and `pipelines`?

WillGibson: Ah, maybe not, I think this would return every directory in `copilot` which has a direct descendant called `manifest.yml`?

commands/utils.py
```diff
@@ -105,7 +106,16 @@ def setup_templates():
             "aurora-postgres": templateEnv.get_template("addons/env/aurora-postgres.yml"),
             "redis": templateEnv.get_template("addons/env/redis-cluster.yml"),
             "s3": templateEnv.get_template("addons/env/s3.yml"),
+            "waf": templateEnv.get_template("env/addons/waf.yml"),
+            "parameters": templateEnv.get_template("env/addons/addons.parameters.yml"),
```
WillGibson, line 110: I think these paths are wrong?

```suggestion
            "waf": templateEnv.get_template("addons/env/waf.yml"),
            "parameters": templateEnv.get_template("addons/env/addons.parameters.yml"),
```

tests/conftest.py
```diff
@@ -0,0 +1,12 @@
+from pathlib import Path
+
+import pytest
+
+
+@pytest.fixture
+def fakefs(fs):
+    """Mock file system fixure with the templates and schemas dirs retained"""
```
WillGibson, line 8:
```suggestion
    """Mock file system fixture with the templates and schemas dirs retained"""
```

tests/test_copilot_cli.py
```diff
@@ -0,0 +1,72 @@
...
+from click.testing import CliRunner
+import pytest
+import yaml
+
+from commands.copilot_cli import copilot as cli, WAF_ACL_ARN_KEY
+
+
+class TestApplyWAF:
+
+    def test_require_copilot_dir(self, fakefs):
```
WillGibson, line 12: Could we use more meanigful names for the test methods so that it's clearer what the intent of the test is and thereby easier to be sure it's verifying the correct thing?

E.g.
```suggestion
    def test_apply_waf_with_no_copiliot_directory_throws_error(self, fakefs):
```

commands/copilot_cli.py
yusufsheiqh: @WillGibson Yes, that is correct. Those directories that contain the `manifest.yml` file which are returned are the Copilot services.

commands/utils.py
lgarvey: Nah they are correct. WAF applies to the ALB which is environment level.  I moved the templates back to this directory.

tests/test_copilot_cli.py
lgarvey: The test are fairly self explanatory, but I'm happy make them even more descriptive.

The `apply_waf` bit is evident in the containing class name.

```
class TestApplyWaf:
```

To make it clearer I'll rename it to 

```
class TestApplyWafCommand:
```

Maybe a better name for this method then would be `test_exit_if_no_copilot_directory()` ?

commands/utils.py
WillGibson: I think you got `env` and `addons` the wrong way round though?

templates/addons/svc/waf.yml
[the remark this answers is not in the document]
yusufsheiqh: Resolved in https://github.com/uktrade/copilot-tools/pull/44/commits/6f6749b53fde6f309a212e5c14f92c5d02aea181.

PR #45, opened 2023-06-07
DBTP-198 Enable Autoscaling

WillGibson:
I think these changes should replicate the default we use in GOV.UK PaaS as closely as we can get with AWS Copilot and Elastic Container Service.

Jira ticket: https://uktrade.atlassian.net/browse/DBTP-178


lgarvey (approved):
Looks good and seems fairly straight forward.

yusufsheiqh (approved):

PR #46, opened 2023-06-07
DBTP-105 Create unit tests for codebuild_cli.py

gabelton:
https://uktrade.atlassian.net/browse/DBTP-105

Adding unit tests and type hints for codebuild_cli.py commands.

N.B. `codedeploy` test is commented out pending bug investigation in [this ticket](https://uktrade.atlassian.net/browse/DBTP-184)


WillGibson:
Agree with @yusufsheiqh's comment. Might be worth pulling in the main branch, it's got to happen sometime and it might be the fix for the builds, you never know ;-)

tests/test_codebuild_cli.py
```diff
@@ -0,0 +1,204 @@
...
+def alias_session(aws_credentials):
+    with mock_iam():
+        session = boto3.session.Session()
+        session.client("iam").create_account_alias(AccountAlias="foo")
+        
+        yield session
+
+
+# Not much value in testing these while moto doesn't support `import_source_credentials`` or `list_source_credentials`
+def test_import_pat():
```
lgarvey, line 37: Does something need doing with these empty functions?

utils/aws.py
```diff
@@ -3,7 +3,7 @@
 import botocore
 
 
-def check_aws_conn(aws_profile):
```
lgarvey, line 6: We have utils in two places.  It'll be worth consolidating them in a future PR.

lgarvey (approved):
Looks good!

yusufsheiqh (approved):

tests/test_codebuild_cli.py
```diff
@@ -0,0 +1,204 @@
...
+# Not much value in testing these while moto doesn't support `import_source_credentials`` or `list_source_credentials`
+# def test_import_pat():
+#     ...
+# def test_check_github_conn():
+#     ...
+
+
+@mock_iam
+def test_check_service_role_does_not_exist(capfd, aws_credentials):
+    """Test that check_service_role outputs expects error method when codebuild role does not exist."""
```
WillGibson, line 45: Can you explain the purpose of this string?

It looks to me like it might be kind of a comment, but done as a string which is not output.

tests/test_codebuild_cli.py
```diff
@@ -0,0 +1,204 @@
...
+
+# Not much value in testing these while moto doesn't support `import_source_credentials`` or `list_source_credentials`
+# def test_import_pat():
+#     ...
+# def test_check_github_conn():
+#     ...
+
+
+@mock_iam
+def test_check_service_role_does_not_exist(capfd, aws_credentials):
```
WillGibson, line 44: In line with the conversation @lgarvey and I are having over here: https://github.com/uktrade/copilot-tools/pull/44#discussion_r1221373977

I wonder if these test methods could be named so as to better express what's actually being tested?

E.g.
```suggestion
def test_check_service_role_throws_error_when_codebuild_role_does_not_exist(capfd, aws_credentials):
```

gabelton: I've tried to use docstrings for this purpose, while keeping the test names a bit terser because people will be typing or copypasting these and it's useful to keep them shortish in that context

tests/test_codebuild_cli.py
gabelton: I'm using docstrings to make it more explicit what each test is actually testing. This one unhelpfully contains a typo, however, making it less useful...

yusufsheiqh: This is known as a Docstring. Moar info here: https://peps.python.org/pep-0257/#what-is-a-docstring

commands/codebuild_cli.py
```diff
@@ -3,13 +3,15 @@
...
+from mypy_boto3_codebuild.client import CodeBuildClient
+
 from utils.aws import check_response, check_aws_conn
 
 AWS_REGION = "eu-west-2"
 DEFAULT_CI_BUILDER = "public.ecr.aws/uktrade/ci-image-builder"
 
-
-def import_pat(pat, client):
+def import_pat(pat: str, client: CodeBuildClient):
```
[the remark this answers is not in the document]
gabelton: Relatively recent feature, but yeah they're cool. Using them with boto is a bit more fiddly. Hence the new package

tests/test_codebuild_cli.py
WillGibson: I prefer more explicit method names for tests, then the output from the test run, especially if there are failures, is more helpful.

PR #48, opened 2023-06-07
Add ip filter command

lgarvey:


commands/copilot_cli.py
```diff
@@ -119,14 +133,20 @@ def _normalise_keys(source: dict):
     default=True,
     help="Overwrite existing cloudformation? Defaults to True",
 )
-def make_cloudformation(storage_config_file, output, overwrite):
-    """Generate storage cloudformation for each environment."""
+def make_storage(storage_config_file, output, overwrite):
```
yusufsheiqh, line 136: Are we going back with `make_storage` now? I thought this was changed to `make_cloudformation`.

shehzadashiq (approved):
LGTM

yusufsheiqh (approved):

commands/copilot_cli.py
lgarvey: I'm flip flopping.  `make-cloudformation` is a bit cumbersome.  Also `make-storage` seems to make more sense as it is run in response to editing `storage.yml`.

My plan initially was to have a parent command that runs everything, like `make-cfn`, but I don't think that's necessary now.

What do you think?

yusufsheiqh: I agree. The Docstring seems to agree as well that this is specific to storage configuration.

PR #54, opened 2023-06-08
Add pytest env and pre commit hooks

gabelton:
- Add pytest-env library and set AWS_REGION in pytest.ini file
- Add some more pre-commit hooks (ones already in use in the pipelines project)
- Run pre-commit against all files


lgarvey (approved):
Looks good!  👍

PR #57, opened 2023-06-10
Fix opensearch addon

lgarvey:


gabelton (approved):

templates/addons/env/opensearch.yml
```diff
@@ -72,42 +84,48 @@ Resources:
...
+        Enabled: true
       EBSOptions:
         EBSEnabled: true
         VolumeSize: !FindInMap [{{ service.prefix }}EnvScalingConfigurationMap, !Ref Env, VolumeSize]
         VolumeType: gp2
       ClusterConfig:
-        DedicatedMasterEnabled: false
+        DedicatedMasterEnabled: !FindInMap [{{ service.prefix }}EnvScalingConfigurationMap, !Ref Env, DedicatedMaster]
         InstanceCount: !FindInMap [{{ service.prefix }}EnvScalingConfigurationMap, !Ref Env, InstanceCount]
         InstanceType: !FindInMap [{{ service.prefix }}EnvScalingConfigurationMap, !Ref Env, InstanceType]
```
WillGibson, line 113: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-opensearchservice-domain-vpcoptions.html#cfn-opensearchservice-domain-vpcoptions-subnetids

If you specify more than one subnet, you must also configure ZoneAwarenessEnabled and ZoneAwarenessConfig within ClusterConfig, otherwise you'll see the error "You must specify exactly one subnet" during template creation.

Something like this untested example...

```suggestion
        InstanceType: !FindInMap [{{ service.prefix }}EnvScalingConfigurationMap, !Ref Env, InstanceType]
        ZoneAwarenessConfig:
          AvailabilityZoneCount: 
            Fn::Length: 
              !Split [ ',', { 'Fn::ImportValue': !Sub '${App}-${Env}-PrivateSubnets' } ]
        ZoneAwarenessEnabled: true
```

shehzadashiq (approved):
LGTM

templates/addons/env/opensearch.yml
lgarvey: Ah OK. Testing this now.  Did you find a cleaner way to apply both subnets?

I was hoping this would work:

```
SubnetIds: [!Split [ ',', { 'Fn::ImportValue': !Sub '${App}-${Env}-PrivateSubnets' } ] ]
```
Did you find another way to set the subnetids or did you have to select index 0, then select 1 ?

WillGibson: That looks right to me, see...

https://github.com/uktrade/digital-workspace-deploy/blob/main/copilot/environments/addons/dw-redis.yml#L74

WillGibson: Are you sticking with just one subnet for now?

PR #60, opened 2023-06-12
DBTP-149 Add ./copilot-helper.py check command

WillGibson:


tests/test_check_cloudformation.py
```diff
@@ -0,0 +1,73 @@
...
+
+    @patch("commands.check_cloudformation.valid_checks")
+    def test_runs_all_checks_when_given_all(self, valid_checks_mock, valid_checks_dict):
+        runner = CliRunner()
+        valid_checks_mock.return_value = valid_checks_dict
+
+        result = runner.invoke(check_cloudformation_command, ["all"])
+
+        assert result.exit_code == 0
+        assert result.output.__contains__("Running all checks")
```
lgarvey, line 71: Looking good. 

You shouldn't need to hit the __contains__ magic method directly.  Because the object implements it, you can use 
"x" in object,  e.g. 
```
assert "Running all checks" in result.output
```

commands/check_cloudformation.py
```diff
@@ -0,0 +1,36 @@
...
+
+from commands.cloudformation_checks.lint import lint
+
+
+@click.group()
+def copilot():
+    pass
+
+
+def valid_checks():
```
lgarvey, line 14: I wonder if there's a way to do this within click.

So:
`copilot-helper check` runs the full suite of tests
`copilot-helper check lint waf cfn` runs individual tests (that's click command chaining AFAIK)

would be valid, assuming lint, waf and cfn are part of the check command group.  But you can also run:

`copilot-helper check` to execute the full set of check.

I'd certainly favour leveraging click to do the command routing, validation etc. 

https://click.palletsprojects.com/en/8.1.x/commands/#group-invocation-without-command
https://click.palletsprojects.com/en/7.x/commands/#multi-command-pipelines

tests/test_check_cloudformation.py
```diff
@@ -0,0 +1,73 @@
+from unittest.mock import patch
+
+import pytest
+from click.testing import CliRunner
+
+import commands
+from commands.check_cloudformation import check_cloudformation as check_cloudformation_command, valid_checks
+
+
+class TestCheckCloudformationCommand:
```
gabelton, line 10: What's the rationale for using a class here? We don't do this in other test modules in copilot-tools or the pipelines repo

tests/test_check_cloudformation.py
```diff
@@ -0,0 +1,73 @@
...
+    @pytest.fixture
```
gabelton, line 11: I think @ejayesh has created a conftest file, if you wanted to separate the fixture out into that

tests/test_check_cloudformation.py
```diff
@@ -0,0 +1,73 @@
...
+        runner = CliRunner()
+        valid_checks_mock.return_value = valid_checks_dict
+
+        result = runner.invoke(check_cloudformation_command, ["does-not-exist"])
+
+        assert result.exit_code == 1
+        assert isinstance(result.exception, ValueError)
+        assert str(result.exception).__contains__("Invalid value (does-not-exist) for 'CHECK'")
+
+    test_data = [
```
gabelton, line 40: I would move this inside `test_runs_specific_check_when_given_check`, rather than have it floating, as it's only used by the one test

tests/test_check_cloudformation.py
```diff
@@ -0,0 +1,73 @@
...
+import pytest
+from click.testing import CliRunner
+
+import commands
+from commands.check_cloudformation import check_cloudformation as check_cloudformation_command, valid_checks
+
+
+class TestCheckCloudformationCommand:
+    @pytest.fixture
+    def valid_checks_dict(self):
```
gabelton, line 12: Is this meant to match the return value of the `valid_checks` method?

tests/test_check_cloudformation.py
ejayesh: Yes, pull it in and start using it.

tests/test_check_cloudformation.py
WillGibson: I took my cue from https://github.com/uktrade/copilot-tools/blob/main/tests/test_copilot_cli.py#L11

I'll lose the class though actually as I think it's better to break things out into more separate files to keep it all manageable.

tests/test_check_cloudformation.py
WillGibson: Done, reads much better now.

tests/test_check_cloudformation.py
WillGibson: It's used by the `@pytest.mark.parametrize` annotation though, so needs to be defined before that line.

tests/test_check_cloudformation.py
WillGibson: Yes. We're patching that method so that these tests can check the behaviour of the check command, not the specifics of how it ends up being used.

Idea being that we can add or remove checks, without needing to come back and adjust all these tests.

tests/test_check_cloudformation.py
lgarvey: It makes sense to just have a test_{file} for each {file} in which case it's helpful to group related tests under classes.  The class is just a grouping mechanism.

commands/check_cloudformation.py
WillGibson: It now allows for running multiple checks at once

lgarvey: You're probably still better leveraging click to do this in the example I showed.  It does useful stuff like automatically renders the doc string as help text etc. and  it seems simpler.

lgarvey: Here's the example:

```
#!/usr/bin/env python

import click


@click.group(invoke_without_command=True, chain=True)
@click.pass_context
def check(ctx):
    if ctx.invoked_subcommand is None:
        # https://click.palletsprojects.com/en/7.x/advanced/#invoking-other-commands
        ctx.invoke(lint)
        ctx.invoke(cfn)
        ctx.invoke(waf)


@check.command()
def lint():
    click.echo("Lint check")


@check.command()
def cfn():
    click.echo("CFN check")


@check.command()
def waf():
    click.echo("WAF check")


if __name__ == "__main__":
    check()

```

WillGibson: Yes, I think that makes sense. It's the next thing to play with.

commands/check_cloudformation.py
```diff
@@ -0,0 +1,71 @@
...
+
+    if len(checks) == 0:
+        checks = valid_checks()
+        running_checks = "all"
+    else:
+        running_checks = ' & '.join(', '.join(checks).rsplit(', ', 1))
+
+    for check in checks:
+        if check not in valid_checks():
+            click.secho(f"""Invalid check requested "{check}" """, fg="red")
```
yusufsheiqh, line 44:
```suggestion
            click.secho(f"Invalid check requested {check}", fg="red")
```

WillGibson: I wanted to keep the check name in quotes so it's more obvious which bit of the text it is...

image

commands/check_cloudformation.py
```diff
@@ -0,0 +1,84 @@
...
+        click.secho("The CloudFormation templates did not pass the following checks:", fg="red")
+        for failed_check in failed_checks:
+            click.secho(f"  - {failed_check}", fg="red")
+        exit(1)
+    else:
+        click.secho("The CloudFormation templates passed all the checks :-)", fg="green")
+
+
+@check_cloudformation.command()
+def lint():
```
WillGibson, line 68: I'm not sure how to unit test this one. Maybe we are happy with unit testing the main `check-cloudformation` command, which does all the orchestration?

lgarvey: You could lint something and check that a string exists in the output?

commands/check_cloudformation.py
yusufsheiqh: In that case... I reckon we have two options.

```suggestion
            click.secho(f'Invalid check requested "{check}"', fg="red")
```

```suggestion
            click.secho(f"Invalid check requested \"{check}\"", fg="red")
```

commands/check_cloudformation.py
lgarvey: I think they're saying it isn't advised for one sub command to kick off another sub command, which the example shows. This would be harder to understand from a usability perspective.  But I'm surprised there isn't already logic for a group to run all sub commands.

commands/check_cloudformation.py
```diff
@@ -0,0 +1,84 @@
...
+
+    if len(checks) == 0:
+        checks = valid_checks()
+        running_checks = "all"
+    else:
+        running_checks = ' & '.join(', '.join(checks).rsplit(', ', 1))
+
+    for check in checks:
+        if check not in valid_checks():
+            click.secho(f"""Invalid check requested "{check}" """, fg="red")
```
lgarvey, line 37: You could leverage click to do this validation and render the help page etc.

tests/test_check_cloudformation.py
```diff
@@ -0,0 +1,145 @@
...
+from unittest.mock import patch
+
+import pytest
+from click.testing import CliRunner
+from pathlib import Path
+
+from commands.check_cloudformation import check_cloudformation as check_cloudformation_command
+from commands.exceptions.CheckCloudformationFailure import CheckCloudformationFailure
+
+BASE_DIR = Path(__file__).parent.parent
```
yusufsheiqh, line 12: Is this the project root directory? I assume this test would fail if we move it elsewhere.

```suggestion
BASE_DIR = os.path.dirname(os.path.abspath(__file__))
```

commands/check_cloudformation.py
WillGibson: Maybe, but I'd have to get somehow put a known failing template in place. Currently all the addons are generated at the start of the command after completely removing the `tests/test-application/copilot` directory, so I think it might be a pain in the wotsit.

commands/check_cloudformation.py
WillGibson: Outputting the help is a good call.

I don't think we can specify what the valid arguments are for clieck when we're allowing any number from 0 to many...

https://click.palletsprojects.com/en/8.1.x/arguments/

commands/check_cloudformation.py
WillGibson: What's wrong with using triple quotes so that we don't have to escape the quotes in the string?

commands/check_cloudformation.py
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+
+@click.group(invoke_without_command=True)
+@click.argument("checks", nargs=-1)
+@click.pass_context
+def check_cloudformation(ctx, checks):
+    """Runs the checks passed in the command arguments.\n
+    If no argument is passed, it will run all the checks.
+    """
```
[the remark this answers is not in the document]
WillGibson: I actually on lined this to save vertical space in the output.

commands/check_cloudformation.py
yusufsheiqh: Not sure, but I don't think it's the Pythonic way since triple quotes are primarily used to indicate Docstrings. How about my first option? No escaping involved there.

commands/check_cloudformation.py
```diff
@@ -0,0 +1,84 @@
...
+
+    if len(checks) == 0:
```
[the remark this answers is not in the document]
WillGibson: Done

commands/check_cloudformation.py
lgarvey: In the demo I mocked up, it already does this, including pulling in the sub command doc strings and rendering those on the help page

```
./clicktest.py --help
Usage: clicktest.py [OPTIONS] COMMAND1 [ARGS]... [COMMAND2 [ARGS]...]...

Options:
  --help  Show this message and exit.

Commands:
  cfn   This command generates cfn
  lint  This command lints code
  waf   this verifies the WAF exists
```

And if you enter the wrong command:

```
./clicktest.py lintxx
Usage: clicktest.py [OPTIONS] COMMAND1 [ARGS]... [COMMAND2 [ARGS]...]...
Try 'clicktest.py --help' for help.

Error: No such command 'lintxx'.
```

commands/check_cloudformation.py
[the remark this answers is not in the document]
yusufsheiqh: Resolved in https://github.com/uktrade/copilot-tools/pull/60/commits/713ddc16235f394fdd08bdf633eec68248d7c706.

commands/check_cloudformation.py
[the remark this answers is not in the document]
yusufsheiqh: Resolved in https://github.com/uktrade/copilot-tools/pull/60/commits/6de3e8b9b989231becc4b097d2c4965ae42be843.

tests/test_check_cloudformation.py
WillGibson: It would still fail if we moved things about with your suggestion. I'm not clear on the benefits of using one approach over another.

yusufsheiqh: You are right! We could have this in the `utils` file, set it there and then import it. Not a big deal though, of course.

yusufsheiqh: I have now realised that we already set this in the `conftest.py` file, so we could simply import it here.

commands/check_cloudformation.py
WillGibson: Being able to declare a string like that and not mess about escaping quotes or changing the quotes to get around needing to escape them is a something I like about Python

I agree that triple quotes are used for Docstrings.

I cannot find anything saying it is unPythonic to use them for other strings, quite the opposite in fact.

commands/check_cloudformation.py
```diff
@@ -0,0 +1,67 @@
...
+    if result.returncode == 0:
+        ctx.obj["passing_checks"].append("lint")
+    else:
+        click.secho(result.stderr.decode())
+        ctx.obj["failing_checks"].append("lint")
+
+
+@check_cloudformation.result_callback()
+@click.pass_context
+def process_result(ctx, result):
```
WillGibson, line 56: There's no unit test for this. I will have a look at whether I can override `prepare_cloudformation_templates()` for the tests and put some known quantity CloudFormation templates in position.

tests/test_check_cloudformation.py
```diff
@@ -0,0 +1,52 @@
...
+
+def test_runs_all_checks_when_given_no_arguments():
+    result = CliRunner().invoke(check_cloudformation_command)
+
+    assert ">>> Running all checks" in result.output
+    assert ">>> Running lint check" in result.output
+
+
+def test_prepares_cloudformation_templates():
+    def path_exists(path):
```
lgarvey, line 18: Nitpicking, but if you use the new python 3 pathlib stuff, this is superfluous

```
from pathlib importh Path

path = Path("my/path")

if path.exists():
 ...

File: tests/test_check_cloudformation.py (hunk 10 above)
Comment (yusufsheiqh) [reply tests/test_check_cloudformation.py]: Resolved in https://github.com/uktrade/copilot-tools/pull/60/commits/a8569e306874b8c83f5c846cc7a9afc54ae47093.

File: commands/check_cloudformation.py (hunk 7 above)
Comment (lgarvey) [reply commands/check_cloudformation.py]: PEP8 doesn't have much to say except avoid slashes and be consistent: https://peps.python.org/pep-0008/#string-quotes

File: tests/test_check_cloudformation.py (hunk 14 above)
Comment (WillGibson) [reply tests/test_check_cloudformation.py]: Done. Code more readable now :-)

File: commands/check_cloudformation.py (hunk 13 above)
Comment (WillGibson) [reply commands/check_cloudformation.py]: There are unit tests for this now.

**Review by yusufsheiqh — COMMENTED** [review]
```
$ ./copilot_helper.py check-cloudformation

>>> Preparing CloudFormation templates

>>> Generating Copilot configuration files

...
```

Is this expected behaviour when running this command? I get a lot of initial output such as the instructions as well... I have omitted this in the above terminal output.

WillGibson:
> ```
> $ ./copilot_helper.py check-cloudformation
> 
> >>> Preparing CloudFormation templates
> 
> >>> Generating Copilot configuration files
> 
> ...
> ```
> 
> Is this expected behaviour when running this command? I get a lot of initial output such as the instructions as well... I have omitted this in the above terminal output.

Yes, it generates CloudFormation templates for a test application so that it can then run the tests against them and check that we haven't added badness in the templates they are generated from.

PR #64, opened 2023-06-14
Add a test for the copilot/get-secret command

lgarvey:


commands/copilot_cli.py
```diff
@@ -239,16 +239,14 @@ def _validate_arn(arn):
...
 @copilot.command()
-@click.argument("service_name", type=str)
-@click.argument("env", type=str, default="prod")
-def get_service_secrets(service_name, env):
-    """List secret names and values for a service."""
-
-    ensure_cwd_is_repo_root()
+@click.argument("app", type=str)
+@click.argument("env", type=str)
+def get_secrets(app, env):
```
yusufsheiqh, line 244: Should we rename this to be a bit more specific or does it not matter since we have `env` as an argument anyway?

```suggestion
def get_env_secrets(app, env):
```

yusufsheiqh (approved):

commands/copilot_cli.py
lgarvey: Yep, that makes sense, particularly given it emphasises secrets are bound to the env.

PR #65, opened 2023-06-15
Update SSM put_parameter to use Intelligent Tiering

shehzadashiq:
Configuring SSM to use intelligent tiering to allow for secrets with greater than 4096 characters

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ssm/client/put_parameter.html


gabelton (approved):

yusufsheiqh (commented):
Is there a way to test this change? A unit test addressing this change would be great.

lgarvey (approved):
Looks good.

A test is probably more hassle than it's worth here.  But if you were to do it, you could patch the boto3 function and check the tier arg is set correctly.

PR #69, opened 2023-06-16
update pre-commit test command and add sample-envrc file

gabelton:


sample-envrc
```diff
@@ -0,0 +1 @@
+export AWS_PROFILE=tools
```
yusufsheiqh, line 1: Do we need to export it?

```suggestion
AWS_PROFILE=tools

```

yusufsheiqh (approved):

sample-envrc
yusufsheiqh: Yes, we do.

Reference: https://direnv.net

PR #71, opened 2023-06-16
refactor domain

ejayesh:
fix for when we have multiple services in a project and the correct domain name was not getting set.


tests/test_dns_cli.py
```diff
@@ -142,7 +142,10 @@ def test_check_domain(check_aws_conn, check_r53, fakefs):
         "commands.dns_cli.check_response",
         return_value="{}"
 )
-def test_assign_domain(check_aws_conn, check_response):
+@patch(
```
yusufsheiqh, line 145: We can make this a single line. I thought our `pre-commit` hooks would take care of this.

yusufsheiqh (approved):

tests/test_dns_cli.py
ejayesh: fixed

PR #75, opened 2023-06-20
Prepend rediss:// to Redis URL

lgarvey:
Also the output is not needed for env addons

NOTE: This will cause problems for already deployed stacks which will prepend rediss:// to the URL in settings.py


yusufsheiqh (approved):
This was very much needed! 🚀

PR #77, opened 2023-06-20
Add Python package documentation

yusufsheiqh:
## Context

- Add documentation for the Python package separately
  - Add instructions for installation and basic usage - this can be fleshed out in detail.
- Update the documentation link to point to `commands/docs.md`


commands/README.md
```diff
@@ -0,0 +1,50 @@
+# DBT Copilot Tools
+
+This package contains a set of tools in the form of a Command Line Interface (CLI) for transferring applications/services from [GOV.UK PaaS](https://www.cloud.service.gov.uk) to Department for Business and Trade (DBT) PaaS which augments [AWS Copilot](https://aws.github.io/copilot-cli/).
```
WillGibson, line 3: I think some of the commands will also be used in an ongoing way, not just for migration, as an application's requirements evolve, like if they needed an additional addon, or possibly to apply best practice changes to the CloudFormation templates.

WillGibson (commented):
Looking good. One small comment, which may apply to the main `README.md` too.

gabelton (approved):

commands/README.md
yusufsheiqh: Updated in https://github.com/uktrade/copilot-tools/pull/77/commits/6dc71e41c39acfb30e07910ac5af40024f23e5bc. Thoughts?

WillGibson: Looks good.

WillGibson (approved):

PR #78, opened 2023-06-20
Add Publishing documentation

yusufsheiqh:
## Context

- Add documentation on how to update and publish the Python package `dbt-copilot-tools`.


README.md
```diff
@@ -36,6 +36,43 @@ Run `docker build -f Dockerfile.test -t alpine/python .` to build the image.
 
 For Platform developers, the `push` commands can be found in [AWS ECR](https://eu-west-2.console.aws.amazon.com/ecr/repositories).
 
+## Publishing
+
+To publish the Python package `dbt-copilot-tools`, you will need an API token.
+
+1. Acquire credentials from [Passman for PyPi](https://passman.ci.uktrade.digital/secret/244eb2ad-245d-47d9-9f6a-95ee30810944/).
+   - Request access from the SRE team.
```
WillGibson, line 44: I'm not convinced that this process gives us a tidy way of revoking someone's access if they leave. 

Can we use Organisations/teams, assign people to the correct things and then we all use our individual accounts instead of a shared one?

yusufsheiqh: It's part of a group in Passman and logging into PyPi requires an MFA token which is also in Passman. It would be the same process if we all have individual accounts - remove/deactivate user or remove user from group so they will not have access to the secret anymoar.

WillGibson: Using a shared identity feels like something we should avoid to me.

I don't have much experience of PyPi, but it looks to me like it supports organisations and groups in much the same way as (for example) GitHub.

So I would hope that we would use our individual identities to login etc. and these would have permissions to do what the need to do at the individual, group or organisation level.

Maybe a shared identity is an existing thing and changing it is outside of the scope of this pull request? That would be a fair response, but maybe it could go on the SRE team radar to look at?

yusufsheiqh: Seems like a fairly new thing over at PyPi - so it is definitely something worth exploring actually. Here is the blog post: https://blog.pypi.org/posts/2023-04-23-introducing-pypi-organizations/

I have requested for an organisation to be created - it seems like it needs to go through a reviewal process, so have added a [ticket](https://uktrade.atlassian.net/browse/SR-1809) to keep this on the radar.

PR #79, opened 2023-06-20
DBT-149 Address cfn-lint findings

WillGibson:
Now we have the `check-cloudformation lint` command, it seems appropriate to make it pass.

### DIFF
--- a/templates/addons/env/rds-postgres.yml
+++ b/templates/addons/env/rds-postgres.yml
@@ -19,10 +19,10 @@ Mappings:
     {{ env_name }}:
       DBInstanceClass: '{{ config.instance }}'
       MultiAZ: {% if config.replicas %} true {% else %} false {% endif %}
-      AllocatedStorage: {{ config.volumesize }}
+      AllocatedStorage: {{ config.volume_size }}
 
     {% endfor %}
-  
+
 Resources:
   # Subnet group to control where the DB gets placed
   DBSubnetGroup:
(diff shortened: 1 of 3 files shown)


templates/addons/env/rds-postgres.yml
```diff
@@ -19,10 +19,10 @@ Mappings:
     {{ env_name }}:
       DBInstanceClass: '{{ config.instance }}'
       MultiAZ: {% if config.replicas %} true {% else %} false {% endif %}
-      AllocatedStorage: {{ config.volumesize }}
+      AllocatedStorage: {{ config.volume_size }}
```
WillGibson, line 22: `cfn-lint` picked up on this regression.

templates/addons/env/rds-postgres.yml
```diff
@@ -19,10 +19,10 @@ Mappings:
...
 
     {% endfor %}
-  
+
```
WillGibson, line 25: My IDE helped.

templates/addons/env/s3.yml
```diff
@@ -35,15 +35,16 @@ Resources:
...
         - Key: Copilot-name
-          Value: !Ref Name 
+          Value: !Ref Name
 
   {{ service.prefix }}BucketPolicy:
     Metadata:
       'aws:copilot:description': 'A bucket policy to deny unencrypted access to the bucket and its contents'
     Type: AWS::S3::BucketPolicy
     DeletionPolicy: Retain
+    UpdateReplacePolicy: Retain
```
WillGibson, line 47: See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatereplacepolicy.html

templates/addons/svc/s3-policy.yml
```diff
@@ -1,3 +1,9 @@
+Metadata:
+  cfn-lint:
+    config:
+      ignore_checks:
+        - W2001 # Parameter not used
```
WillGibson, line 5: We could remove the unused `Parameters`, but @lgarvey and something I saw [here](https://github.com/aws/copilot-cli/issues/4730#issuecomment-1595029932) suggest they are needed by Copilot for something. Is it for the tags in AWS?

yusufsheiqh (approved):

templates/addons/env/s3.yml
WillGibson: Would be good to get your eyes on this one in particular before I merge @lgarvey

lgarvey: So s3 buckets names are unique. In this instance the bucket would be retained, and then if we attempt to recreate the env it would fail due to the existing bucket name.  That doesn't seem to be an issue though - in this instance, they could replace the s3 type with an s3-policy so that access to the bucket is maintained.

PR #81, opened 2023-06-22
DBTP-287 Use addons.parameters.yml to get environment stack references into the addons stack

WillGibson:
https://uktrade.atlassian.net/browse/DBTP-287

### DIFF
--- a/templates/addons/env/addons.parameters.yml
+++ b/templates/addons/env/addons.parameters.yml
@@ -1,2 +1,6 @@
 Parameters:
+  EnvironmentSecurityGroup: !Ref EnvironmentSecurityGroup
   LBARN: !Ref PublicLoadBalancer
+  PrivateSubnet1: !Ref PrivateSubnet1
+  PrivateSubnet2: !Ref PrivateSubnet2
+  VPCID: !Ref VPC
--- a/templates/addons/env/opensearch.yml
+++ b/templates/addons/env/opensearch.yml
@@ -7,6 +7,10 @@ Parameters:
     Type: String
     Description: The environment name your service, job, or workflow is being deployed to.
 
+  # Parameters from the parent stack brought in via addons.parameters.yml...
+  EnvironmentSecurityGroup:
+    Type: String
+
 Transform: AWS::LanguageExtensions
 
 Mappings:
@@ -77,7 +81,7 @@ Resources:
       IpProtocol: tcp
       FromPort: 443
       ToPort: 443
-      SourceSecurityGroupId: { 'Fn::ImportValue': !Sub '${App}-${Env}-EnvironmentSecurityGroup' }
+      SourceSecurityGroupId: !Ref EnvironmentSecurityGroup
 
   # Single opensearch instance
   {{ service.prefix }}OpenSearchDomain:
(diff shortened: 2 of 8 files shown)


templates/addons/env/addons.parameters.yml
```diff
@@ -1,2 +1,6 @@
 Parameters:
+  EnvironmentSecurityGroup: !Ref EnvironmentSecurityGroup
```
WillGibson, line 2: @lgarvey This parameter is currently only needed if one or more of the following conditions is satisfied...

* We need a `redis-cluster`
* We need a `shared` `aurora-postgres`
* We need an `rds-postgres`
* We need an `opensearch`

In the `aurora-postgres` template I made including the parameter optional so the linting is not upset by the unused parameter.

However, now when I try to deploy `update-supply-chain-information` using newly generated CloudFormation templates it goes...

```
➜ copilot env deploy --name development
✘ upload artifacts for environment development: parse environment addons: template does not require the parameter "EnvironmentSecurityGroup" in parameters file
```

I am a Jinja2 newbie, but I think it should be possible to only include it in this file if the conditions listed are satisfied?

I do suspect that we do not need this for version `0.1` though, we could just document deleting unused entries from addons.parameters.yml` and ticket this up for `0.2`.

What do you think?

WillGibson: I definitely think it should at least be deferred to s separate pull request as this one already solves some major future developer headaches. I will create the ticket now, we can prioritise it later.

lgarvey: Yes, this makes sense.  It should be fairy trivial to add this.

templates/addons/env/opensearch.yml
```diff
@@ -178,11 +188,11 @@ Resources:
         ZoneAwarenessConfig:
           AvailabilityZoneCount: 2
           #Fn::length always resolves to 1 despite there being subnets.
-          #Fn::Length: [ !Split [ ',', { 'Fn::ImportValue': !Sub '${App}-${Env}-PrivateSubnets' } ] ]
+          #Fn::Length: [ !Ref PrivateSubnet1, !Ref PrivateSubnet2 ]
```
yusufsheiqh, line 191: This is commented out ATM - is that intentional?

yusufsheiqh (approved):

templates/addons/env/opensearch.yml
WillGibson: It was commented out before. I supect it relates to some woes @lgarvey is looking at with this addon.

templates/addons/env/opensearch.yml
```diff
@@ -1,11 +1,24 @@
 Parameters:
+  # Copilot required Parameters...
   App:
     Type: String
     Description: Your application's name.
   Env:
     Type: String
     Description: The environment name your service, job, or workflow is being deployed to.
+
```
lgarvey, line 9: I don't think this is necessary.  Does Opensearch have any sequencing issues?

WillGibson: I wouldn't know, I haven't had to deploy/delete one yet, but I expect it would trip up on `EnvironmentSecurityGroup` like the Aurora Postgres addon did and, to me anyway, it makes sense refer to things from the environment stack in a consistent way in all the addons.

templates/addons/env/opensearch.yml
WillGibson: We should probably have a ticket to come back and figure this one out.

PR #82, opened 2023-06-22
Fix "duplicate SSM parameter name" issue in Opensearch addon template

lgarvey:
The error was due to having two SSM parameters with the same SSM name, even though CFN conditions meant that only one parameter would be active.

The solution was to use an Fn::if statement to select the endpoint url from the HA or non HA opensearch domain resource based on the EnableHA condition.  However, this resulted in another issue (possibly a CFN bug, or limitation) due to a Fn:If and a complicated nested statement. To solve this I created an intermediate SSM parameter which stores the domain endpoint url, then a final config parameter which retrieves the endpoint url, and the username and password from secrets manager.  We use one secrets manager secret and two SSM parameters, which is not ideal, but required to overcome limitations in cloudformation.

### DIFF
--- a/templates/addons/env/opensearch.yml
+++ b/templates/addons/env/opensearch.yml
@@ -52,7 +52,6 @@ Resources:
       'aws:copilot:description': 'A Secrets Manager secret to store your OS credentials'
     Type: AWS::SecretsManager::Secret
     Properties:
-      Name: !Sub '/copilot/${App}/${Env}/secrets/{{ service.name|upper|replace("-", "_") }}'
       Description: !Sub Aurora main user secret for ${AWS::StackName}
       GenerateSecretString:
         SecretStringTemplate: '{"username": "opensearch"}'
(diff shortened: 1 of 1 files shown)


templates/addons/env/opensearch.yml
```diff
@@ -193,35 +192,32 @@ Resources:
...
-          password:
-            !Join [ "",  [ '{% raw %}{{resolve:secretsmanager:{% endraw %}', !Ref {{ service.prefix }}OpenSearchSecret, "{% raw %}:SecretString:password}}{% endraw %}" ]]
+      Value: !If
+        - EnableHA
+        - !Sub
+          - "https://${url}"
+          - url: !GetAtt dwOpensearchOpenSearchDomainHA.DomainEndpoint
+        - !Sub
+          - "https://${url}"
+          - url: !GetAtt dwOpensearchOpenSearchDomain.DomainEndpoint
```
yusufsheiqh, line 208: `dwOpensearchOpenSearchDomain` -> Looks like this is specific to `dw`(?).

yusufsheiqh: Might need replacing with the `service.prefix`.

shehzadashiq (approved):
LGTM

templates/addons/env/opensearch.yml
lgarvey: Yep, doh.

yusufsheiqh: Resolved in https://github.com/uktrade/copilot-tools/pull/82/commits/fc0d2eb458b574a2d80fe900d85f0a9f03c2f5a7.

WillGibson (approved):

PR #83, opened 2023-06-22
Add Commands documentation

yusufsheiqh:
## Context

We need to be able to automatically generate documentation for all commands, so that we  are aware of what commands are available and how to potentially use them. This will prove to be exceptionally useful as we continue to develop the CLI commands.

- Add script to automatically generate commands documentation
  - Tests are added as well.
  - Example CLI is also added to test the template output.
- Add commands documentation (see [COMMANDS.md](https://github.com/uktrade/copilot-tools/blob/add-command-docs/commands/COMMANDS.md) for a preview)
- Add `pre-commit` hook to keep docs up to date when committing
- Update Python package version to `0.1.5`
- Update `README.md` documentation to link to Commands documentation.


tests/test_create_command_docs.py
```diff
@@ -0,0 +1,76 @@
...
+class TestCreateCommandDocsCli(TestCase):
+    def setUp(self) -> None:
+        self.runner = CliRunner()
+
+    @classmethod
+    def tearDownClass(cls):
+        os.remove(f"{BASE_DIR}/tests/fixtures/docs.md")
+
+    def test_check_required_module_option(self):
+        result = self.runner.invoke(docs)
```
WillGibson, line 19: Might be worth passing the `cmd` parameter here to be sure it's failing for the right reason. E.g. pass everything except the one we are testing.

Comment applies on the others probably too.

tests/test_create_command_docs.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def test_check_required_output_option(self):
+        result = self.runner.invoke(docs, ["--module", "foo", "--cmd", "bar"])
+
+        output = result.output
+
+        assert result.exit_code != 0
+        assert "Error: Missing option '--output' / '-o'." in output
+
+    def test_check_valid_module_option(self):
```
WillGibson, line 42: Maybe...
```suggestion
    def test_check_invalid_module_option(self):
```

tests/test_create_command_docs.py
```diff
@@ -0,0 +1,76 @@
...
+        result = self.runner.invoke(docs, ["--module", "foo", "--cmd", "bar", "--output", "baz"])
+
+        output = result.output
+
+        assert result.exit_code != 0
+        assert "Could not find module: foo. Error: No module named 'foo'" in output
+
+    def test_check_valid_cmd_option(self):
```
WillGibson, line 50: Maybe...
```suggestion
    def test_check_invalid_cmd_option(self):
```

tests/test_create_command_docs.py
```diff
@@ -0,0 +1,76 @@
...
+        result = self.runner.invoke(docs, ["--module", "copilot_helper", "--cmd", "bar", "--output", "baz"])
+
+        output = result.output
+
+        assert result.exit_code != 0
+        assert "Error: Could not find command bar in copilot_helper module" in output
+
+    def test_create_command_docs(self):
+        output_path = f"{BASE_DIR}/tests/fixtures/docs.md"
```
WillGibson, line 59: Maybe delete this file and assert it does not exist as part of the arrange.

utils/create_command_docs.py
```diff
@@ -0,0 +1,191 @@
...
+## Options
+
+{options}
+
+{commands}
+
+## CLI Help
+
+```
+{help}
```
WillGibson, line 35: I think the stuff above with the heading basically duplicates this help output, so we probably only need one. With the headings seems nice in the context of a web page.

tests/test_create_command_docs.py
```diff
@@ -0,0 +1,76 @@
+import os
+from unittest import TestCase
+
+from click.testing import CliRunner
+
+from tests.conftest import BASE_DIR
+from utils.create_command_docs import docs
+
+
+class TestCreateCommandDocsCli(TestCase):
```
WillGibson, line 10: I wonder if we can have a test that it generates the documentation correctly? Not sure how practical that is?

WillGibson (commented):
Looks great on a quick first pass, really nice to be able to review the commands on a single place like that.

tests/test_create_command_docs.py
yusufsheiqh: This would be quite difficult for our current implementation, but also not sure how practical it would be to test this given this is more of a UI test. I could create a small CLI module for the purpose of testing and we could check the documentation with tests, what are your thoughts?

utils/create_command_docs.py
yusufsheiqh: I was thinking the same actually. This bit is quite likely redundant - will remove this in a subsequent commit.

tests/test_create_command_docs.py
yusufsheiqh: The file is always deleted at the end of the test suite. I could add the assertion though to check that the file does not exist yet - will add this in a subsequent commit.

tests/test_create_command_docs.py
yusufsheiqh: Definitely worth doing that. Resolved in https://github.com/uktrade/copilot-tools/pull/83/commits/7ac92954f136625fd97c0bb01fe97877a3539733.

tests/test_create_command_docs.py
WillGibson: That sounds like a reasonable and not too time consuming suggestion.

tests/test_create_command_docs.py
yusufsheiqh: Resolved in https://github.com/uktrade/copilot-tools/pull/83/commits/225e94c3b9eb6d7be3b7f81b03df9903b43bd707.

utils/create_command_docs.py
yusufsheiqh: Resolved in https://github.com/uktrade/copilot-tools/pull/83/commits/20bff77010b784638062155a21bd3886f4e24367.

tests/test_create_command_docs.py
yusufsheiqh: Resolved in https://github.com/uktrade/copilot-tools/pull/83/commits/08b6f086173ae08a0e1f7ec72507266e97b248bc.

tests/test_create_command_docs.py
```diff
@@ -0,0 +1,102 @@
...
+
+        output = result.output
+
+        assert result.exit_code != 0
+        assert "Error: Could not find command bar in copilot_helper module" in output
+
+    def test_create_command_docs(self):
+        output_path = f"{BASE_DIR}/tests/test-docs/test-docs.md"
+
+        assert os.path.exists(output_path) is False
```
lgarvey, line 62: Bit more pythonic. Also pathlib is the new shiny replacement to os.path.*

```
assert not os.path.exists(output_path)
```

utils/create_command_docs.py
```diff
@@ -0,0 +1,194 @@
...
+    table_of_contents = ["# Commands Reference\n"]
+    markdown_docs = []
+
+    for meta in get_cmd_metadata(base_command):
+        table_of_contents.append(f"- [{meta.name}](#{meta.name.replace(' ', '-').lower()})")
+
+        markdown_docs.append(
+            TEMPLATE.format(
+                command_name=meta.name,
+                parent=f"[↩ Parent]({meta.parent_reference})" if meta.parent_reference else "Base command.",
```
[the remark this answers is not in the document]
WillGibson: Same would apply to other things like description and arguments, but I'm totally happy for it to stay as it is and iterate on that in the future.

WillGibson (approved):
Looks good, there's nothing in here that makes it unmergeworthy as far as I am concerned.

utils/create_command_docs.py
```diff
@@ -0,0 +1,194 @@
...
+from pathlib import Path
+from typing import Any
+from typing import Generator
+from typing import List
+from typing import NamedTuple
+from typing import Optional
+
+import click
+
+TEMPLATE = """
```
lgarvey, line 13: Is it worth leveraging Jinja2?

tests/test_create_command_docs.py
yusufsheiqh: Resolved in https://github.com/uktrade/copilot-tools/pull/83/commits/2e556097cc5247d42d0a4ebe2cbda92397599860.

utils/create_command_docs.py
yusufsheiqh: Yes. Resolved in https://github.com/uktrade/copilot-tools/pull/83/commits/074403a9c2867a89ef12e585a955ad11013e839d.

utils/create_command_docs.py
```diff
@@ -0,0 +1,194 @@
...
+    for meta in get_cmd_metadata(base_command):
+        table_of_contents.append(f"- [{meta.name}](#{meta.name.replace(' ', '-').lower()})")
+
+        markdown_docs.append(
+            TEMPLATE.format(
+                command_name=meta.name,
+                parent=f"[↩ Parent]({meta.parent_reference})" if meta.parent_reference else "Base command.",
+                description=meta.description if meta.description else "No description.",
+                usage=meta.usage,
+                arguments="## Arguments\n\n"
```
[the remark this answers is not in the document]
yusufsheiqh: Resolved in https://github.com/uktrade/copilot-tools/pull/83/commits/074403a9c2867a89ef12e585a955ad11013e839d.

utils/create_command_docs.py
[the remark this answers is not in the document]
yusufsheiqh: Resolved in https://github.com/uktrade/copilot-tools/pull/83/commits/074403a9c2867a89ef12e585a955ad11013e839d.

PR #86, opened 2023-06-27
Dbtp 282 waf management

ejayesh:
Code to give the ability to add either a default waf rule which already exists or to add a custom waf rule supplied via a CF template.

Missing Unit testing and CF linting.  This can be done in a seperate PR.

### DIFF
--- a/commands/dns_cli.py
+++ b/commands/dns_cli.py
@@ -177,10 +177,10 @@ def add_records(client, records, subdom_id, action):
 
     check_response(response)
     click.echo(
-        click.style(f"{records['Name']}, Type: {records['Type']}", fg="white", bold=True) +
-        click.style("Added.", fg="magenta")
-        )
-    return response['ChangeInfo']['Status']
+        click.style(f"{records['Name']}, Type: {records['Type']}", fg="white", bold=True)
+        + click.style("Added.", fg="magenta"),
+    )
+    return response["ChangeInfo"]["Status"]
 
 
 def check_for_records(client, parent_id, subdom, subdom_id):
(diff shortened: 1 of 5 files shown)


commands/dns_cli.py
```diff
@@ -177,10 +177,10 @@ def add_records(client, records, subdom_id, action):
...
     check_response(response)
     click.echo(
-        click.style(f"{records['Name']}, Type: {records['Type']}", fg="white", bold=True) +
-        click.style("Added.", fg="magenta")
-        )
-    return response['ChangeInfo']['Status']
+        click.style(f"{records['Name']}, Type: {records['Type']}", fg="white", bold=True)
+        + click.style("Added.", fg="magenta"),
+    )
+    return response["ChangeInfo"]["Status"]
```
shehzadashiq, line 184: There is a trailing comma at the end. Should that be removed?

WillGibson (commented):
```
poetry install --with pre-commit && pre-commit install
Installing dependencies from lock file
Warning: poetry.lock is not consistent with pyproject.toml. You may be getting improper dependencies. Run `poetry lock [--no-update]` to fix it.

Because dbt-copilot-tools depends on cfn-flip (1.3.0) which doesn't match any versions, version solving failed.
```

gabelton (approved):

ejayesh:
fixing poetry

WillGibson (commented):
The check-cloudformation stuff is pretty easy to add in. You just need to invoke the appropriate command(s) here and the resulting CloudFormation templates will be checked...

https://github.com/uktrade/copilot-tools/blob/main/commands/check_cloudformation.py#L78

PR #88, opened 2023-06-27
DBTP-302 Strip out old WAF stuff

WillGibson:
It's cruft now.


yusufsheiqh (commented):
I think you may need to merge/rebase `main`. This will update the `pre-commit` hook so that the commands document can be updated automagically.

PR #89, opened 2023-06-28
Update `pre-commit` configuration to check file changes

yusufsheiqh:
## Context

- Update `pre-commit` configuration to check for tracked file changes as a result of `pre-commit` hooks
  - Stash all changes, check if there are any changes, remove stashed state.
- Update order of `pre-commit` hooks
  - Remove stashed state at the end of the `pre-commit` workflow.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,12 @@
 repos:
+  - repo: local
+    hooks:
+      - id: stash-all-files
+        name: Stash all files including untracked files
+        entry: git stash --keep-index --include-untracked
+        language: system
+        pass_filenames: false
+        always_run: true
   - repo: https://github.com/psf/black
     rev: 23.3.0
     hooks:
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -1,4 +1,12 @@
 repos:
+  - repo: local
+    hooks:
+      - id: stash-all-files
```
lgarvey, line 4: Ah, so only staged files will have precommit hook checks ran on them?  This probably fits my workflow better as I never do a `git add --all`  👍

.pre-commit-config.yaml
```diff
@@ -65,3 +65,27 @@ repos:
...
+        pass_filenames: false
+        always_run: true
+  - repo: local
+    hooks:
+      - id: pop-stashed-files
+        name: Remove stash state and apply on current working tree state
+        entry: git stash pop
+        language: system
+        pass_filenames: false
+        always_run: true
```
WillGibson, line 91: Does this mean that it will always run even if an earlier hook failed?

WillGibson: Yes, as discussed verbally just now.

WillGibson (approved):

.pre-commit-config.yaml
yusufsheiqh: Correct! 😄

PR #95, opened 2023-06-29
DBTP-305 Hardcode path to copilot directory and some more bootstrap.yml

WillGibson:


tests/bootstrap_strings.py
```diff
@@ -1,7 +1,4 @@
...
...as to AWS/Copilot\n\ncopilot-bootstrap.py migrate-secrets bootstrap.yml\n\nYou need to be authenticated via the CF CLI for this command to work.\nUse the --dry-run flag initially to check the output\n\n# 6. deploy each service into each environment\n\n# test-service:\ncopilot svc deploy --name test-service --env test\ncopilot svc deploy --name test-service --env production\n\n\n# 7. Point DNS entries to ALB urls\n\nSpeak to SRE.\n\n# 8. Generate the storage addons\n\nCreate a storage.yml and add the relevant configuration.  See SRE for assistance.\n\nGenerate environment level storage addons\n\ncopilot-bootstrap generate-storage storage.yaml\n\nAdd required secrets to each service's manifest.yml.\n\nRedeploy each environment to provision those secrets\n\n\ncopilot env deploy --name test\ncopilot env deploy --name production\n\n# 9. Switch the bootstrap container to the service container\n\nEdit each copilot/{{service-name}}/manifest.yaml file and replace the image_location with the service's ECR registry url and redeploy.\n\nIt's expected that you'll need to iterate over this step a number of times whilst making changes to the services manifest.yml until the service is stable.\n\n"
```
WillGibson, line 4: Just hardcoded it to "bootstrap.yml" in there somewhere.

tests/test_dns_cli.py
```diff
@@ -1,10 +1,16 @@
 from unittest.mock import patch
```
WillGibson, line 1: Black went to town in here.

lgarvey (approved):
Looks good. 

I'm not too concerned about this command being opinionated, as it'll only be used during the initial migration. But also, it seems fine to remove the option.

tests/bootstrap_strings.py
```diff
@@ -1,7 +1,4 @@
-from pathlib import Path
-
 TEST_MANIFEST = '# The manifest for the "dev" environment.\n# Read the full specification for the "Environment" type at:\n#  https://aws.github.io/copilot-cli/docs/manifest/environment/\n\n# Your environment name will be used in naming your resources like VPC, cluster, etc.\nname: test\ntype: Environment\n\n# Import your own VPC and subnets or configure how they should be created.\n# network:\n#   vpc:\n#     id:\n\n# Configure the load balancers in your environment, once created.\n\nhttp:\n  public:\n    certificates:\n      - ACM-ARN-FOR-test.landan.cloudapps.digital\n#   private:\n\n\n# Configure observability for your environment resources.\nobservability:\n  container_insights: true\n'
```
lgarvey, line 1: Just pointing out, you raised an issue with the env names in these test files in case you did want to remediate here.

WillGibson: 🤣 Brilliant. Touché. I walked right into that one. Let me sort it out so that we never have to speak of this again!

WillGibson: I kind of found a bug, but it's only the environment hardcoded in a comment in the template, so wouldn't break anything, just confuse anyone who actually read it.

WillGibson: Sorted I think.

PR #99, opened 2023-07-12
Fix rds recommendations

gabelton:
part of https://uktrade.atlassian.net/browse/DBTP-308

recommendations:
- **DB clusters with only one DB instance** - added logic to set MultiAZ property to true for prod envs
- **Old minor versions** - added `AutoMinorVersionUpgrade: true`
- **Encryption isn't enabled for the DB Cluster** - added `StorageEncrypted: true`
- **Performance Insights is not enabled** - added `EnablePerformanceInsights: true`
- **Running a DB instance with magnetic storage** - added logic for setting StorageType based on env, as per [these docs](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Storage.html)
- **DB instance isn't a Multi-AZ DB instance** - this logic is already there, but we don't want multi instances for non-prod dbs
- **Storage autoscaling isn't turned on** - set `MaxAllocatedStorage` to be 26% higher than `AllocatedStorage` as per [aws recommendations](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_PIOPS.StorageTypes.html#USER_PIOPS.Autoscaling)

### DIFF
--- a/commands/check_cloudformation.py
+++ b/commands/check_cloudformation.py
@@ -31,6 +31,15 @@ def check_cloudformation(ctx: click.Context) -> None:
             ctx.invoke(command)
 
 
+def get_lint_result(path: str):
+    command = ["cfn-lint", path]
+
+    click.secho(f"\n>>> Running lint check", fg="yellow")
+    click.secho(f"""    {" ".join(command)}\n""", fg="yellow")
+
+    return run(command, capture_output=True)
+
+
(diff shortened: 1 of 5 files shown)


shehzadashiq (approved):
LGTM

yusufsheiqh (approved):

PR #105, opened 2023-07-19
Implement trusted advisor addons recommendations

gabelton:
https://uktrade.atlassian.net/browse/DBTP-308

Fixing various trusted advisor recommendations and a few bugs encountered when redeploying return-to-office.

Recommendations and action taken:
- Enable bucket logging for most buckets - adding a logging bucket and policy to s3 template
- Enable bucket versioning on most buckets to prevent accidental deletion or overwriting - updated s3 template
- S3 Block Public Access setting should be enabled at the bucket-level - set `IgnorePublicAcls` and `RestrictPublicBuckets` to true in s3 template
- Amazon ECS Containers should only have read-only access to its root filesystems - set `storage.readonly_fs: true` in public and backend templates

Bugs with rds template:
- `MaxAllocatedStorage` needs to be an int
- `EngineVersion` 13.4 not supported


WillGibson:
Looks like you have conflicts to resolve @gabelton.

Also, how has this been tested?

templates/addons/env/rds-postgres.yml
```diff
@@ -73,7 +73,7 @@ Resources:
       AutoMinorVersionUpgrade: true
       EnablePerformanceInsights: true
       Engine: postgres
-      EngineVersion: '13.4'
+      EngineVersion: '13'
```
WillGibson, line 76: Not for this pull request, but do we want to make this something that developer teams can specify themselves?

templates/addons/env/s3.yml
```diff
@@ -19,6 +16,51 @@ Mappings:
     {% endfor %}
 
 Resources:
+  {{ service.prefix }}LoggingBucket:
+    Metadata:
+      'aws:copilot:description': 'An Amazon S3 bucket to store logs from {{ service.prefix }}Bucket'
+    Type: AWS::S3::Bucket
+    Properties:
+      BucketName: !Join [-, [!FindInMap [{{ service.prefix }}BucketNameMap, !Ref Env, BucketName], logs]]
```
WillGibson, line 24: `!Join [-, [!FindInMap [{{ service.prefix }}BucketNameMap, !Ref Env, BucketName], logs]]` appears twice in this template. I wonder if we can set it in a parameter to remove the risk that we might change it in one place, but not the other?

yusufsheiqh:
To resolve the conflicts... you will have to rebase `main` and make sure the `templates` folder is inside the `commands` folder.

templates/addons/env/rds-postgres.yml
WillGibson: [DBTP-336 Allow overriding of default Postgres version](https://uktrade.atlassian.net/browse/DBTP-336)

gabelton:
I've resolved merge conflicts and, in terms of testing, I had to fix one of the `make_config` tests and tested manually by redeploying return to office and adding s3 buckets. I think there's probably a gap in our unit testing, though, because more tests should fail when I make several changes to rds and s3 templates. I don't know whether that belongs in this ticket or not though

gabelton:
I've created a ticket for adding more unit tests here https://uktrade.atlassian.net/browse/DBTP-338

WillGibson (approved):

PR #108, opened 2023-07-21
Make AWS profile not configured error more helpful

WillGibson:
I got this error and I had no idea what it meant to me, so I made it a little more informative.

### DIFF
--- a/commands/utils.py
+++ b/commands/utils.py
@@ -133,7 +133,7 @@ def check_aws_conn(aws_profile: str) -> boto3.session.Session:
     try:
         session = boto3.session.Session(profile_name=aws_profile)
     except botocore.exceptions.ProfileNotFound:
-        click.secho("AWS profile not configured, please ensure they are set.", fg="red")
+        click.secho(f"""AWS profile "{aws_profile}" is not configured.""", fg="red")
         exit()
     except botocore.errorfactory.UnauthorizedException:
         click.secho(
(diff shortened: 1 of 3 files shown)


lgarvey (approved):

commands/utils.py
```diff
@@ -133,7 +133,7 @@ def check_aws_conn(aws_profile: str) -> boto3.session.Session:
     try:
         session = boto3.session.Session(profile_name=aws_profile)
     except botocore.exceptions.ProfileNotFound:
-        click.secho("AWS profile not configured, please ensure they are set.", fg="red")
+        click.secho(f"""AWS profile "{aws_profile}" is not configured.""", fg="red")
```
yusufsheiqh, line 136: Do we want to keep the `ensure they are set` bit?

WillGibson: I don't think it adds any value, telling us it's not configured is enough.

lgarvey: There's probably a wider discussion about how credentials are set.  Boto3 will attempt to source creds from multiple places, so we should probably let it do that.  But not something you need to consider now.

```
Boto3 will look in several locations when searching for credentials. The mechanism in which Boto3 looks for credentials is to search through a list of possible locations and stop as soon as it finds credentials. The order in which Boto3 searches for credentials is:

Passing credentials as parameters in the boto.client() method
Passing credentials as parameters when creating a Session object
Environment variables
Shared credential file (~/.aws/credentials)
AWS config file (~/.aws/config)
Assume Role provider
Boto2 config file (/etc/boto.cfg and ~/.boto)
Instance metadata service on an Amazon EC2 instance that has an IAM role configured.
```

yusufsheiqh (approved):

PR #109, opened 2023-07-21
Add region exports to .envrc.sample

WillGibson:


.envrc.sample
```diff
@@ -1 +1,3 @@
-export AWS_PROFILE=tools
\ No newline at end of file
+export AWS_PROFILE=tools
+export AWS_REGION=eu-west-2
+export AWS_DEFAULT_REGION=eu-west-2
```
WillGibson, line 3: At least `assign-domain` seems to need `AWS_DEFAULT_REGION`.

yusufsheiqh (approved):

PR #110, opened 2023-07-24
DBTP-337 Move to a command in copilot-tools

gabelton:
I've bundled together two tickets because it seemed easier to write unit tests and refactor some of the original script after shifting it all over to copilot tools:

https://uktrade.atlassian.net/browse/DBTP-337
https://uktrade.atlassian.net/browse/DBTP-328

The original POC code is [here](https://github.com/uktrade/copilot-tunnel/blob/master/util.py). I've refactored it a bit, where possible, to use the boto client instead of `popen` .


commands/conduit_cli.py
```diff
@@ -0,0 +1,84 @@
...
+    # so we need to wait before running the exec command or timeout if taking too long.
+    timeout = time.time() + 60
+    connected = False
+    while time.time() < timeout:
+        if is_task_running(cluster_arn):
+            os.system(f"copilot task exec --app {app} --env {env}")
+            connected = True
+            break
+
+    if connected == False:
```
WillGibson, line 56:
```suggestion
    if not connected:
```

pyproject.toml
```diff
@@ -30,6 +30,7 @@ python = "^3.9"
 schema = "0.7.5"
 cfn-flip = "1.3.0"
 moto = {extras = ["all"], version = "^4.1.12"}
+freezegun = "^1.2.2"
```
yusufsheiqh, line 33: Can we add this as a `dev` dependency?

commands/conduit_cli.py
gabelton: I think what I have currently is more explicit. I don't want to return an empty string or list and, if I did by accident, I'd want the test to complain

WillGibson: That's fair.

I only just found out because of this that `==` seems to be an exact comparison in Python, equivalent to `===` in most other languages I have used.

I wonder if it's worth type hinting this code?

gabelton: Ah yep, I forgot to add type hinting. Will do that now, although I've just realised it won't help in this particular instance because the function doesn't return anything

WillGibson: In fact, as my brain is starting to spin up better, I've realised that `connected` can only ever be  `True` or `False` in this context, so maybe my attempt to suggest something more "pythonic" is still valid?

commands/conduit_cli.py
```diff
@@ -0,0 +1,85 @@
...
+
+@click.group()
+def conduit():
+    pass
+
+
+@conduit.command()
+@click.option("--project-profile", required=True, help="aws account profile name")
+@click.option("--app", help="aws app name", required=True)
+@click.option("--env", help="aws environment name", required=True)
```
yusufsheiqh, line 71:
```suggestion
@click.option("--project-profile", required=True, help="AWS account profile name")
@click.option("--app", help="AWS application name", required=True)
@click.option("--env", help="AWS environment name", required=True)
```

tests/test_conduit_cli.py
```diff
@@ -0,0 +1,197 @@
...
+
+    arn = boto3.client("ecs").create_cluster(
+        tags=[
+            {"key": "copilot-application", "value": "dbt-app"},
+            {"key": "copilot-environment", "value": "staging"},
+            {"key": "aws:cloudformation:logical-id", "value": "Cluster"},
+        ]
+    )["cluster"]["clusterArn"]
+
+    assert get_cluster_arn("dbt-app", "staging") == arn
```
WillGibson, line 36: This would be easier to understand if it was broken into two stages...
```suggestion
    mocked_cluster = boto3.client("ecs").create_cluster(
        tags=[
            {"key": "copilot-application", "value": "dbt-app"},
            {"key": "copilot-environment", "value": "staging"},
            {"key": "aws:cloudformation:logical-id", "value": "Cluster"},
        ]
    )
    expected_arn = mocked_cluster["cluster"]["clusterArn"]

    assert get_cluster_arn("dbt-app", "staging") == expected_arn
```

tests/test_conduit_cli.py
```diff
@@ -0,0 +1,197 @@
...
+@mock_secretsmanager
+def test_get_postgres_secret_arn():
+    """Test that, given app and environment strings, get_postgres_secret_arn
+    returns the app's secret arn string."""
+
+    arn = boto3.client("secretsmanager").create_secret(
+        Name="/copilot/dbt-app/staging/secrets/POSTGRES", SecretString="secretivestring"
+    )["ARN"]
+
+    assert get_postgres_secret_arn("dbt-app", "staging") == arn
```
WillGibson, line 61:
```suggestion
    mocked_secret = boto3.client("secretsmanager").create_secret(
        Name="/copilot/dbt-app/staging/secrets/POSTGRES", SecretString="secretivestring"
    )
    expected_arn = mocked_secret["ARN"]

    assert get_postgres_secret_arn("dbt-app", "staging") == expected_arn
```

tests/test_conduit_cli.py
```diff
@@ -0,0 +1,197 @@
...
+    ecs_client.register_container_instance(cluster=arn, instanceIdentityDocument=instance_id_document)
+    definition_arn = ecs_client.register_task_definition(
+        family="copilot-dbtunnel",
+        containerDefinitions=[
+            {"name": "test_container", "image": "test_image", "cpu": 100, "memory": 500, "essential": True}
+        ],
+    )["taskDefinition"]["taskDefinitionArn"]
+    ecs_client.run_task(cluster=arn, taskDefinition=definition_arn)
+
+    assert is_task_running(arn)
```
WillGibson, line 98: @yusufsheiqh and I were looking at this together and as part of getting to grips with what it's doing, we kind of ended up doing a bit of a rewrite in our IDE.

in summary, the test was asserting on two, opposite, outcomes, so we split it into two tests, one for the expected `False` and one for the expected `True`.

Then we renamed some things to make it more explicit that we are working with mocked things and separated the tests into Arrange, Act and Assert blocks.

The result follows in massive suggestion!

```suggestion
@mock_ecs
def test_is_task_running_when_task_is_not_running():
    """Given an ECS Cluster ARN string, is_task_running should return False when the task is not running."""

    mocked_ecs_client = boto3.client("ecs")
    mocked_cluster_arn = mocked_ecs_client.create_cluster(
        tags=[
            {"key": "copilot-application", "value": "dbt-app"},
            {"key": "copilot-environment", "value": "staging"},
            {"key": "aws:cloudformation:logical-id", "value": "Cluster"},
        ]
    )["cluster"]["clusterArn"]

    assert not is_task_running(mocked_cluster_arn)


@mock_ec2
@mock_ecs
def test_is_task_running():
    """Given an ECS Cluster ARN string, is_task_running should return True when the task is running."""

    # Create mocked ECS Cluster
    mocked_ecs_client = boto3.client("ecs")
    mocked_cluster_arn = mocked_ecs_client.create_cluster(
        tags=[
            {"key": "copilot-application", "value": "dbt-app"},
            {"key": "copilot-environment", "value": "staging"},
            {"key": "aws:cloudformation:logical-id", "value": "Cluster"},
        ]
    )["cluster"]["clusterArn"]
    # Create mocked EC2 instance
    mocked_ec2_client = boto3.client("ec2")
    mocked_ec2_images = mocked_ec2_client.describe_images(Owners=["amazon"])["Images"]
    mocked_ec2_client.run_instances(ImageId=mocked_ec2_images[0]["ImageId"], MinCount=1, MaxCount=1)
    mocked_ec2_instances = boto3.client("ec2").describe_instances()
    mocked_ec2_instance_id = mocked_ec2_instances["Reservations"][0]["Instances"][0]["InstanceId"]
    mocked_ec2 = boto3.resource("ec2")
    mocked_ec2_instance = mocked_ec2.Instance(mocked_ec2_instance_id)
    mocked_instance_id_document = json.dumps(ec2_utils.generate_instance_identity_document(mocked_ec2_instance))
    # Attach mocked EC2 instance to the mocked ECS Cluster
    mocked_ecs_client.register_container_instance(cluster=mocked_cluster_arn, instanceIdentityDocument=mocked_instance_id_document)
    mocked_task_definition_arn = mocked_ecs_client.register_task_definition(
        family="copilot-dbtunnel",
        containerDefinitions=[
            {"name": "test_container", "image": "test_image", "cpu": 100, "memory": 500, "essential": True}
        ],
    )["taskDefinition"]["taskDefinitionArn"]

    mocked_ecs_client.run_task(cluster=mocked_cluster_arn, taskDefinition=mocked_task_definition_arn)

    assert is_task_running(mocked_cluster_arn)
```

WillGibson: We were also wondering about extracting things like the recurring `mocked_ecs_client.create_cluster...` pieces out to helper methods to reduce duplication and perhaps make the test scenarios more readable.

tests/conftest.py
```diff
@@ -57,3 +58,15 @@ def alias_session(aws_credentials):
...
+@pytest.fixture(scope="function")
+def mocked_cluster():
+    with mock_ecs():
+        yield boto3.client("ecs").create_cluster(
+            tags=[
+                {"key": "copilot-application", "value": "dbt-app"},
+                {"key": "copilot-environment", "value": "staging"},
+                {"key": "aws:cloudformation:logical-id", "value": "Cluster"},
+            ]
+        )
```
WillGibson, line 72: ❤️

tests/test_conduit_cli.py
```diff
@@ -0,0 +1,179 @@
...
+    expected_arn = mocked_secret["ARN"]
+
+    assert get_postgres_secret_arn("dbt-app", "staging") == expected_arn
+
+
+def test_is_task_running_when_task_is_not_running(mocked_cluster):
+    """Given an ECS Cluster ARN string, is_task_running should return False when
+    the task is not running."""
+
+    assert not is_task_running(mocked_cluster["cluster"]["clusterArn"])
```
WillGibson, line 62: This is so clean now :-)

tests/test_conduit_cli.py
```diff
@@ -0,0 +1,179 @@
...
+    system.assert_called_once_with("copilot task exec --app dbt-app --env staging")
+
+
+@freeze_time("Jan 14th, 2020", auto_tick_seconds=60)
+def test_exec_into_task_timeout(capsys):
+    exec_into_task("dbt-app", "staging", "arn:random")
+
+    assert (
+        capsys.readouterr().out
+        == "Attempt to exec into running task timed out. Try again by running `copilot task exec --app dbt-app --env staging or check status of task in Amazon ECS console.\n"
```
WillGibson, line 114: Has the exec into the task really timed out? Or is the task just not running?

tests/test_conduit_cli.py
```diff
@@ -0,0 +1,179 @@
...
+
+@mock_ecs
+@mock_resourcegroupstaggingapi
+@mock_sts
+def test_tunnel_no_cluster_resource(alias_session):
+    """Test that tunnel command prints an exit code when a cluster with app and
+    env tags is not found."""
+
+    runner = CliRunner()
+    result = runner.invoke(tunnel, ["--project-profile", "foo", "--app", "dbt-app", "--env", "staging"])
```
WillGibson, line 126: I'd be inclined to inline `CliRunner` for these simple use cases.

tests/test_conduit_cli.py
```diff
@@ -0,0 +1,179 @@
...
+
+    assert result.exit_code == 0
+    assert "No cluster resource found with tag filter values dbt-app and staging" in result.output
+
```
WillGibson, line 130: How is the AWS profile `foo` set/configured for the test above, but not the one below?

yusufsheiqh (approved):

tests/test_conduit_cli.py
```diff
@@ -0,0 +1,179 @@
...
+@mock_sts
+def test_tunnel_profile_not_configured():
+    """Test that tunnel calls check_aws_conn and outputs expected error when AWS
+    profile variable isn't set."""
+
+    runner = CliRunner()
+    result = runner.invoke(tunnel, ["--project-profile", "foo", "--app", "dbt-app", "--env", "staging"])
+
+    assert 'AWS profile "foo" is not configured.' in result.output
```
WillGibson, line 140: The error message in the assertion does not seem to align with the one in the docstring.

tests/test_conduit_cli.py
```diff
@@ -0,0 +1,179 @@
...
+def test_tunnel_task_not_running(create_task, exec_into_task, alias_session, mocked_cluster):
+    """Test that, when a task is not already running, command creates and execs
+    into a task."""
+
+    cluster_arn = mocked_cluster["cluster"]["clusterArn"]
+    secret_arn = boto3.client("secretsmanager").create_secret(
+        Name="/copilot/dbt-app/staging/secrets/POSTGRES", SecretString="secretivestring"
+    )["ARN"]
+    runner = CliRunner()
+    runner.invoke(tunnel, ["--project-profile", "foo", "--app", "dbt-app", "--env", "staging"])
```
WillGibson, line 157: Just to separate the arranging and the action...
```suggestion
    
    runner.invoke(tunnel, ["--project-profile", "foo", "--app", "dbt-app", "--env", "staging"])
```

tests/test_conduit_cli.py
```diff
@@ -0,0 +1,179 @@
...
+@patch("commands.conduit_cli.is_task_running", return_value=True)
+@patch("commands.conduit_cli.exec_into_task")
+@patch("commands.conduit_cli.create_task")
+def test_tunnel_task_already_running(create_task, exec_into_task, is_task_running, alias_session, mocked_cluster):
+    """Test that, when a task is already running, command execs into this task
+    and does not create a new one."""
+
+    cluster_arn = mocked_cluster["cluster"]["clusterArn"]
+    runner = CliRunner()
+    runner.invoke(tunnel, ["--project-profile", "foo", "--app", "dbt-app", "--env", "staging"])
```
WillGibson, line 176: Just to separate the arranging and the action (plus a little inlining)...
```suggestion
    cluster_arn = mocked_cluster["cluster"]["clusterArn"]
    
    CliRunner().invoke(tunnel, ["--project-profile", "foo", "--app", "dbt-app", "--env", "staging"])
```

WillGibson (commented):
Some more comments, but this is a great piece of work @gabelton.

tests/test_conduit_cli.py
yusufsheiqh: Agree with this.

tests/test_conduit_cli.py
gabelton: The way I've used freezegun means that it doesn't hit the logic where it checks whether the task is running. It just sees that `time.time() >= timeout` and so `connected` is never set to True. I could remove the freezegun logic and it would eventually timeout because the task isn't running, but I don't want the test to take 60 seconds

tests/test_conduit_cli.py
gabelton: I think they mean the same thing, but you're right that the wording could be clearer

tests/test_conduit_cli.py
WillGibson: So does the test start at `2020-01-14 00:00:00` and the timeout period starts then, after which it somehow zooms forward 60 seconds?

gabelton: yeah it's the `auto_tick_seconds` bit that zooms it forward

WillGibson: 🪄

tests/test_conduit_cli.py
WillGibson: Maybe just remove "variable" from the docstring?

PR #112, opened 2023-07-25
DBTP-343   Handle hyphens in application names

WillGibson:


WillGibson:
Meant to create draft

PR #113, opened 2023-07-25
DBTP-343 Handle hyphens in application names when getting Elastic Load Balancer details

WillGibson:
To satisfy https://uktrade.atlassian.net/browse/DBTP-343

See comments inline.

### DIFF
--- a/commands/dns_cli.py
+++ b/commands/dns_cli.py
@@ -321,7 +321,6 @@ def check_r53(domain_session, project_session, domain, base_domain):
     return cert_arn
 
 
-# Todo: Maybe rename method to something which describes its function?
 def get_elastic_load_balancer_domain_and_configuration(
     project_session: Session, app: str, svc: str, env: str
 ) -> Tuple[str, dict]:
--- a/commands/dns_cli.py
+++ b/commands/dns_cli.py
@@ -355,12 +355,11 @@ def lb_domain(project_session: Session, app: str, svc: str, env: str) -> Tuple[s
     check_response(response)
     no_items = True
     for service_arn in response["serviceArns"]:
-        service_name = service_arn.split("/")[2]
-        service_name_items = separate_hyphenated_application_environment_and_service(service_name)
-        service_app = service_name_items[0]
-        service_env = service_name_items[1]
-        service_service = service_name_items[2]
-        if service_app == app and service_env == env and service_service == svc:
+        fully_qualified_service_name = service_arn.split("/")[2]
+        service_app, service_env, service_name = separate_hyphenated_application_environment_and_service(
+            fully_qualified_service_name
+        )
+        if service_app == app and service_env == env and service_name == svc:
             no_items = False
             break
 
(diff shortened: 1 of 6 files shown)


tests/test_dns_cli.py
```diff
@@ -61,7 +62,7 @@ def test_create_cert(wait_for_certificate_validation, mock_click, acm_session, r
     with stubber:
         acm_session.describe_certificate(CertificateArn="arn:1234567890123456789")
 
-    assert create_cert(acm_session, route53_session, "test.1234", 1).startswith("arn:aws:acm:") == True
+    assert create_cert(acm_session, route53_session, "test.1234", 1).startswith("arn:aws:acm:")
```
WillGibson, line 65: It's more readable without the `== True`s.

commands/dns_cli.py
```diff
@@ -321,15 +321,22 @@ def check_r53(domain_session, project_session, domain, base_domain):
     return cert_arn
 
 
-def lb_domain(project_session: Session, app: str, svc: str, env: str) -> Tuple[str, dict]:
+# Todo: Maybe rename method to something which describes its function?
+def get_elastic_load_balancer_domain_and_configuration(
```
WillGibson, line 325: Renamed to try and better explain the method's expected behaviour.

commands/dns_cli.py
```diff
@@ -321,15 +321,22 @@ def check_r53(domain_session, project_session, domain, base_domain):
...
+    project_session: Session, app: str, svc: str, env: str
+) -> Tuple[str, dict]:
+    def separate_hyphenated_application_environment_and_service(hyphenated_string):
```
WillGibson, line 328: As it stands this method is just to increase readability and reduce duplication in it's parent method. 

I put it in here so that it's effectively a private method (for now anyway) and doesn't need to be tested directly.

tests/conftest.py
```diff
@@ -18,6 +18,21 @@
...
+
+
+@pytest.fixture
+def alphanumeric_environment_name() -> str:
+    return "alphanumericenvironmentname123"
+
+
+@pytest.fixture
+def alphanumeric_service_name() -> str:
+    return "alphanumericservicename123"
```
WillGibson, line 33: Created these as fixtures so it's easier to change later if we decide we want different behaviour.

commands/dns_cli.py
lgarvey: technically it's not an elastic load balancer, it's an application load balancer.

tests/conftest.py
lgarvey: It'd make more sense to just put some constants at the top of the file.

commands/dns_cli.py
WillGibson: Done

tests/conftest.py
WillGibson: I had wondered about that myself. What's our criteria for this? If it's a scalar type use a constant, more complicated use a fixture?

yusufsheiqh: @WillGibson That makes sense to me. I think these would be constants in this file and if used elsewhere, we put them in the `conftest.py` file.

WillGibson: They are constants now.

PR #116, opened 2023-07-26
DBTP-340 Add buildspec file for creating conduit task image

gabelton:
Follow up to previous PR, adding buildspec file for creating conduit task image

Also fixing an unrelated [security vulnerability](https://github.com/uktrade/copilot-tools/security/dependabot/5)

### DIFF
--- a/Dockerfile.conduit
+++ b/Dockerfile.conduit
@@ -1,0 +1,2 @@
+FROM public.ecr.aws/docker/library/postgres:latest
+ENTRYPOINT ["tail", "-f", "/dev/null"]
--- a/Dockerfile.conduit
+++ b/Dockerfile.conduit
@@ -1,2 +1,1 @@
 FROM public.ecr.aws/docker/library/postgres:latest
-ENTRYPOINT ["tail", "-f", "/dev/null"]
--- a/buildspec-ecr.yml
+++ b/buildspec-ecr.yml
@@ -1,0 +1,18 @@
+version: 0.2
+
+phases:
+  pre_build:
+    commands:
+      - echo Logging in to Amazon ECR...
+      - aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/uktrade
+  build:
+    commands:
+      - echo Build started on `date`
+      - echo Building the Docker image...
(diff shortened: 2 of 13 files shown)


pyproject.toml
```diff
@@ -31,6 +31,7 @@ schema = "0.7.5"
 cfn-flip = "1.3.0"
 moto = {extras = ["all"], version = "^4.1.12"}
 aiohttp = "^3.8.4"
+certifi = "^2023.07.22"
```
WillGibson, line 34: Why is this added? It doesn't seem relevant to the other changes to me.

WillGibson, line 34: Why is this added? It doesn't seem relevant to the other changes to me.

Dockerfile.conduit
```diff
@@ -0,0 +1,2 @@
+FROM public.ecr.aws/docker/library/postgres:latest
+ENTRYPOINT ["tail", "-f", "/dev/null"]
```
yusufsheiqh, line 2: What is the purpose of this? This can be removed as we will be building an image and pushing to the ECR.

```suggestion

```

buildspec-ecr.yml
```diff
@@ -0,0 +1,18 @@
...
+    commands:
+      - echo Build started on `date`
+      - echo Building the Docker image...
+      - docker build -t tunnel:latest --file Dockerfile.conduit .
+      - docker tag tunnel:latest public.ecr.aws/uktrade/tunnel:latest
+  post_build:
+    commands:
+      - echo Build completed on `date`
+      - echo Pushing the Docker image...
+      - docker push public.ecr.aws/uktrade/tunnel:latest
```
WillGibson, line 18: Can we add the line break at the end to stop GitHub moaning?

You can probably automate this on save in your IDE.

pyproject.toml
WillGibson: Ah, I may have connected that if I was more intelligent, but as I'm not, maybe explicitly mentioning `certifi` in the description would have saved me starting this thread.

buildspec-ecr.yml
yusufsheiqh, line 1: I reckon we can rename this file to `buildspec.conduit.yml` or something similar.

yusufsheiqh (approved):

PR #119, opened 2023-07-27
DBTP-343 Correct Handling of hyphens when getting Load Balancer details

WillGibson:
For https://uktrade.atlassian.net/browse/DBTP-343

I should have tested my code manually, there were extra hyphen separated things on the end of the cluster and service ARNs in the real world which are not included in the unit tests.

### DIFF
--- a/commands/dns_cli.py
+++ b/commands/dns_cli.py
@@ -328,8 +328,8 @@ def get_load_balancer_domain_and_configuration(
         hyphenated_string, number_of_items_of_interest, number_of_trailing_items
     ):
         # The application name may be hyphenated, so we start splitting
-        # at the hyphen before the trailing items and return the
-        # first items of interest only...
+        # at the hyphen after the first item of interest and return the
+        # items of interest only...
         rsplit = hyphenated_string.rsplit("-", number_of_trailing_items + number_of_items_of_interest - 1)
         interest = rsplit[:number_of_items_of_interest]
         return interest
--- a/commands/dns_cli.py
+++ b/commands/dns_cli.py
@@ -330,9 +330,9 @@ def get_load_balancer_domain_and_configuration(
         # The application name may be hyphenated, so we start splitting
         # at the hyphen after the first item of interest and return the
         # items of interest only...
-        rsplit = hyphenated_string.rsplit("-", number_of_trailing_items + number_of_items_of_interest - 1)
-        interest = rsplit[:number_of_items_of_interest]
-        return interest
+        return hyphenated_string.rsplit("-", number_of_trailing_items + number_of_items_of_interest - 1)[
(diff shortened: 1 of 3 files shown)


commands/dns_cli.py
```diff
@@ -324,9 +324,15 @@ def check_r53(domain_session, project_session, domain, base_domain):
...
-    def separate_hyphenated_application_environment_and_service(hyphenated_string):
-        # The application name may be hyphenated, so we start splitting at the second from last occurrence of a hyphen
-        return hyphenated_string.rsplit("-", 2)
+    def separate_hyphenated_application_environment_and_service(
+        hyphenated_string, number_of_items_of_interest, number_of_trailing_items
+    ):
+        # The application name may be hyphenated, so we start splitting
+        # at the hyphen after the first item of interest and return the
+        # items of interest only...
+        rsplit = hyphenated_string.rsplit("-", number_of_trailing_items + number_of_items_of_interest - 1)
```
yusufsheiqh, line 333: Can we rename this variable to something else?

tests/test_dns_cli.py
```diff
@@ -183,7 +183,7 @@ def test_get_load_balancer_domain_and_configuration_no_clusters(capfd):
 @mock_ecs
 def test_get_load_balancer_domain_and_configuration_no_services(capfd):
     boto3.Session().client("ecs").create_cluster(
-        clusterName=f"{HYPHENATED_APPLICATION_NAME}-{ALPHANUMERIC_ENVIRONMENT_NAME}-{ALPHANUMERIC_SERVICE_NAME}"
+        clusterName=f"{HYPHENATED_APPLICATION_NAME}-{ALPHANUMERIC_ENVIRONMENT_NAME}-Cluster-c0PIlotiD3ntIF3r"
```
yusufsheiqh, line 186: Shall we create a constant for `Cluster-c0PIlotiD3ntIF3r` as well? I see it is used in another test as well.

commands/dns_cli.py
gabelton, line 333: Not a blocker, but these variable names are a bit of a mouthful

gabelton (approved):

commands/dns_cli.py
WillGibson: These were only meant to be temporary for breakpoints while I figured it out. Gone now.

WillGibson: But you don't have to think about what they mean ;-)

I'm open to suggestions for conveying the same meaning in less words.

tests/test_dns_cli.py
WillGibson: I extracted a couple more while I was at it. Less for duplication reasons, that readability.

yusufsheiqh (approved):

PR #120, opened 2023-07-28
DBTP-344 Expedient pass at making `copilot-helper domain check-domain` work

WillGibson:
For https://uktrade.atlassian.net/browse/DBTP-344, which I think we can maybe defer for today, but need to address pretty soon.

Address some issues we hit when using the `check-domain` command to create a new environment.

Increases the "helpfulness" of the output.

No tests have failed as a result of these changes 😬 

This is in no way the whole of the work required for the ticket, just an expedient attempt to unblock us deploying a new environment while @WillGibson got a lot more familiar with the workings of dns_cli.py than he was before 🤣

### DIFF
--- a/commands/dns_cli.py
+++ b/commands/dns_cli.py
@@ -100,9 +100,10 @@ def create_cert(client, domain_client, domain, base_len):
         response = client.describe_certificate(CertificateArn=arn)
 
     cert_record = response["Certificate"]["DomainValidationOptions"][0]["ResourceRecord"]
-    response = domain_client.list_hosted_zones_by_name()
 
+    click.secho(f"Looking fo ID of domain {domain_to_create}...", fg="yellow")
     domain_id = False
+    response = domain_client.list_hosted_zones_by_name()
     for hz in response["HostedZones"]:
         if hz["Name"] == domain_to_create:
             domain_id = hz["Id"]
@@ -273,7 +274,7 @@ def create_hosted_zone(client, domain, start_domain, base_len):
             ChangeBatch={
                 "Changes": [
                     {
-                        "Action": "UPSERT",
+                        "Action": "CREATE",
                         "ResourceRecordSet": {
                             "Name": subdom,
                             "Type": "NS",
--- a/commands/dns_cli.py
+++ b/commands/dns_cli.py
@@ -23,21 +23,29 @@ MAX_DOMAIN_DEPTH = 2
 AWS_CERT_REGION = "eu-west-2"
 
 
-def wait_for_certificate_validation(acm_client, certificate_arn, sleep_time=5, timeout=600):
-    click.secho("waiting for cert...", fg="yellow")
+def wait_for_certificate_validation(acm_client, certificate_arn, sleep_time=10, timeout=600):
+    click.secho(f"Waiting up to {timeout} seconds for certificate to be validated...", fg="yellow")
     status = acm_client.describe_certificate(CertificateArn=certificate_arn)["Certificate"]["Status"]
     elapsed_time = 0
     while status == "PENDING_VALIDATION":
-        if elapsed_time > timeout:
-            raise Exception(f"Timeout ({timeout}s) reached for certificate validation")
+        if elapsed_time >= timeout:
+            raise Exception(
+                f"Timeout ({timeout}s) reached for certificate validation, might be worth checking things in the AWS Console"
+            )
         click.echo(
             click.style(f"{certificate_arn}", fg="white", bold=True)
-            + click.style(f": Waiting {sleep_time}s for validation, {elapsed_time}s elapsed...", fg="yellow"),
+            + click.style(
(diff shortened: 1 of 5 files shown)


commands/dns_cli.py
```diff
@@ -98,6 +98,9 @@ def create_cert(client, domain_client, domain, base_len):
         if hz["Name"] == domain_to_create:
             domain_id = hz["Id"]
             break
+    if not domain_id:
+        click.secho(f"Unable to find Domain ID for {domain_to_create} in the hosted zones")
```
yusufsheiqh, line 102: Do we want styling here? Maybe a red colour... If we do not want colour then we should be using `click.echo()` rather than `click.secho()`.

```suggestion
        click.secho(f"Unable to find Domain ID for {domain_to_create} in the hosted zones", fg="red")
```

commands/dns_cli.py
```diff
@@ -448,8 +453,10 @@ def check_domain(domain_profile, project_profile, base_domain):
                             click.secho("Environment: " + env + "\t=> Domain: " + domain["http"]["alias"], fg="yellow")
                             cert_arn = check_r53(domain_session, project_session, domain["http"]["alias"], base_domain)
                             cert_list.update({domain["http"]["alias"]: cert_arn})
+                            click.echo(" ")
```
yusufsheiqh, line 456: What is this for?

WillGibson: Makes a line break after each environment's block of output. Helps with grokking.

commands/dns_cli.py
WillGibson: It is now.

migration-tools/build-copilot-config.py
```diff
@@ -85,8 +85,6 @@ def space_to_copilot_app(app_name, ns_conf):
             continue
 
         web_processes = [p for p in processes if p["type"] == "web"]
-        if len(web_processes) != 1:
-            breakpoint()
```
WillGibson, line 89: I think this was a stray from someone's work.

commands/dns_cli.py
```diff
@@ -195,7 +214,7 @@ def check_for_records(client, parent_id, subdom, subdom_id):
         if subdom in records["Name"]:
             click.secho(records["Name"] + " found", fg="green")
             records_to_add.append(records)
-            add_records(client, records, subdom_id)
+            add_records(client, records, subdom_id, "UPSERT")
```
WillGibson, line 217: This was the line that started it all!

commands/dns_cli.py
```diff
@@ -266,7 +287,7 @@ def create_hosted_zone(client, domain, start_domain, base_len):
             },
         )
 
-        return True
+    return True
```
WillGibson, line 290: This was causing the function to return after doing the first hosted zone, which was problematic if more than one part of the chain needs to be created.

commands/dns_cli.py
WillGibson: It's going away now.

commands/dns_cli.py
```diff
@@ -23,21 +23,30 @@
...
         )
         time.sleep(sleep_time)
         status = acm_client.describe_certificate(CertificateArn=certificate_arn)["Certificate"]["Status"]
         elapsed_time += sleep_time
-    click.secho("cert validated...", fg="green")
+
+    if status == "ISSUED":
+        click.secho("Certificate validated...", fg="green")
+    else:
+        raise Exception(f"""Certificate validation failed with the status "{status}".""")
```
WillGibson, line 49: It could be something other than ISSUED after PENDING_VALIDATION, I figured it would be helpful to know, and to stop execution at that point.

commands/dns_cli.py
```diff
@@ -92,19 +101,27 @@ def create_cert(client, domain_client, domain, base_len):
...
+    domain_id = False
+    response = domain_client.list_hosted_zones_by_name()
     for hz in response["HostedZones"]:
         if hz["Name"] == domain_to_create:
             domain_id = hz["Id"]
             break
+    if not domain_id:
+        # Will got here more than once during manual testing, it might be a race condition we need to handle better
+        click.secho(f"Unable to find Domain ID for {domain_to_create} in the hosted zones", fg="red", bold=True)
+        exit(1)
```
WillGibson, line 115: Previously it could carry on past here without any clue that the result of it's labours were not going to work 😆

commands/dns_cli.py
yusufsheiqh: Resolved in https://github.com/uktrade/copilot-tools/pull/120/commits/2d0bcd9cceb3ed36678e238e1eef1c1a1bbcedbe.

commands/dns_cli.py
ejayesh: good catch

ejayesh (approved):
Nice work.  Approved

PR #121, opened 2023-08-01
Update `copy-secrets` command

yusufsheiqh:
## Context

- Update `copy-secrets` command to properly set tags
  - Display a message if the parameter/secret already exists in Parameter Store for the target environment.
  - Creates the parameter/secret only if it does not exist already.

## Additional information

This has been manually tested as well. We still need to test the utils method - is that part of this ticket or another?

Jira ticket - https://uktrade.atlassian.net/browse/DBTP-350

### DIFF
--- a/commands/bootstrap_cli.py
+++ b/commands/bootstrap_cli.py
@@ -298,6 +298,8 @@ def copy_secrets(project_profile, source_environment, target_environment):
                     f"""The "{secret_name.split("/")[-1]}" parameter already exists for the "{target_environment}" environment.""",
                     fg="yellow",
                 )
+            else:
+                raise e
 
 
 @bootstrap.command()
--- a/commands/bootstrap_cli.py
+++ b/commands/bootstrap_cli.py
@@ -5,6 +5,7 @@ from pathlib import Path
 
 import click
 import yaml
+from botocore.exceptions import ClientError
 from cloudfoundry_client.client import CloudFoundryClient
 from schema import Optional
 from schema import Schema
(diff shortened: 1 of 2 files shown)


commands/bootstrap_cli.py
```diff
@@ -278,18 +279,26 @@ def copy_secrets(project_profile, source_environment, target_environment):
...
+            set_ssm_param(
+                config["app"],
+                target_environment,
+                secret_name,
+                secret[1],
+                False,
+                False,
+                f"Copied from {source_environment} environment.",
+            )
+        except ClientError as e:
```
lgarvey, line 295: It could be worth adding the exception handling logic to the set_ssm_param function and raising something more specific?
```
raise ParameterExists("xxx")
```

It might save some duplication going forward.

Also don't you want to re-raise in an else statement?

yusufsheiqh: I was thinking about adding this to the `set_ssm_param` method actually, but that entire method is not tested ATM. Left it as a question in the description of this PR.

Not sure I want to re-raise. Is there any reason to do this?

WillGibson: I think re-raising the exception if the code is not `ParameterAlreadyExists` would be correct.

WillGibson: We are just using `set_ssm_param` here, I would ticket it up to get that under test separately so we don't hold up getting this value add published.

yusufsheiqh: That makes sense!

Resolved in https://github.com/uktrade/copilot-tools/pull/121/commits/19403a3365f839cdea15e47840ae15cd74e8dc03.

yusufsheiqh: Created a ticket to add unit tests for the `set_ssm_param` utils method here -> https://uktrade.atlassian.net/browse/DBTP-352.

WillGibson (approved):

PR #122, opened 2023-08-01
Update bootstrap configuration files

yusufsheiqh:
## Context

- Update bootstrap configuration files
  - These were slightly outdated when we were refreshing the `staff-sso` application, as we needed more secrets that were not there before.


antroy-madetech (approved):
LGTM

WillGibson (approved):
Ticking without detailed read as I believe these are auto generated.

PR #123, opened 2023-08-01
Use defaults for public service healthcheck timeout and interval

WillGibson:
As discussed in Slack, we will use the defaults and revisit this as required on a per application basis.

### DIFF
--- a/commands/templates/svc/manifest-public.yml
+++ b/commands/templates/svc/manifest-public.yml
@@ -23,8 +23,6 @@ http:
     success_codes: '200,301,302'
     healthy_threshold: 3
     unhealthy_threshold: 2
-    interval: 5s
-    timeout: 30s
     grace_period: 90s
 
 sidecars:
--- a/commands/templates/svc/manifest-public.yml
+++ b/commands/templates/svc/manifest-public.yml
@@ -30,12 +30,7 @@ sidecars:
     port: 443
     image: public.ecr.aws/uktrade/nginx-reverse-proxy:latest
     variables:
-      {% if ipfilter -%}
-      SERVER: localhost:8000
-      {% else -%}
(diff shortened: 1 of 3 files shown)


commands/templates/svc/manifest-public.yml
```diff
@@ -23,8 +23,6 @@ http:
     success_codes: '200,301,302'
     healthy_threshold: 3
     unhealthy_threshold: 2
-    interval: 5s
-    timeout: 30s
     grace_period: 90s
```
yusufsheiqh, line 26: What is the default for grace periods?

yusufsheiqh (approved):

commands/templates/svc/manifest-public.yml
WillGibson: Zero.

https://docs.aws.amazon.com/AmazonECS/latest/developerguide/service_definition_parameters.html#sd-networkconfiguration

tests/test_bootstrap_cli.py
```diff
@@ -109,17 +109,20 @@ def test_make_config(tmp_path):
...
 
+    def real_line_breaks(input: any) -> str:
+        return str(input).replace("\\n", "\n")
+
     with open(str(tmp_path / "copilot/.workspace")) as workspace:
         assert workspace.read() == "application: test-app"
 
     with open(str(tmp_path / "copilot/environments/test/manifest.yml"), "rb") as test:
-        assert test.read() == test_environment_manifest
+        assert real_line_breaks(test.read()) == real_line_breaks(test_environment_manifest)
```
WillGibson, line 119: Makes it easier to see what's different.

yusufsheiqh (approved):

PR #126, opened 2023-08-02
Add new ip filter config with appconfig container

lgarvey:
This PR adds the appconfig and ipfilter containers.  The new IP filter logic is always inline, but IP filtering can be disabled by setting the an env var: `IPFILTER_ENABLED=False`.  The IP filter can also be disabled on a per environment basis. When disabled the IP filter reverse proxies traffic to the service.

### DIFF
--- a/commands/templates/svc/manifest-public.yml
+++ b/commands/templates/svc/manifest-public.yml
@@ -34,9 +34,7 @@ sidecars:
     image: public.ecr.aws/uktrade/ip-filter:latest
     variables:
       PORT: 8000
-      EMAIL_NAME: DBT
-      EMAIL: sre-team@digital.trade.gov.uk
-      ORIGIN_HOSTNAME: localhost:8080
+      SERVER: localhost:8080
       APPCONFIG_PROFILES: ipfilter:default:default
       IPFILTER_ENABLED: {{ ipfilter }}
 
--- a/commands/templates/svc/manifest-public.yml
+++ b/commands/templates/svc/manifest-public.yml
@@ -34,7 +34,7 @@ sidecars:
     image: public.ecr.aws/uktrade/ip-filter:latest
     variables:
       PORT: 8000
-      EMAIL_NAME: DIT
+      EMAIL_NAME: DBT
       EMAIL: sre-team@digital.trade.gov.uk
       ORIGIN_HOSTNAME: localhost:8080
       APPCONFIG_PROFILES: ipfilter:default:default
--- a/commands/templates/svc/manifest-public.yml
+++ b/commands/templates/svc/manifest-public.yml
@@ -35,7 +35,7 @@ sidecars:
     variables:
       PORT: 8000
       EMAIL_NAME: DIT
-      EMAIL: sre@digital.trade.gov.uk
+      EMAIL: sre-team@digital.trade.gov.uk
       ORIGIN_HOSTNAME: localhost:8080
       APPCONFIG_PROFILES: ipfilter:default:default
       IPFILTER_ENABLED: {{ ipfilter }}
--- a/commands/templates/svc/manifest-public.yml
+++ b/commands/templates/svc/manifest-public.yml
@@ -15,9 +15,6 @@ http:
   # healthcheck: '/'
   target_container: nginx
   healthcheck:
-    {%- if ipfilter %}
-    {% else %}
-    {% endif -%}
     path: '/'
     port: 8080
     success_codes: '200,301,302'
@@ -30,25 +27,28 @@ sidecars:
     port: 443
     image: public.ecr.aws/uktrade/nginx-reverse-proxy:latest
     variables:
-      SERVER: localhost:{% if ipfilter %}8000{% else %}8080{% endif %}
+      SERVER: localhost:8000
 
-{% if ipfilter %}
   ipfilter:
     port: 8000
     image: public.ecr.aws/uktrade/ip-filter:latest
     variables:
       PORT: 8000
-      EMAIL_NAME: 'The Department for International Trade WebOps team'
-      EMAIL: test@test.test
-      LOG_LEVEL: DEBUG
+      EMAIL_NAME: DIT
+      EMAIL: sre@digital.trade.gov.uk
       ORIGIN_HOSTNAME: localhost:8080
-      ORIGIN_PROTO: http
-      CONFIG_FILE: 's3://ipfilter-config/ROUTES.yaml'
-{% endif %}
+      APPCONFIG_PROFILES: ipfilter:default:default
+      IPFILTER_ENABLED: {{ ipfilter }}
+
+  appconfig:
+    port: 2772
+    image: public.ecr.aws/aws-appconfig/aws-appconfig-agent:2.x
+    essential: true
+    variables:
+      ROLE_ARN: arn:aws:iam::763451185160:role/AppConfigIpFilterRole
 
 # Configuration for your containers and service.
 image:
-
   location: {{ image_location }}
     # Port exposed through your container to route traffic to it.
   port: 8080
@@ -80,7 +80,7 @@ variables:                    # Pass environment variables as key value pairs.
 secrets:                      # Pass secrets from AWS Systems Manager (SSM) Parameter Store.
   {%- for secret, value in secrets.items() %}
   {{ secret }}: /copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_ENVIRONMENT_NAME}/secrets/{{ value }}
-  {%- endfor -%}
+  {% endfor -%}
 {% else %}
 # secrets:                      # Pass secrets from AWS Systems Manager (SSM) Parameter Store.
 {% endif %}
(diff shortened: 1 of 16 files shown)


commands/templates/svc/manifest-public.yml
```diff
@@ -30,25 +27,28 @@ sidecars:
...
-{% if ipfilter %}
   ipfilter:
     port: 8000
     image: public.ecr.aws/uktrade/ip-filter:latest
     variables:
       PORT: 8000
-      EMAIL_NAME: 'The Department for International Trade WebOps team'
-      EMAIL: test@test.test
-      LOG_LEVEL: DEBUG
+      EMAIL_NAME: DIT
```
yusufsheiqh, line 37: We should probably go with DBT here.

```suggestion
      EMAIL_NAME: DBT
```

commands/templates/svc/manifest-public.yml
```diff
@@ -30,25 +27,28 @@ sidecars:
...
+      EMAIL: sre@digital.trade.gov.uk
```
yusufsheiqh, line 38: Shall we use the `sre-team` email address here?

```suggestion
      EMAIL: sre-team@digital.trade.gov.uk
```

commands/templates/svc/manifest-public.yml
```diff
@@ -30,25 +27,28 @@ sidecars:
...
-{% endif %}
+      APPCONFIG_PROFILES: ipfilter:default:default
+      IPFILTER_ENABLED: {{ ipfilter }}
+
+  appconfig:
+    port: 2772
+    image: public.ecr.aws/aws-appconfig/aws-appconfig-agent:2.x
+    essential: true
+    variables:
+      ROLE_ARN: arn:aws:iam::763451185160:role/AppConfigIpFilterRole
```
yusufsheiqh, line 48: What are the security implications of leaving the account ID?

commands/utils.py
```diff
@@ -130,6 +130,8 @@ def setup_templates():
             "public-manifest": templateEnv.get_template("svc/manifest-public.yml"),
             "backend-manifest": templateEnv.get_template("svc/manifest-backend.yml"),
             "s3-policy": templateEnv.get_template("addons/svc/s3-policy.yml"),
+            "s3": templateEnv.get_template("addons/svc/s3-policy.yml"),
```
yusufsheiqh, line 133: These two reference the same template - is this intentional?

lgarvey: Yes, I'm  using the contents of `templates["env"]` and `templates["svc"]` to determine if addons should be created at the svc or env level.  

```
storage_type = "s3"

if storage_type in templates["svc"]:
    # create service level template
```

I'm considering refactoring the template logic so that we use a template loader, ditch manual loading in `utils.get_templates()` and then have a dict or yaml file that determines which addons require which files, e.g. 

```
redis:
  env:
    - template: addons/env/redis.yml
    - template: addons/env/addons.parameters.yml
       filename: addons.parameters.yml
s3:
  env:
    - template: addons/env/s3.yml
  svc:
    - template: addons/svc/s3-policy.yml
s3-policy:
  svc:
    - template: addons/svc/s3-policy.yml
```

commands/templates/svc/manifest-public.yml
lgarvey: The account number alone is not directly exploitable, and the ARN should not be either - as long as it's configured correctly.  It's still not ideal to expose this though.

pyproject.toml
```diff
@@ -3,7 +3,7 @@ line-length = 119
 
 [tool.poetry]
 name = "dbt-copilot-tools"
-version = "0.1.15"
+version = "0.1.16"
```
yusufsheiqh, line 6: Looks like you're a couple of versions behind btw. Might need to rebase `main`.

https://github.com/uktrade/copilot-tools/blob/main/pyproject.toml#L6

lgarvey: This has been bumped to 0.1.19

commands/utils.py
yusufsheiqh: Will be resolved in https://github.com/uktrade/copilot-tools/pull/127.

tests/test_check_cloudformation.py
```diff
@@ -91,5 +91,5 @@ def test_outputs_failed_results_summary(patched_prepare_cloudformation_templates
     result = CliRunner().invoke(check_cloudformation_command)
 
     assert (
-        "The CloudFormation templates failed the following checks :-(\n  - lint" in result.output
+       "The CloudFormation templates passed the following checks :-)\n  - lint\n" in result.output
```
yusufsheiqh, line 94: Hmmm... is this meant to pass? Looks like this test is supposed to output the failure of the CloudFormation templates.

tests/test_bootstrap_cli.py
```diff
@@ -109,20 +109,17 @@ def test_make_config(tmp_path):
...
 
-    def real_line_breaks(input: any) -> str:
-        return str(input).replace("\\n", "\n")
-
     with open(str(tmp_path / "copilot/.workspace")) as workspace:
         assert workspace.read() == "application: test-app"
 
-    with open(str(tmp_path / "copilot/environments/test/manifest.yml"), "rb") as test:
-        assert real_line_breaks(test.read()) == real_line_breaks(test_environment_manifest)
+    with open(str(tmp_path / "copilot/environments/test/manifest.yml"), "r") as test:
```
WillGibson, line 115: "r" is the default, so can be removed from these calls.

tests/test_waf_cli.py
```diff
@@ -122,6 +124,7 @@ def test_custom_waf_file_not_found(alias_session):
     assert result.exit_code == 0
 
 
+@pytest.mark.skip("The WAF logic is broken and needs to be remediated in another PR")
```
WillGibson, line 127: Is that under the same ticket or can we reference the ticket here?

yusufsheiqh: I think this is the PR that resolves this logic issue -> https://github.com/uktrade/copilot-tools/pull/128.

tests/fixtures/test_service_manifest.yml
```diff
@@ -74,7 +75,9 @@ variables:                    # Pass environment variables as key value pairs.
 
 secrets:                      # Pass secrets from AWS Systems Manager (SSM) Parameter Store.
   TEST_SECRET: /copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_ENVIRONMENT_NAME}/secrets/TEST_SECRET
+
```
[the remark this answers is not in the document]
yusufsheiqh: Not desirable. It looks weird in the manifest files, IMO.

tests/test_check_cloudformation.py
yusufsheiqh: Interesting... was this test passing AKA failing the checks before because of W2001 and now it goes through because we added this check to the `--ignore-checks` flag?

tests/test_waf_cli.py
lgarvey: Gabe has solved this in another PR.  Either we get this PR merged and rebased or work the other way around.

tests/fixtures/test_service_manifest.yml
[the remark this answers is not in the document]
yusufsheiqh: Looks to be resolved in https://github.com/uktrade/copilot-tools/pull/126/commits/039fa9d6b375b12754196130b43629bb5fb1643c.

tests/test_bootstrap_cli.py
lgarvey: Done.

tests/test_waf_cli.py
WillGibson: I think whoever gets their pull request in first gets it in first, we have quite a few open now, need to start getting them merged.

WillGibson (dismissed):
Approving. Looks like you need to pull in the latest from the `main` branch though, hopefully no stress from that.

tests/test_check_cloudformation.py
WillGibson: How is this resolved? The test name still says "failed", but the assert says "passed".

tests/test_waf_cli.py
WillGibson: @gabelton won 🤣

tests/test_check_cloudformation.py
yusufsheiqh: @WillGibson Not sure if you are looking at the right commit, but the assert also says "failed"...

yusufsheiqh (approved):

tests/test_check_cloudformation.py
WillGibson: The test method name is `test_outputs_failed_results_summary`.

WillGibson: Oh, ignore. I think I just discovered that if you navigate to a file from one of these threads on the conversation tab it doesn't show you all commits by default. I wonder how much nonsense I've said because of that. Will approve again.

WillGibson (approved):

PR #132, opened 2023-08-04
DBTP 330 allow for connecting to RDS

gabelton:
Selection of bugfixes that I found while testing the conduit command against the RTO's RDS instance:
- refactoring `get_cluster_arn` because the resources api was showing deleted clusters
- adding logic to `is_task_running` to check that the task is not only running but ready for `exec` commands to be run against it
- removing some of the duplicated logic for getting secrets

Also a few tweaks to the RDS template, to make it more inline with aurora


commands/conduit_cli.py
```diff
@@ -11,36 +12,63 @@
 
 
 def get_cluster_arn(app: str, env: str) -> str:
-    client = boto3.client("resourcegroupstaggingapi")
+    ecs_client = boto3.client("ecs")
+    clusters_response = ecs_client.list_clusters()
```
yusufsheiqh, line 16: The `response` bit sounds redundant... I feel like this could be `ecs_clusters`.

```suggestion
    ecs_clusters_ = ecs_client.list_clusters()
```

lgarvey (approved):

yusufsheiqh (approved):

commands/conduit_cli.py
WillGibson: Could you go straight for this here?
```suggestion
    cluster_arns = ecs_client.list_clusters()["clusterArns"]
```

PR #136, opened 2023-08-14
login alias check

ejayesh:
Small change to check that Alias is set on the AWS account name


ejayesh:
Seems Tests are failing, not sure why as the PR shouldnt be changing the results. Investigating

PR #138, opened 2023-08-15
Created the missing happy path test in the TestMakeAddonCommand class

antroy-madetech:
Investigated the MakeAddon command for missing test coverage. The error cases seem to all be covered, however there was no happy path test that tested the actual generated files. This has been added.


commands/copilot_cli.py
```diff
@@ -16,6 +16,8 @@
 from .utils import mkfile
 from .utils import setup_templates
 
+ADDON_CONFIG_FILENAME = "addons.yml"
```
yusufsheiqh, line 19: Not sure we need this since it's only used in one place or do we foresee that we will be using this elsewhere?

tests/fixtures/make_addons/full/config/addons.yml
yusufsheiqh, line 1: The `full` folder seems a bit redundant. I wonder if we could create a `test-application` folder inside `fixtures` with all the files there instead?

tests/fixtures/make_addons/full/config/copilot/web/manifest.yml
```diff
@@ -0,0 +1,151 @@
...
+  readonly_fs: true       # Limit to read-only access to mounted root filesystems.
+
+# Optional fields for more advanced use-cases.
+#
+variables:                    # Pass environment variables as key value pairs.
+  PORT: 8080
+
+
+secrets:                      # Pass secrets from AWS Systems Manager (SSM) Parameter Store.
+  APP_ENV: /copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_ENVIRONMENT_NAME}/secrets/APP_ENV
```
yusufsheiqh, line 76: There are wayyy too many secrets here - might be worth cutting down on some of these in the `bootstrap.yml` file.

tests/fixtures/make_addons/full/config/copilot/web/manifest.yml
```diff
@@ -0,0 +1,151 @@
...
+  TWITTER_ACCESS_TOKEN: /copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_ENVIRONMENT_NAME}/secrets/TWITTER_ACCESS_TOKEN
+  TWITTER_DEPT_USER: /copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_ENVIRONMENT_NAME}/secrets/TWITTER_DEPT_USER
+  TWITTER_OAUTH_CONSUMER_KEY: /copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_ENVIRONMENT_NAME}/secrets/TWITTER_OAUTH_CONSUMER_KEY
+  TWITTER_OAUTH_CONSUMER_SECRET: /copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_ENVIRONMENT_NAME}/secrets/TWITTER_OAUTH_CONSUMER_SECRET
+  WAGTAIL_BASE_URL: /copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_ENVIRONMENT_NAME}/secrets/WAGTAIL_BASE_URL
+  X_AUTOSCALING_MAX: /copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_ENVIRONMENT_NAME}/secrets/X_AUTOSCALING_MAX
+
+
+# You can override any of the values defined above by environment.
+environments:
```
yusufsheiqh, line 139: I think we can also remove some of the environments as well... maybe one prod. environment and one non-prod. environment.

tests/test_copilot_cli.py
```diff
@@ -58,6 +61,37 @@
...
+        contents."""
+        # Arrange
+        shutil.copytree(FIXTURES_DIR / "make_addons/full/config", tmp_path, dirs_exist_ok=True)
+        os.chdir(tmp_path)
+
+        # Act
+        result = CliRunner().invoke(make_addons)
+
+        # Assert:
+        # The run should have been a success:
```
lgarvey, line 75: The comment seems superfluous as the assertion is fairly obvious; or you could you could add the text to the assertion, e.g 

```
assert condition, "some text if assertion is triggered"
```

tests/conftest.py
```diff
@@ -81,3 +83,9 @@ def mocked_pg_secret():
             Name="/copilot/dbt-app/staging/secrets/POSTGRES",
             SecretString='{"password":"abc123","dbname":"main","engine":"postgres","port":5432,"dbInstanceIdentifier":"dbt-app-staging-addons-postgresdbinstance-blah","host":"dbt-app-staging-addons-postgresdbinstance-blah.whatever.eu-west-2.rds.amazonaws.com","username":"postgres"}',
         )
+
+
+def switch_to_tmp_dir_and_copy_config_file(tmp_path, valid_config_file, destination_filename):
```
yusufsheiqh, line 88: Shall we default the `destination_filename` to `bootstrap.yml` so we don't need to pass it in each time?

```suggestion
def switch_to_tmp_dir_and_copy_config_file(tmp_path, valid_config_file, destination_filename = "bootstrap.yml"):
```

tests/test_copilot_cli.py
```diff
@@ -58,6 +61,37 @@
...
+        assert result.exit_code == 0
+        # We have an aurora DB, so we expect a warning:
+        assert "Note: The key DATABASE_CREDENTIALS may need to be changed" in result.stdout
+
+        # The files generated are as expected
+        expected_dir = FIXTURES_DIR / "make_addons" / "full" / "expected"
+        for path, _, files in os.walk(expected_dir):
```
lgarvey, line 82: minor one, but most (all?) of the `os.*` functionality is available in the newer `pathlib.Path` module, e.g. `Path().iterdir` could replace `os.walk`

lgarvey (approved):
Looks good!  👍 

I just added a couple of minor comments.

tests/test_copilot_cli.py
lgarvey: It'd make sense to run a test per addon and verify that those files exist. That way the test is more focused which will make them easier to fix in the future.

tests/test_copilot_cli.py
antroy-madetech: I'm inclined to leave this one, as iterdir is the equivalent of os.listdir, so doesn't recurse.

commands/copilot_cli.py
antroy-madetech: I meant to use this in the test file as it is used a lot. Done now (pending push).

tests/fixtures/make_addons/full/config/copilot/web/manifest.yml
antroy-madetech: These last two I'm inclined to leave as is. If the test fails due to a change in how/what addon config is generated, then leaving it as is means the expected data is easy to pull back in from the copliot check-cloudformation command.

commands/copilot_cli.py
WillGibson: Bit of an edge case, but if you use the same constant in the tests, is there is a chance that we could change the value of this constant, break things in the real world, but have all the tests still passing?

tests/test_bootstrap_cli.py
```diff
@@ -389,3 +384,9 @@ def setup_newenv_environment(tmp_path, runner):
     os.mkdir(f"{tmp_path}/copilot/environments/newenv")
     to_file = Path(tmp_path / "copilot/environments/newenv/manifest.yml")
     shutil.copy(my_file, to_file)
+
+
+def switch_to_tmp_dir_and_copy_config_file(tmp_path, valid_config_file):
+    os.chdir(tmp_path)
+    Path("bootstrap.yml").parent.mkdir(parents=True, exist_ok=True)
```
WillGibson, line 391: What is this line doing?

antroy-madetech: No longer needed. I was using it to create parent dirs when I'd repurposed it to copy arbitrary files. No longer needed though.

WillGibson (approved):

tests/fixtures/make_addons/full/config/copilot/web/manifest.yml
yusufsheiqh: The environments and secrets/env vars are taken from the `bootstrap.yml` file, so if we strip them from there, then the generation of these files will be a lot less than what we have now. It also makes our code base a bit more secure as we don't publicise potential env vars that are used within DBT applications.

tests/fixtures/make_addons/full/config/addons.yml
yusufsheiqh: Resolved in https://github.com/uktrade/copilot-tools/pull/138/commits/3852d901e00dafc78621c5184a53afe529b04eba.

yusufsheiqh (approved):
Looks great to me! 🚀

tests/fixtures/make_addons/full/config/copilot/web/manifest.yml
yusufsheiqh: Resolved in https://github.com/uktrade/copilot-tools/pull/138/commits/fa584a8a39aafc481d523c696b86e40c0dec4a8e.

tests/fixtures/make_addons/full/config/copilot/web/manifest.yml
yusufsheiqh: Resolved in https://github.com/uktrade/copilot-tools/pull/138/commits/fa584a8a39aafc481d523c696b86e40c0dec4a8e.

tests/test_copilot_cli.py
```diff
@@ -56,10 +59,67 @@
       bucket-name: my-bucket-dev
 """
 
+ADDON_CONFIG_FILENAME = "addons.yml"
+
 
 class TestMakeAddonCommand:
+    @pytest.mark.parametrize(
```
lgarvey, line 66: Spot on! 👍

PR #139, opened 2023-08-17
PyPI package publishing

antroy-madetech:
This PR creates a codebuild pipline that runs the tests using tox, and if successful and the version number has not already been published will build and push the package to PyPI.

The utility script is to facilitate getting the package version from the pyproject.toml and compare it to the published versions from the PyPI RSS feed.

### DIFF
--- a/buildspec.pypi.yml
+++ b/buildspec.pypi.yml
@@ -1,0 +1,29 @@
+version: 0.2
+
+env:
+  shell: bash
+  secrets-manager:
+    PYPI_TOKEN: "pypi:token"
+
+phases:
+  install:
+    commands:
+      - echo Setting local Python versions
+      - pyenv versions | awk 'match($0, /\d\.\d+\.\d+/) { print substr($0, RSTART, RLENGTH) }' > .python-version
+      - echo Installing dependencies
+      - pip install poetry
+      - poetry install
+
+  build:
+    commands:
+      - echo Running check-cloudformation
+      - poetry run ./copilot_helper.py check-cloudformation
+      - echo Running tests
+      - poetry run tox
+      - echo Building python package
+      - poetry build
+      - echo Publishing python package
+      - poetry config pypi-token.pypi ${PYPI_TOKEN}
+      - poetry publish
+      - echo Checking the package has reached PyPI
+      - python utils/check_pypi.py
--- a/buildspec.pypi.yml
+++ b/buildspec.pypi.yml
@@ -10,6 +10,8 @@ phases:
     commands:
       - echo Setting local Python versions
       - pyenv versions | awk 'match($0, /\d\.\d+\.\d+/) { print substr($0, RSTART, RLENGTH) }' > .python-version
+      - echo Ensure we are using python 3.11
+      - pyenv local 3.11
       - echo Installing dependencies
       - pip install poetry
       - poetry install
@@ -21,8 +23,6 @@ phases:
       - poetry run ./copilot_helper.py check-cloudformation
       - echo Running tests
       - poetry run tox
-      - echo Ensure we are using python 3.11
-      - pyenv local 3.11
       - |
         if ! python utils/check_pypi.py
         then
--- a/buildspec.pypi.yml
+++ b/buildspec.pypi.yml
@@ -2,8 +2,8 @@ version: 0.2
 
 env:
   shell: bash
-  secrets-manager:
-    PYPI_TOKEN: "pypi:token"
+  parameter-store:
+    PYPI_TOKEN: "pypi-token"
 
 phases:
   install:
--- a/buildspec.pypi.yml
+++ b/buildspec.pypi.yml
@@ -10,8 +10,6 @@ phases:
     commands:
       - echo Setting local Python versions
       - pyenv versions | awk 'match($0, /\d\.\d+\.\d+/) { print substr($0, RSTART, RLENGTH) }' > .python-version
-      - echo Ensure we are using python 3.11
-      - pyenv local 3.11
       - echo Installing dependencies
       - pip install poetry
       - poetry install
@@ -26,7 +24,11 @@ phases:
       - |
         if ! python utils/check_pypi.py
         then
+          echo Ensure we are using python 3.11
+          pyenv local 3.11
           echo Building python package
+          pip install poetry
+          poetry install
           poetry build
           VERSION=$(python utils/check_pypi.py --version)
           echo Publishing python package ${VERSION}
--- a/buildspec.pypi.yml
+++ b/buildspec.pypi.yml
@@ -9,29 +9,24 @@ phases:
   install:
     commands:
       - echo Setting local Python versions
-      - pyenv versions | awk 'match($0, /\d\.\d+\.\d+/) { print substr($0, RSTART, RLENGTH) }' > .python-version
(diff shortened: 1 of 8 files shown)


buildspec.pypi.yml
```diff
@@ -0,0 +1,36 @@
...
+      - poetry install
+
+  build:
+    commands:
+      - echo Running check-cloudformation
+      - poetry run ./copilot_helper.py check-cloudformation
+      - echo Running tests
+      - poetry run tox
+      - |
+        if ! python utils/check_pypi.py
```
acodeninja, line 24: It feels like a lot of this could be replaced with the following.

```shell
REMOTE_VERSION="$(curl -s "https://pypi.org/pypi/dbt-copilot-tools/json" | jq -rc '.info.version')"
LOCAL_VERSION="$(cat pyproject.toml | grep "^version" | awk '{ print $3; }' | jq -rc)"

if [[ $LOCAL_VERSION > $REMOTE_VERSION ]]; then
  echo "updating"
else
  echo "not updating"
fi
```

Though happy with a python script if we feel that is more maintainable and tests are added to it.

acodeninja (commented):
Thinking this could be a bit shorter, or tests added.

buildspec.pypi.yml
yusufsheiqh, line 1: Does this replace the `buildspec.test.yml` file?

utils/check_pypi.py
yusufsheiqh, line 1: We will probably want to add some tests around this to confirm this is working as expected.

buildspec.pypi.yml
antroy-madetech: No. That buildspec will still run on PRs. This will just run on merge to main.

buildspec.pypi.yml
antroy-madetech: I prefer the python as it is much more readable than awk/grep/jq chains. I'll add some tests.

antroy-madetech: Didn't realise there was a json endpoint in PyPI though, so I'll change to that from the rss feed, as searching the XML is fairly horrible.

buildspec.pypi.yml
yusufsheiqh: The `buildspec.test.yml` is run on the `main` branch as well, so are we looking to change that to be only triggered on PR branches? Otherwise, this will run tests twice.

yusufsheiqh: Resolved.

yusufsheiqh (commented):
CodeBuild project build is failing - will review once this is passing.

buildspec.pypi.yml
```diff
@@ -0,0 +1,41 @@
...
+        then
+          echo Ensure we are using python 3.11
+          pyenv local 3.11
+          echo Building python package
+          pip install poetry
+          poetry install
```
yusufsheiqh, line 31: I wonder if we could do a `pyenv global 3.11` in the `install` phase to avoid having to install `poetry` and all the dependencies again.

buildspec.pypi.yml
```diff
@@ -0,0 +1,38 @@
...
+    commands:
+      - echo Setting local Python versions
+      - pyenv versions | awk 'match($0, /\d\.\d+\.\d+/) { print substr($0, RSTART, RLENGTH) }' | sort -rV > .python-version
+      - echo Installing dependencies
+      - pip install poetry
+      - poetry install
+
+  build:
+    commands:
+      - python --version
```
yusufsheiqh, line 19: We can also remove this line.

```suggestion
```

buildspec.pypi.yml
```diff
@@ -0,0 +1,38 @@
...
+      - echo Running check-cloudformation
+      - poetry run ./copilot_helper.py check-cloudformation
+      - echo Running tests
+      - poetry run tox
+      - |
+        python --version
```
yusufsheiqh, line 25: This line is no longer needed.

```suggestion
```

buildspec.pypi.yml
```diff
@@ -0,0 +1,38 @@
...
+        VERSION=$(python utils/check_pypi.py --version)
+        if ! python utils/check_pypi.py
+        then
+          echo Building python package
+          poetry build
+          echo Publishing python package ${VERSION}
```
yusufsheiqh, line 31:
```suggestion
          echo Publishing Python package ${VERSION}
```

yusufsheiqh: Resolved in https://github.com/uktrade/copilot-tools/pull/139/commits/22ab4bf423dcd58c1b3e639e6fceaed250e85267.

buildspec.pypi.yml
yusufsheiqh: Resolved in https://github.com/uktrade/copilot-tools/pull/139/commits/22ab4bf423dcd58c1b3e639e6fceaed250e85267.

buildspec.pypi.yml
yusufsheiqh: Resolved in https://github.com/uktrade/copilot-tools/pull/139/commits/22ab4bf423dcd58c1b3e639e6fceaed250e85267.

buildspec.pypi.yml
```diff
@@ -10,6 +10,8 @@ phases:
     commands:
       - echo Setting local Python versions
       - pyenv versions | awk 'match($0, /\d\.\d+\.\d+/) { print substr($0, RSTART, RLENGTH) }' > .python-version
+      - echo Ensure we are using python 3.11
+      - pyenv local 3.11
```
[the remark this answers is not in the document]
yusufsheiqh: Resolved.

buildspec.pypi.yml
yusufsheiqh: Resolved in https://github.com/uktrade/copilot-tools/pull/139/commits/22ab4bf423dcd58c1b3e639e6fceaed250e85267.

yusufsheiqh (approved):

buildspec.pypi.yml
```diff
@@ -0,0 +1,36 @@
...
+    commands:
+      - echo Setting local Python versions
+      - pyenv versions | awk 'match($0, /\d\.\d+\.\d+/) { print substr($0, RSTART, RLENGTH) }' | sort -rV > .python-version
+      - echo Installing dependencies
+      - pip install poetry
+      - poetry install
+
+  build:
+    commands:
+      - echo Running check-cloudformation
```
lgarvey, line 19: Do we want this to only happen against merges to `main`?

yusufsheiqh: We have another CodeBuild build project that does this on PRs: https://github.com/uktrade/copilot-tools/blob/main/buildspec.test.yml

yusufsheiqh: `buildspec.test.yml` for PR branches and `buildspec.pypi.yml` on `main`.

lgarvey: Sorry, not the CFN linting, I mean do we want it to public to pypi only on merging into main?  Or is this how the codebuild project will bet set up?

yusufsheiqh: Yes, I think so. If we had multiple PRs open, then the version bumps may conflict with each other, so merging into `main` and then triggering this pipeline would be better, IMO, as the other PR branches would have to merge/rebase `main`.

PR #140, opened 2023-08-18
DBPT-377 & DBPT-330 - copilot-helper conduit with new images.

acodeninja:
## What?

* New conduit command format `copilot-helper conduit [addon_type] --app [application] --env [environment] <--addon-name [addon_name]>`
* New conduit images that auto-terminate 5 minutes after all users disconnect covering services:
    * opensearch (opensearch-cli)
    * postgres (psql)
    * redis (redis-cli)
* Fix template issue with access policy for opensearch domains, correcting the resource ARN

## Why?

Developers need to connect to backing services provisioned on their account, these services may have customised names like `sso-postgres`.

Developers should not gain shell access to a container as copilot starts sessions to containers under the root user which cannot be changed.

Conduit containers should not run for extended lengths of time when no developer is connected to them.

Developers must never paste credentials on the command line as they will appear in ECS exec session logs.

### Conduit command

* New single command function for all conduit addon types
* Full test suite for each function and command

### No-shell containers

For postgres and redis, the login shell profile launches the needed cli app with credentials taken from secrets manager.

For opensearch a repl is used to pass commands to the `opensearch-cli` program with the appropriate connection profile.

When the developer exits the client, the shell session automatically exits, never allowing the developer to land in a running shell.

### Automatic container shutdown

Each container has an entrypoint bash script that looks for running sessions using the client processes to detect when no developers are connected to the container. If no developer is connected for 5 minutes, the entrypoint script finishes, and the container exits with a 0 exit code.

## Todo before merging

- [ ] Add new build pipelines for each docker image in terraform tools

## Todo after merging

- [ ] Update all deployment repositories that use opensearch to use the new access policy


commands/conduit_cli.py
```diff
@@ -38,144 +55,117 @@ def get_cluster_arn(app: str, env: str) -> str:
...
-    if connected == False:
-        print(
-            f"Attempt to exec into running task timed out. Try again by running `copilot task exec --app {app} --env {env}` or check status of task in Amazon ECS console."
+@click.command()
+@click.argument("addon_type")
+@click.option("--app", help="AWS application name", required=True)
+@click.option("--env", help="AWS environment name", required=True)
+@click.option("--addon-name", help="Name of custom addon", required=False)
+def conduit(addon_type: str, app: str, env: str, addon_name: str):
+    """Create a conduit connection to a backing service of ADDON_TYPE."""
```
yusufsheiqh, line 153:
```suggestion
    """Create a conduit connection to a ADDON_TYPE backing service."""
```

images/conduit/opensearch/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+1. `aws sso login`
+2. `aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/uktrade`
+3. `docker build -t public.ecr.aws/uktrade/tunnel:opensearch .`
+4. `docker tag public.ecr.aws/uktrade/tunnel:opensearch public.ecr.aws/uktrade/tunnel:opensearch-$(git rev-parse --short HEAD)`
+5. `docker push public.ecr.aws/uktrade/tunnel:opensearch`
+6. `docker push public.ecr.aws/uktrade/tunnel:opensearch-$(git rev-parse --short HEAD)`
+7. `docker logout public.ecr.aws/uktrade`
+
+## Testing Locally
```
yusufsheiqh, line 20:
```suggestion
## Testing locally
```

images/conduit/opensearch/entrypoint.sh
```diff
@@ -0,0 +1,39 @@
...
+echo $CONNECTION_SECRET | gawk '{
+  match($0, /^https:\/\/([A-Za-z0-9_]+):([^@]+)@(.+)$/, arr);
+  print "profiles:"
+  print "    - name: connection"
+  print "      endpoint: \"https://"arr[3]"\""
+  print "      user: \""arr[1]"\""
+  print "      password: \""arr[2]"\""
+}' > /root/.opensearch-cli/config.yaml
+
+chmod 600 /root/.opensearch-cli/config.yaml
```
yusufsheiqh, line 15: Should we have a comment here to explain why we set this to `600`? It won't work if it's something else.

images/conduit/redis/README.md
```diff
@@ -0,0 +1,33 @@
+# Redis Conduit
+
+## Publishing Manually
```
yusufsheiqh, line 3:
```suggestion
## Publishing manually
```

images/conduit/redis/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+1. `aws sso login`
+2. `aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin public.ecr.aws/uktrade`
+3. `docker build -t public.ecr.aws/uktrade/tunnel:redis .`
+4. `docker tag public.ecr.aws/uktrade/tunnel:redis public.ecr.aws/uktrade/tunnel:redis-$(git rev-parse --short HEAD)`
+5. `docker push public.ecr.aws/uktrade/tunnel:redis`
+6. `docker push public.ecr.aws/uktrade/tunnel:redis-$(git rev-parse --short HEAD)`
+7. `docker logout public.ecr.aws/uktrade`
+
+## Testing Locally
```
yusufsheiqh, line 20:
```suggestion
## Testing locally
```

images/conduit/opensearch/entrypoint.sh
acodeninja: Added comment here

commands/conduit_cli.py
```diff
@@ -38,144 +55,117 @@ def get_cluster_arn(app: str, env: str) -> str:
...
-    subprocess.call(command, shell=True)
 
+def create_addon_client_task(app: str, env: str, addon_type: str, addon_name: str):
+    connection_secret_arn = get_connection_secret_arn(app, env, addon_name.upper())
 
-def is_task_running(cluster_arn: str, addon_type: str) -> bool:
-    tasks = boto3.client("ecs").list_tasks(
-        cluster=cluster_arn, desiredStatus="RUNNING", family=f"copilot-tunnel-{addon_type}"
+    subprocess.call(
+        f"copilot task run --app {app} --env {env} --task-group-name conduit-{addon_name} "
```
acodeninja, line 84: Lower case and hyphenate this for consistency, also in the other functions that use addon_name.

commands/conduit_cli.py
```diff
@@ -38,144 +55,117 @@ def get_cluster_arn(app: str, env: str) -> str:
...
-            return described_tasks["tasks"][0]["lastStatus"] == "RUNNING" and agent_running
+def addon_client_is_running(cluster_arn: str, addon_name: str) -> bool:
+    tasks = boto3.client("ecs").list_tasks(
+        cluster=cluster_arn,
+        desiredStatus="RUNNING",
+        family=f"copilot-conduit-{addon_name}",
+    )
 
-    except ValueError:
+    if not tasks["taskArns"]:
```
lgarvey, line 100: Minor one, you could probably just do

```
return not tasks["taskArns"]
```

commands/templates/addons/env/opensearch.yml
```diff
@@ -152,7 +152,7 @@ Resources:
             AWS: '*'
           Action:
           - 'es:ESHttp*'
-          Resource: !Sub 'arn:aws:es:${AWS::Region}:${AWS::AccountId}:domain/${App}-${Env}-{{ service.name }}/*'
+          Resource: !Sub 'arn:aws:es:${AWS::Region}:${AWS::AccountId}:domain/{{ service.name|replace("-", "") }}*'
```
lgarvey, line 155: Will this cause problems if you have two environments in an account, e.g dev and staging, and they each have "mydatabase"?

commands/conduit_cli.py
acodeninja: This is a guard clause to stop execution if task arns is empty, for the next line where we call describe tasks. We could refactor to a try except return false if that makes it an easier read.

commands/templates/addons/env/opensearch.yml
acodeninja: This is down to the way the opensearch cloudformation resource works. The opensearch domain if named [cannot be updated in a way that would replace the service](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-opensearchservice-domain.html#cfn-opensearchservice-domain-domainname) and the default arn convention is `arn:aws:es:[region]:[account]:domain/[logical id without hyphens max 16 chars]-[random string]`.

This access policy will make the claim (that will be ignored) that anyone can access the other opensearch domains on the account, but as the principal is any resource in the account this is the case anyway.

lgarvey (approved):
Great work! 🚀

PR #145, opened 2023-08-23
DBTP-330 redis and postgres conduits

acodeninja:
# What?

* Switch all conduits to base off debian slim
* Reduce size of postgres image from 163.69MB to 62.13MB
* Install ca certificates for redis to get latest AWS ca for redis clusters

# Why?

* redis conduit was unable to connect to aws redis cluster "ssl_verify" error
* use debian slim for everything so all environments are consistent and small in size.

### DIFF
--- a/images/conduit/postgres/Dockerfile
+++ b/images/conduit/postgres/Dockerfile
@@ -1,7 +1,9 @@
-# Pull latest PostgreSQL Docker image
-FROM public.ecr.aws/docker/library/postgres:latest
+FROM public.ecr.aws/docker/library/debian:12-slim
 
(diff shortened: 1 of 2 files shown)


acodeninja:
> Can you add the notes in the PR description into comments in the Dockerfiles.

Would you be alright with this going into the readmes? That way we can document the reasoning. The other reason not to use the postgres image is that we don't want a postgres server starting up on the task...

yusufsheiqh (approved):

antroy-madetech:
> 

May be worth a comment in the Dockerfiles pointing at the README. Not everyone will read the README before changing the Dockerfile.

antroy-madetech (approved):

PR #147, opened 2023-08-25
Lock image builder version

ejayesh:
add copilot version to an env VAR that can be used in codebuild


WillGibson (approved):
Looks legit to me. Proof will be in the eating ;-)

PR #151, opened 2023-08-31
Fix opensearch secret name and description

gabelton:
Minor issue encountered when completing https://uktrade.atlassian.net/browse/DBTP-403


yusufsheiqh (commented):
Do we even need this secret in Secrets Manager? I believe we use OpenSearch with Parameter Store.

PR #154, opened 2023-09-01
DBTP-433 - Revert VPC configuration to use default configuration

yusufsheiqh:
## Context

- Revert to using the default VPC configuration
  - The default is 2 public and 2 private subnets.
- Update version from `0.1.32` -> `0.1.33`

## Additional information

Jira ticket - https://uktrade.atlassian.net/browse/DBTP-433

### DIFF
--- a/commands/templates/env/manifest.yml
+++ b/commands/templates/env/manifest.yml
@@ -12,25 +12,6 @@ type: Environment
 #   vpc:
 #     id:
 
-network:
-  vpc:
-    cidr: '10.0.0.0/16'
-    subnets:
-      public:
-        - cidr: '10.0.0.0/24'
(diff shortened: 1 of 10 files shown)


lgarvey (approved):

commands/templates/env/manifest.yml
```diff
@@ -12,25 +12,6 @@ type: Environment
 #   vpc:
 #     id:
```
WillGibson, line 13: While we are in the neighbourhood, shall we just delete the four lines above as they are more likely to cause confusion than help? If yes, applies elsewhere in too.

WillGibson (approved):
One suggestion that you are welcome to ignore, so approving in case you do.

PR #157, opened 2023-09-06
Add temporary information about readonly file system causing woes

WillGibson:


commands/templates/svc/manifest-public.yml
```diff
@@ -66,6 +66,15 @@ network:
   vpc:
     placement: 'private'
 
+# The application currently does not work of the file system is readonly, e.g...
```
yusufsheiqh, line 69:
```suggestion
# The application currently does not work if the file system is readonly, e.g...
```

commands/templates/svc/manifest-public.yml
```diff
@@ -66,6 +66,15 @@ network:
...
+#
+# File "/layers/paketo-buildpacks_cpython/cpython/lib/python3.9/tempfile.py", line 291, in gettempdir
+# tempdir = _get_default_tempdir()
+# File "/layers/paketo-buildpacks_cpython/cpython/lib/python3.9/tempfile.py", line 223, in _get_default_tempdir
+# raise FileNotFoundError(_errno.ENOENT
+# FileNotFoundError: [Errno 2] No usable temporary directory found in ['/tmp', '/var/tmp', '/usr/tmp', '/workspace']
```
yusufsheiqh, line 75: I am actually not sure we need all this. A sentence or two stating why the following two lines need commenting out would be sufficient, IMO.

WillGibson: How about now?

yusufsheiqh: Yup, that's nice and smol.

yusufsheiqh: Resolved in https://github.com/uktrade/copilot-tools/pull/157/commits/4f781d00eeffc32e0ef67bb0b4b6da00437f1d4e.

yusufsheiqh (commented):
Tests are failing ATM, and the package version will also need to be bumped up.

yusufsheiqh (approved):

PR #158, opened 2023-09-07
fix: passwords in urls with invalid characters

acodeninja:
# What?

Do not allow using characters in addons credentials that cause an invalid connection url to be generated.

# Why?

Python URL library fails to parse a URL where the password contains `/`.

# How?

Use the `ExcludeCharacters` property on the `AWS::SecretsManager::Secret` resource to exclude characters that are not valid in a URL.

### DIFF
--- a/commands/templates/addons/env/aurora-postgres.yml
+++ b/commands/templates/addons/env/aurora-postgres.yml
@@ -76,6 +76,7 @@ Resources:
         ExcludePunctuation: true
         IncludeSpace: false
         PasswordLength: 16
+        ExcludeCharacters: '"@/\;='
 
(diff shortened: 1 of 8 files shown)


yusufsheiqh (commented):
Do we not need to do the same for Redis?

yusufsheiqh (approved):

acodeninja:
Redis does not use `AWS::SecretsManager::Secret` so no need to update.

yusufsheiqh:
> Redis does not use `AWS::SecretsManager::Secret` so no need to update.

@acodeninja Yeahhh, that makes sense!

PR #161, opened 2023-09-11
Update check-domain command to handle prod env

gabelton:
Originating from [this ticket](https://uktrade.atlassian.net/browse/DBTP-398).

Currently `check-domain` will iterate over each environment in the web `manifest.yml`, but when we're deploying a prod service we're only interested in... prod. 

This PR adds an optional `--env` flag to the command, so that users can target individual environments instead of the whole list.


lgarvey (approved):
Looks good! 🚀

PR #162, opened 2023-09-13
Added further characters to exclude from passwords

antroy-madetech:
The ExcludeCharacters field needed updating, as python's urlparse function was failing to parse opensearch URLs that contain the generated password. In this case, it was square brackets that were causing the issue, but i have excluded parentheses and braces as well to ensure we don't hit this issue again.


yusufsheiqh (approved):
Interesting... these passwords will only be alphanumeric.

PR #163, opened 2023-09-14
Added EMAIL variable to the ipfilter sidecar

antroy-madetech:
EMAIL variable was required for the ipfilter sidecar. This PR adds it.


yusufsheiqh (commented):
Looks good, but the version needs to be bumped so this will be picked up in the latest version of `dbt-copilot-tools`.

yusufsheiqh (approved):

PR #167, opened 2023-09-19
Refactor structure

acodeninja:
- change name of module folder as `commands` may conflict with other site packages
- move commands into separate module and change to consistent names


yusufsheiqh (commented):
I see we have used `dbt_copilot_helper`, but I think it would make more sense to use `copilot_tools` or `dbt_copilot_tools` instead as that is the package name. The script entrypoint is kept as `copilot-helper` for convenience as we use `copilot` CLI as well.

tests/test_command_bootstrap.py
```diff
@@ -107,6 +107,8 @@ def test_make_config(tmp_path):
     switch_to_tmp_dir_and_copy_config_file(tmp_path, "test_config.yml")
     os.mkdir(f"{tmp_path}/copilot")
 
+    print(os.listdir(tmp_path))
```
WillGibson, line 110: Did you mean to leave this in?

acodeninja:
> I see we have used `dbt_copilot_helper`, but I think it would make more sense to use `copilot_tools` or `dbt_copilot_tools` instead as that is the package name. The script entrypoint is kept as `copilot-helper` for convenience as we use `copilot` CLI as well.

I did think about this, can move to `dbt_copilot_tools` but my thought was that the command is `copilot_helper` so it made sense to keep that name going, as conceptually `copilot_helper` and associated commands is part of the dbt-copilot-tools package.

yusufsheiqh:
> > I see we have used `dbt_copilot_helper`, but I think it would make more sense to use `copilot_tools` or `dbt_copilot_tools` instead as that is the package name. The script entrypoint is kept as `copilot-helper` for convenience as we use `copilot` CLI as well.
> 
> I did think about this, can move to `dbt_copilot_tools` but my thought was that the command is `copilot_helper` so it made sense to keep that name going, as conceptually `copilot_helper` and associated commands is part of the dbt-copilot-tools package.

@acodeninja I think I prefer `dbt_copilot_tools` in that case, although the `dbt` prefix may be redundant. I really like the dropping of the `_cli` suffix for the commands.

Another point to bring up... we are using `utils.py` and the `utils` folder, but I think it would be worthwhile if we could consolidate both of these as part of this restructure as they are both used for `dbt_copilot_tools`.

WillGibson:
Is now the time to clarify the purpose of this repository and possibly rename it?

Is it expecting to include more tools than `copilot-helper`?

If it is, would we publish them all under the "tools` package or perhaps individually?

Currently as it is only `copilot-helper` and no sign of it being something else too, I keep having a desire to rename the repository and package to `dbt-copilot-helper`.

WillGibson (approved):

yusufsheiqh:
The repository name makes sense to me, since we do have migration scripts and files such as the `bootstrap-config` files.

But, it might be time to rename the package to `dbt-copilot-helper` as we only package up the things required for our `copilot-helper` script entrypoint such as templates, commands, etc.

yusufsheiqh (approved):
Let's get this merged as this PR makes sense and move forwards based on the discussions we are having. 🚀

PR #169, opened 2023-09-19
Added Name tags to the redis and rds addon security groups

antroy-madetech:
Redis and RDS Postgres did not have Name tags on their security groups.


acodeninja (approved):
You'll want to bump the version, otherwise great!

WillGibson (approved):

yusufsheiqh (approved):
What @acodeninja said.

PR #172, opened 2023-09-21
update make-config to add autoscaler info per env

gabelton:
Part of [this ticket](https://uktrade.atlassian.net/browse/DBTP-398)

Prod environments are likely to have different autoscaling requirements. This PR, therefore, updates the `make-config` command to allow adding separate scaling info via the `bootstrap.yml` file.

- Updates environment schema based on copilot autoscaler docs
- adds a to_yaml filter to web manifest jinja template
- bumps package
- updates make-config test with new prod example


dbt_copilot_helper/commands/bootstrap.py
```diff
@@ -101,9 +125,13 @@ def load_and_validate_config(path):
 
     # validate the file
     schema = Schema(config_schema)
-    config = schema.validate(conf)
```
gabelton, line 104: this line was reordering the config (all nested dicts were moved to the bottom), which I don't think was desired behaviour.

tests/fixtures/test_service_manifest.yml
```diff
@@ -90,6 +90,19 @@ environments:
   production:
     http:
       alias: test-service.trad.gav.ik
+    count:
```
acodeninja, line 93: Would be nice to add a comment here:

```yaml
count: # For options see https://aws.github.io/copilot-cli/docs/manifest/lb-web-service/#count
...
```

acodeninja (commented):
Looks good, though can we update the `migration-tools/build-copilot-config.py` app to include the count in the bootstrap configs it generates even if it just defaults to `1`.

I wonder if the cloudfoundry client supports getting scaling info from app instances?

dbt_copilot_helper/commands/bootstrap.py
```diff
@@ -41,6 +42,29 @@
                         "paas": str,
                         Optional("url"): str,
                         Optional("ipfilter"): bool,
+                        Optional("count"): Or(
+                            int,
+                            {  # https://aws.github.io/copilot-cli/docs/manifest/lb-web-service/#count
+                                "range": str,  # e.g. 1-10
```
yusufsheiqh, line 48: Can we change this to be more specific? I can put `range: asdf` and it would work.

yusufsheiqh (commented):
I think if we can be more specific with the validation, then it would be more helpful, but as it stands, the values could be any string and it would be fine in the output.

gabelton:
> Looks good, though can we update the `migration-tools/build-copilot-config.py` app to include the count in the bootstrap configs it generates even if it just defaults to `1`.
> 
> I wonder if the cloudfoundry client supports getting scaling info from app instances?

I think we already have a default now:
```
count: # See https://aws.github.io/copilot-cli/docs/manifest/lb-web-service/#count
  range: 2-10
  cooldown:
    in: 120s
    out: 60s
  cpu_percentage: 50
```
(At least that's what the `make-config` command produced for me when I ran it most recently. I think that's preferable to setting the same thing for each environment)

I could create a separate ticket to investigate finding values from cloudfoundry and adding to the bootstrap file?

yusufsheiqh:
> > Looks good, though can we update the `migration-tools/build-copilot-config.py` app to include the count in the bootstrap configs it generates even if it just defaults to `1`.
> > I wonder if the cloudfoundry client supports getting scaling info from app instances?
> 
> I think we already have a default now:
> 
> ```
> count: # See https://aws.github.io/copilot-cli/docs/manifest/lb-web-service/#count
>   range: 2-10
>   cooldown:
>     in: 120s
>     out: 60s
>   cpu_percentage: 50
> ```
> 
> (At least that's what the `make-config` command produced for me when I ran it most recently. I think that's preferable to setting the same thing for each environment)
> 
> I could create a separate ticket to investigate finding values from cloudfoundry and adding to the bootstrap file?

The `make-config` file relies on the `bootstrap.yml` file if I am not mistaken, so we most definitely need to at least pull in these values for the production environment from the Cloud Foundry client, if possible.

gabelton:
> > > Looks good, though can we update the `migration-tools/build-copilot-config.py` app to include the count in the bootstrap configs it generates even if it just defaults to `1`.
> > > I wonder if the cloudfoundry client supports getting scaling info from app instances?
> > 
> > 
> > I think we already have a default now:
> > ```
> > count: # See https://aws.github.io/copilot-cli/docs/manifest/lb-web-service/#count
> >   range: 2-10
> >   cooldown:
> >     in: 120s
> >     out: 60s
> >   cpu_percentage: 50
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > (At least that's what the `make-config` command produced for me when I ran it most recently. I think that's preferable to setting the same thing for each environment)
> > I could create a separate ticket to investigate finding values from cloudfoundry and adding to the bootstrap file?
> 
> The `make-config` file relies on the `bootstrap.yml` file if I am not mistaken, so we most definitely need to at least pull in these values for the production environment from the Cloud Foundry client, if possible.

I was thinking that people will have to at least cast their eye over `bootstrap.yml` before running that command and so could add any custom autoscaler config at that point, using the default already generated by `make-config`, as a fallback. Agree that ideally we'd pull in from cloudfoundry, but don't think it's a blocker.

gabelton:
> Looks good, though can we update the `migration-tools/build-copilot-config.py` app to include the count in the bootstrap configs it generates even if it just defaults to `1`.
> 
> I wonder if the cloudfoundry client supports getting scaling info from app instances?

I can get the below by running `cf app APP_NAME`:

Screenshot 2023-09-21 at 16 31 35

So I think it should be possible to get similar using the python client too.

dbt_copilot_helper/utils.py
```diff
@@ -286,3 +287,12 @@ class ClickDocOptGroup(click.Group):
 
     def format_usage(self, ctx: Context, formatter: HelpFormatter) -> None:
         format_click_usage(ctx, formatter, True)
+
+
+def validate_string(regex_pattern):
+    def validator(string):
+        if not re.match(regex_pattern, string):
+            raise SchemaError(f"String '{string}' does not match the required pattern.")
```
acodeninja, line 295: Maybe include the pattern in the error response possibly also with a link.
Alternatively you could catch, warn and default to `count: 1` if this is thrown.

acodeninja (commented):
Small thing to consider, otherwise nice work.

yusufsheiqh (approved):

PR #179, opened 2023-09-26
Fix pipeline

antroy-madetech:
Changes in the DBTP-456 ticket changed how the linting is done to make it applicable to arbitrary projects. A side effect of this is that I missed changing the command in one of the pipelines.

This change does 2 things:
- A small refactoring of the fixtures in the check-cloudformation tests that I was working on.
- Removing the lint check entirely from the pipeline, since wherever it was used it also ran the unit tests which do the same lint check on the same directory structure (tests/test-application)

### DIFF
--- a/buildspec.pypi.yml
+++ b/buildspec.pypi.yml
@@ -16,8 +16,6 @@ phases:
 
   build:
     commands:
-      - echo Running check-cloudformation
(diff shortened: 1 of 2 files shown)


acodeninja (approved):
Seems reasonable.

PR #185, opened 2023-10-03
DBTP-309 - Create a more restricted PostgreSQL user in CloudFormation

yusufsheiqh:
## Context

- Add secondary PostgreSQL user for RDS and Aurora databases
  - Added a Lambda function for both addons.
  - Added a custom resource for the application user.
  - Added ingress and egress rules for the security group that the Lambda Function sits in.
  - Added two NAT Gateways for internet access from both private subnets.
- Add VPC endpoint for Secrets Manager (`environments/addons/vpc-endpoint.yml`)
  - This allows the Lambda Functions and the Conduit addon to make API calls to Secrets Manager.
  - This is now created as a default environment addon, so we now have improved security.
- Add secrets to Parameter Store
  - This means development teams will not need access to Secrets Manager.

## Additional information

Updating the secret name requires the secret to be deleted and recreated. This change can be made in a subsequent PR as this will require a rebuild of the environment. This can be used like so; `copilot-helper conduit postgres --app application --env environment --addon-name application_APPLICATION_USER`, replacing `application` and `environment`.

There is another ticket in the backlog to investigate if we can potentially reuse the same NAT Gateways to reduce significant costs rather than Copilot spinning up an additional two - https://uktrade.atlassian.net/browse/DBTP-476


dbt_copilot_helper/templates/addons/env/aurora-postgres.yml
```diff
@@ -151,3 +152,309 @@ Resources:
       FromPort: 5432
       ToPort: 5432
       SourceSecurityGroupId: !Ref EnvironmentSecurityGroup
+
+  # Enable ingress from Lambda Functions.
```
WillGibson, line 156: I don't think we need these comments, the descriptions cover it already.

dbt_copilot_helper/templates/addons/env/aurora-postgres.yml
```diff
@@ -151,3 +152,309 @@ Resources:
...
+      GroupId: !Ref '{{ service.prefix }}DBClusterSecurityGroup'
+      IpProtocol: tcp
+      FromPort: 5432
+      ToPort: 5432
+      SourceSecurityGroupId: !Ref '{{ service.prefix }}DBClusterSecurityGroup'
+
+  # Enable ingress from Secrets Manager.
+  {{ service.prefix }}SecretsManagerIngress:
+    Metadata:
+      'aws:copilot:description': 'Allow ingress from Secrets Manager in my application to the DB cluster'
```
WillGibson, line 172: Why does Secrets Manager need ingress to the database cluster? Maybe we can add a little more detail into some of the decsriptions?

dbt_copilot_helper/templates/addons/env/aurora-postgres.yml
```diff
@@ -151,3 +152,309 @@ Resources:
...
+    Metadata:
+      'aws:copilot:description': 'Allow egress for HTTP'
+    Type: AWS::EC2::SecurityGroupEgress
+    Properties:
+      CidrIp: 0.0.0.0/0
+      Description: Egress for HTTP
+      GroupId: !Ref '{{ service.prefix }}DBClusterSecurityGroup'
+      IpProtocol: tcp
+      FromPort: 80
+      ToPort: 80
```
WillGibson, line 232: DO we need HTTPS and HTTP, or is HTTPS enough?

dbt_copilot_helper/templates/addons/env/aurora-postgres.yml
```diff
@@ -151,3 +152,309 @@ Resources:
...
+              )["RandomPassword"]
+
+              application_user_secret_string = {
+                  "username": "application_user",
+                  "password": application_user_password,
+                  "engine": master_user["engine"],
+                  "port": master_user["port"],
+                  "dbname": master_user["dbname"],
+                  "host": master_user["host"],
+                  "dbClusterIdentifier": master_user["dbClusterIdentifier"]
```
WillGibson, line 338: This is the only line which differs in the code of the two lambdas.

For a separate pull request, I wonder if we can publish method in `dbt-copilot-python` that takes a parameter to specify the string used above and reduce the duplication?

yusufsheiqh: We would need to publish the code to an S3 bucket, but we can also retrieve this info through the Custom Resource and use a `try/except` block. Not sure how useful `dbt-copilot-python` would be, unless we install the package in the `-deploy` repositories as well...

dbt_copilot_helper/templates/addons/env/aurora-postgres.yml
yusufsheiqh: It doesn't, and the description is wrong. One thing to note though is that I have placed the Lambda Function in the same security group as the database cluster.

dbt_copilot_helper/templates/addons/env/aurora-postgres.yml
yusufsheiqh: Not sure - I will test this out!

yusufsheiqh: HTTPS is enough! Resolved in https://github.com/uktrade/copilot-tools/pull/185/commits/a385eceb3ab4c605beb82f257a738a8b87386a51.

dbt_copilot_helper/templates/addons/env/aurora-postgres.yml
yusufsheiqh: Resolved in https://github.com/uktrade/copilot-tools/pull/185/commits/a385eceb3ab4c605beb82f257a738a8b87386a51.

dbt_copilot_helper/templates/addons/env/aurora-postgres.yml
yusufsheiqh: Resolved in https://github.com/uktrade/copilot-tools/pull/185/commits/a385eceb3ab4c605beb82f257a738a8b87386a51.

pyproject.toml
```diff
@@ -3,7 +3,7 @@ line-length = 100
 
 [tool.poetry]
 name = "dbt-copilot-tools"
-version = "0.1.56"
+version = "0.1.58"
```
WillGibson, line 6: I have taken a punt that my pull request with `0.1.58` will go in before this one.

dbt_copilot_helper/templates/addons/env/aurora-postgres.yml
```diff
@@ -151,3 +152,291 @@ Resources:
...
+        - arn:aws:lambda:eu-west-2:763451185160:layer:python-postgres:1
+      MemorySize: 128
+      Timeout: 10
+      Role: !GetAtt {{ service.prefix }}LambdaFunctionExecutionRole.Arn
+      VpcConfig:
+        SecurityGroupIds:
+          - !Ref {{ service.prefix }}DBClusterSecurityGroup
+        SubnetIds: !Split [ ",", !Ref PrivateSubnets ]
+      Code:
+        ZipFile: |
```
acodeninja, line 232: Any chance this can be extracted into a python file so it can be tested? Myself and @antroy-madetech are doing this for [creating a healthcheck file for the s3 proxy](https://github.com/uktrade/copilot-tools/compare/main...DBTP-457-s3-object-for-healthchecks) and we will read in the text content of the function when generating the template.

acodeninja (commented):
Little comment about being able to test this functionality.

dbt_copilot_helper/templates/addons/env/aurora-postgres.yml
WillGibson: In short, yes, but not in this pull request.

See the first bullet on [this comment](https://uktrade.atlassian.net/browse/DBTP-309?focusedCommentId=125191) from @yusufsheiqh.

Also, when you have yours working, that will probably inform how we get this out to a python file.

WillGibson (approved):

WillGibson (approved):
Reapproving after today's changes.

acodeninja (approved):

PR #186, opened 2023-10-06
Reduce verbosity of output for failed tests

WillGibson:
When a test fails, by default pytest outputs loads of the file contents around where it failed.

This makes it hard to see the wood from the trees, especially in the author's IDE which just colours all of it red.

Before...

image

After...

image


pytest.ini
```diff
@@ -1,4 +1,4 @@
 [pytest]
 env =
     AWS_DEFAULT_REGION=eu-west-2
-    
+addopts = --tb=short
```
antroy-madetech, line 4: I'm thinking that this would be better off in your IDE's configuration (or in your global pytest.ini) rather than enforcing for all users, as it could be useful in some cases.

WillGibson: As discussed outside of here, less noise as the default in the repository ones is best and we can just delete the line if/when we think the extra output might help.

It links you to the file anyway so we can just click through if we want file contents.

yusufsheiqh (approved):

PR #190, opened 2023-10-10
Remove repository flag

antroy-madetech:
We have removed the --repository flag from the copilot-helper svc deploy command as we can infer this from the service manifest.


acodeninja (approved):
Nice work.

yusufsheiqh (approved):

PR #192, opened 2023-10-11
DBTP-502 Add command to generate environment infrastructure pipeline config

antroy-madetech:
This is a first pass, taking the work @acodeninja spiked in `digital-workspace-deploy` and getting it coming out from `copilot-helper pipeline generate-config`.

This is not the end of the story, but a good first pass.


dbt_copilot_helper/commands/pipeline.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def safe_load_config(filename, schema):
+    try:
+        app_config = load_and_validate_config(filename, schema)
+    except FileNotFoundError:
+        abort_with_error(f"There is no {filename}")
+    except ParserError:
+        abort_with_error(f"The {filename} file is invalid")
+    return app_config
```
WillGibson, line 88:
```suggestion
def safe_load_config(filename, schema):
    try:
        return load_and_validate_config(filename, schema)
    except FileNotFoundError:
        abort_with_error(f"There is no {filename}")
    except ParserError:
        abort_with_error(f"The {filename} file is invalid")
```

dbt_copilot_helper/commands/pipeline.py
```diff
@@ -0,0 +1,101 @@
...
+
+    contents = templates.get_template("pipeline/overrides/cfn.patches.yml").render(template_data)
+    click.echo(
+        mkfile(
+            base_path,
+            pipelines_environments_dir / "overrides/cfn.patches.yml",
+            contents,
+            overwrite=True,
+        )
+    )
```
antroy-madetech, line 78: We can probably refactor these last few lines to tidy things up a bit

dbt_copilot_helper/utils/aws.py
```diff
@@ -163,3 +163,12 @@ def check_response(response):
             fg="red",
         )
         exit()
+
+
+def get_codestar_connection_arn(app_name):
+    client = boto3.client("codestar-connections")
+    response = client.list_connections()
```
WillGibson, line 170: Could inline this...
```suggestion
    response = boto3.client("codestar-connections").list_connections()
```

dbt_copilot_helper/commands/pipeline.py
```diff
@@ -0,0 +1,101 @@
...
+    git_repo = subprocess.run(
+        ["git", "remote", "get-url", "origin"], capture_output=True, text=True
+    ).stdout.strip()
+
+    if not git_repo:
+        return
+
+    domain, repo = git_repo.split("@")[1].split(":")
+
+    return f"https://{domain}/{re.sub(r'.git$', '' , repo)}"
```
antroy-madetech, line 101: Perhaps mark these as private to deter use outside of the module.

tests/fixtures/pipeline/pipelines.yml
```diff
@@ -0,0 +1,4 @@
+environments:
+  - name: dev
+  - name: prod
+    requires_approval: true
```
WillGibson, line 4: To stop GitHub moaning...
```suggestion
    requires_approval: true
    
```

tests/test_command_copilot.py
```diff
@@ -315,15 +314,28 @@ def test_env_addons_parameters_file_with_different_addon_types(
...
+    @patch("dbt_copilot_helper.jinja2_tags.version", new=Mock(return_value="v0.1-TEST"))
+    def test_env_addons_parameters_file_should_not_be_included_for_s3(
+        self, fakefs, validate_version
+    ):
+        fakefs.create_file(
+            ADDON_CONFIG_FILENAME,
+            contents=(S3_STORAGE_CONTENTS),
+        )
+        fakefs.create_file("copilot/web/manifest.yml")
+        fakefs.create_file("copilot/environments/development/manifest.yml")
```
WillGibson, line 330: I think there may be scope for extracting these create_file setup steps to a helper method. It could improve readability of the tests.

dbt_copilot_helper/templates/pipeline/manifest.yml
```diff
@@ -0,0 +1,39 @@
...
+  properties:
+    # Todo: Allow for overriding this, but without risking deploying a branch to higher environments
+    branch: main
+    repository: {{ git_repo }}
+    # Optional: specify the name of an existing CodeStar Connections connection.
+    connection_name: {{ app_name }}
+
+# This section defines the order of the environments your pipeline will deploy to.
+stages:
+  {% for environment in pipeline_environments -%}
```
acodeninja, line 26: For nicer yaml formatting, can we use an indent of 0 on for loops?

dbt_copilot_helper/utils/files.py
```diff
@@ -1,28 +1,133 @@
...
 
-def mkdir(base, path):
-    if (base / path).exists():
-        return f"Directory {path} exists; doing nothing"
+range_validator = validate_string(r"^\d+-\d+$")
+seconds_validator = validate_string(r"^\d+s$")
 
-    (base / path).mkdir(parents=True)
-    return f"Directory {path} created"
+BOOTSTRAP_SCHEMA = Schema(
```
acodeninja, line 14: As the schemas are related to validation rather than directly with files, might be better in the `dbt_copilot_helper.utils.validation` module?

dbt_copilot_helper/commands/pipeline.py
```diff
@@ -0,0 +1,97 @@
...
+    _create_file_from_template(
+        base_path, "manifest.yml", pipelines_environments_dir, template_data, templates
+    )
+    _create_file_from_template(
+        base_path, "overrides/cfn.patches.yml", pipelines_environments_dir, template_data, templates
+    )
+
+
+def _create_file_from_template(
+    base_path, buildspec, pipelines_environments_dir, template_data, templates
```
gabelton, line 72: Should this be renamed to something more generic?

```suggestion
    base_path, file_name, pipelines_environments_dir, template_data, templates
```

acodeninja (commented):
First pass, will come back later today.

dbt_copilot_helper/templates/pipeline/manifest.yml
```diff
@@ -0,0 +1,39 @@
...
+  # The name of the provider that is used to store the source artifacts.
+  # (i.e. GitHub, Bitbucket, CodeCommit)
+  provider: GitHub
+  # Additional properties that further specify the location of the artifacts.
+  properties:
+    # Todo: Allow for overriding this, but without risking deploying a branch to higher environments
+    branch: main
+    repository: {{ git_repo }}
+    # Optional: specify the name of an existing CodeStar Connections connection.
+    connection_name: {{ app_name }}
```
gabelton, line 22: Should this be `connection_name` ?

tests/test_command_pipeline.py
```diff
@@ -0,0 +1,161 @@
...
+    with open("bootstrap.yml", "w") as fh:
+        print("{invalid data", file=fh)
+
+    result = CliRunner().invoke(generate_config)
+
+    assert result.exit_code == 1
+    assert "Error: The bootstrap.yml file is invalid" in result.output
+
+
+def assert_file_created_in_stdout(output_file, result, tmp_path):
```
gabelton, line 126: Think this and below should live in a test utils file somewhere

tests/test_command_copilot.py
```diff
@@ -418,3 +415,12 @@ def _put_ssm_param(client, app, env, name, value):
     assert SSM_PATH.format(app="myapp", env="anotherenv", name="OTHER_ENV") not in result.output
     validate_version.assert_called_once()
     assert result.exit_code == 0
+
+
+def create_test_manifests(addon_file_contents, fakefs):
```
gabelton, line 420: This could live in a utils file

dbt_copilot_helper/templates/pipeline/buildspec.yml
```diff
@@ -0,0 +1,65 @@
+# {% version_info %}
+# Buildspec runs in the build stage of your environment pipeline to generate the environment CloudFormation stack config.
+version: 0.2
+env:
+  variables:
+    COPILOT_VERSION: 1.31.0
```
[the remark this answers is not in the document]
WillGibson: Interesting idea, we could get the latest 1.x version of AWS Copilot and dyff dynamically so we are not using too stale a version of each.

Might be a risk that we introduce instability without it being obvious what changed to make things go wrong.

It would be interesting to see what the others think 👂

dbt_copilot_helper/templates/pipeline/manifest.yml
WillGibson: Done

tests/test_command_pipeline.py
WillGibson: I have moved them to `conftest.py`, though that probably wants breaking up, but not in this pull request which has enough noise already.

tests/test_command_copilot.py
WillGibson: I think this method is actually fairly specific to the running of the `make-addons` tests. I'm not sure moving it would be right.

dbt_copilot_helper/templates/pipeline/buildspec.yml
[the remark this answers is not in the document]
antroy-madetech: I think that it would be fine to pass the version in when the tool is run. If we are running this for the first time for a project, we'd probably want the latest version anyway. If we're updating existing pipelines, then the version change would show in the diff.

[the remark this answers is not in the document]
antroy-madetech: Raised a ticket for this: https://uktrade.atlassian.net/browse/DBTP-513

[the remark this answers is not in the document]
acodeninja: Added a comment on the above ticket about also including this in the check config command and helper.

acodeninja (commented):
One small alteration suggested, otherwise looks good to go.

gabelton (approved):
Think there's still one comment of Lawrence's to be resolved, but otherwise looks good!

acodeninja (approved):
Awesome

PR #194, opened 2023-10-17
DBTP-503 Add Cloud Foundry CLI to conduit image to enable Postgres data migration

WillGibson:
First part of https://uktrade.atlassian.net/browse/DBTP-503


images/conduit/postgres/Dockerfile
```diff
@@ -2,8 +2,18 @@ FROM public.ecr.aws/docker/library/debian:12-slim
...
+# Install Clound Foundry CLI for data migrations
+RUN apt-get install -y apt-utils curl gnupg && \
+    curl -sS https://packages.cloudfoundry.org/debian/cli.cloudfoundry.org.key | gpg --dearmor > /etc/apt/trusted.gpg.d/cloudfoundry-cli.gpg && \
+    echo "deb https://packages.cloudfoundry.org/debian stable main" | tee /etc/apt/sources.list.d/cloudfoundry-cli.list && \
+    apt-get update && \
+    apt-get install -y cf8-cli && \
+    cf install-plugin -f https://github.com/alphagov/paas-cf-conduit/releases/download/v0.1.0/cf-conduit.linux.arm64 && \
+    apt-get remove -y apt-utils curl gnupg
+
+RUN apt-get clean
```
acodeninja, line 16: Clean should be under each of the runs, or combine them, the cached layers for the previous `RUN` statements will store the apt cache, so this will not save any space. Have a look at the built image with [dive](https://github.com/wagoodman/dive) to take a look at the space used by each layer to confirm this.

WillGibson: This was still a draft pull request, but I think we need to tidy up and merge now in order to use it in the wild.

Should be good now.

acodeninja (approved):
Seems reasonable to me 🚀

PR #195, opened 2023-10-17
DBTP-430 Update Paketo builder and run images

gabelton:
https://uktrade.atlassian.net/browse/DBTP-430

The previous builder image was archived and not receiving updates, meaning ecr images built with it contained increasing numbers of vulnerabilities.


acodeninja (approved):
I'm working on the CI image builder part of this, so this will help out.

PR #202, opened 2023-10-27
DBTP-522 use cmk for s3 bucket encryption

gabelton:
https://uktrade.atlassian.net/browse/DBTP-522

- Creates KMS key and adds to s3 buckets in env addon
- Grants service task role required kms perms


acodeninja (approved):
lgtm

PR #210, opened 2023-11-02
DBTP-548 Fix KMS key reference in S3 policy

WillGibson:
Addresses https://uktrade.atlassian.net/browse/DBTP-548


acodeninja (approved):
LGTM 🚀

PR #211, opened 2023-11-02
DBTP-551 Stop checking image repository name contains service name

WillGibson:
Addresses: https://uktrade.atlassian.net/browse/DBTP-551

When @antroy-madetech and wrote this code we thought all image repository names would include the `/web` etc. part. Turns out this is not the case and I think we can simply remove that requirement.


acodeninja (approved):
Seems reasonable.

PR #212, opened 2023-11-02
DBTP-549 Make `svc deploy` OK with the copilot-bootstrap image

WillGibson:
Addresses: https://uktrade.atlassian.net/browse/DBTP-549

The test for this will fail until https://github.com/uktrade/copilot-tools/pull/211 is merged. I commented out the check for the repository containing the service name temporarily for the purposes of verifying this change. Obviously I will not merge this pull request before https://github.com/uktrade/copilot-tools/pull/211 is merged into `main` then this branch and I've seen the tests pass. I don't think that should stop anyone being able to review and approve the change.


tests/copilot_helper/utils/fixtures/test_service_manifest_with_copilot_bootstrap_image.yml
```diff
@@ -0,0 +1,112 @@
...
+  appconfig:
+    port: 2772
+    image: public.ecr.aws/aws-appconfig/aws-appconfig-agent:2.x
+    essential: true
+    variables:
+      ROLE_ARN: arn:aws:iam::763451185160:role/AppConfigIpFilterRole
+
+# Configuration for your containers and service.
+image:
+  location: public.ecr.aws/uktrade/copilot-bootstrap:latest
```
WillGibson, line 52: A whole new file for only one line difference 😬 

I don't like it 🤮 

I'm wondering whether it would be better to just grab `test_service_manifest.yml`s contents and change this one line before writing the `manifest.yml` used in the test.

The same could be said for `test_service_manifest_with_mismatched_service.yml`.

If we think my suggestion is worthy, I feel it should be done under its own ticket as there are undoubtedly loads of other places we could apply the same approach to reduce the amount of files we need to maintain.

acodeninja: Another option is to use a similar approach to [ci-image-builder](https://github.com/uktrade/ci-image-builder/blob/main/test/image_builder/codebase/test_codebase.py#L29) and have functions that can write to the fakefs a parameterised version of the file. This would also have the benefit of reducing file IO which is one of the reasons the test suite takes a full 30 seconds to run...

WillGibson: https://uktrade.atlassian.net/browse/DBTP-552

The approach @antroy-madetech and I took in [the pipeline command tests](https://github.com/uktrade/copilot-tools/blob/main/tests/copilot_helper/test_command_pipeline.py#L133-L141) to stop the tests being so brittle about whitespace chnages, or adding todo comments etc. my be relevant too.

acodeninja (approved):
LGTM

PR #213, opened 2023-11-02
DBTP-544 - Fix custom resource deletion hanging

yusufsheiqh:
## Context

- Add dependencies to PostgreSQL user custom resource
  - The main/parent CloudFormation stack has the internet gateway, public routes, and public subnets, so when a `copilot env delete` is executed, the resources in the root and nested stacks are delete in parallel. This means the Lambda Function loses access to the database and also the internet without these additional dependencies.
  - This change means the custom resource is deleted first along with the Parameter Store secret for the PostgreSQL application user when running `copilot env delete --name <environment>`.
- Update package version from `0.1.72` -> `0.1.73`

## Additional information

This is a far from ideal solution. The reason to take this approach rather than the `dependsOn` attribute within the Custom Resource is because this attribute only allows resources from the **same** stack. More detail in this [Jira comment](https://uktrade.atlassian.net/browse/DBTP-544?focusedCommentId=127364) as well.

In the future, we should consider provisioning database application users for each service at the service-level rather than the environment-level, as mentioned by @acodeninja in this [Jira comment](https://uktrade.atlassian.net/browse/DBTP-476?focusedCommentId=127099).

### DIFF
--- a/dbt_copilot_helper/templates/addons/env/aurora-postgres.yml
+++ b/dbt_copilot_helper/templates/addons/env/aurora-postgres.yml
@@ -12,7 +12,19 @@ Parameters:
   # Parameters from the parent stack brought in via addons.parameters.yml...
   EnvironmentSecurityGroup:
     Type: String
+  DefaultPublicRoute:
+    Type: String
+  InternetGateway:
+    Type: String
+  InternetGatewayAttachment:
+    Type: String
   PrivateSubnets:
+    Type: String
+  PublicRouteTable:
+    Type: String
+  PublicSubnet1RouteTableAssociation:
+    Type: String
+  PublicSubnet2RouteTableAssociation:
     Type: String
   VpcId:
     Type: String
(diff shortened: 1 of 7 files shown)


dbt_copilot_helper/templates/addons/env/aurora-postgres.yml
```diff
@@ -417,6 +435,22 @@ Resources:
...
+      - AdditionalPrivateRouteTable2
+      - AdditionalPrivateRouteTable2Association
+      - {{ service.prefix }}SecretAuroraClusterAttachment
+      - {{ service.prefix }}EnvironmentIngress
+      - {{ service.prefix }}SecretsManagerIngress
+      - {{ service.prefix }}LambdaIngress
+      - {{ service.prefix }}SecretsManagerEgress
+      - {{ service.prefix }}LambdaEgress
+      - {{ service.prefix }}HTTPSEgress
+      - {{ service.prefix }}KeyAlias
```
WillGibson, line 453: Wow, that's quite the list!

acodeninja (approved):
As discussed a little inelegant, but it does the job!

WillGibson (approved):
Based on your description, I think you could create a spike ticket off the back of this to work out the more elegant solution, maybe taking it in with the environment level Nat Gateways ticket.

dbt_copilot_helper/templates/addons/env/aurora-postgres.yml
yusufsheiqh: This is probably the spike ticket which we already have: https://uktrade.atlassian.net/browse/DBTP-476

PR #215, opened 2023-11-03
DBTP-545 Remove old school S3 logging to an S3 bucket

WillGibson:
Addresses https://uktrade.atlassian.net/browse/DBTP-545

We have CloudTrail event logging enabled for the buckets so the old school logging to an S3 bucket is redundant.

Tested on my `demodjango` environment.


acodeninja (approved):
Noice!

yusufsheiqh (approved):

PR #218, opened 2023-11-06
Clarify s3-policy example in addons-example.yml

WillGibson:


addons-example.yml
```diff
@@ -47,7 +47,7 @@ my-s3-bucket:
 ##### s3-policy example:
 
 my-s3-bucket-bucket-access:
-  type: s3-policy  # gives the listed services permissions to access an existing bucket
+  type: s3-policy  # gives the listed services permissions to access a bucket created and managed outside of your DBT Platform configuration
```
acodeninja, line 50:
```suggestion
  type: s3-policy  # Gives the listed services permissions to access a bucket created and managed outside of your DBT Platform configuration.
                   # If the bucket is not in the same account as your service, see https://repost.aws/knowledge-center/cross-account-access-s3
```

Small link out to explain cross-account S3 perms.

acodeninja (approved):
Nice change, small suggestion on link out to other useful info.

yusufsheiqh (approved):

PR #219, opened 2023-11-06
DBTP-559 Allow S3 buckets to be retained on deletion of an environment

WillGibson:
Addresses https://uktrade.atlassian.net/browse/DBTP-559

Includes an example of using the fake file system (`fakefs`) to do a more targeted test/assertion without needing to add new fixture files for each test case. This has the advantages of less test code/fixtures and faster tests because less real file system activity.


dbt_copilot_helper/commands/copilot.py
```diff
@@ -200,13 +200,19 @@ def file_with_formatting_options(padding=0):
...
+                )
+
         environment_addon_config = {
-            "secret_name": addon_name.upper().replace("-", "_"),
-            "name": addon_config.get("name", None) or addon_name,
-            "environments": environments,
-            "prefix": camel_case(addon_name),
             "addon_type": addon_type,
             "custom_resources": custom_resources,
+            "environments": environments,
```
WillGibson, line 212: This line is new, the rest is just me sorting it alphabetically

tests/copilot_helper/test_command_copilot.py
```diff
@@ -201,6 +203,65 @@ def test_make_addons_success(
...
+                },
+            }
+        }
+        if not deletion_policy:
+            del addon_file_contents["my-s3-bucket"]["deletion-policy"]
+        if not deletion_policy_override:
+            del addon_file_contents["my-s3-bucket"]["environments"]["development"][
+                "deletion-policy"
+            ]
+        create_test_manifests(dump(addon_file_contents), fakefs)
```
WillGibson, line 251: In the setup above we create a simple `addons.yml` for each test case.

tests/copilot_helper/test_command_copilot.py
```diff
@@ -201,6 +203,61 @@ def test_make_addons_success(
...
+
+        CliRunner().invoke(copilot, ["make-addons"])
+
+        manifest = yaml.safe_load(Path("/copilot/environments/addons/my-s3-bucket.yml").read_text())
+        assert (
+            manifest["Mappings"]["myS3BucketEnvironmentConfigMap"]["development"]["DeletionPolicy"]
+            == expected_deletion_policy
+        )
```
WillGibson, line 259: As the templates in the round are covered elsewhere, we only need to assert on the thing that will be different for each test case.

tests/copilot_helper/conftest.py
```diff
@@ -36,6 +36,9 @@ def fakefs(fs):
     fs.add_real_file(BASE_DIR / "dbt_copilot_helper/addons-template-map.yml")
     fs.add_real_directory(Path(jsonschema.__path__[0]) / "schemas/vocabularies")
 
+    # To avoid 'Could not find a suitable TLS CA certificate bundle...' error
+    fs.add_real_file(BASE_DIR / "venv/lib/python3.11/site-packages/certifi/cacert.pem")
+
     return fs
```
antroy-madetech, line 42: I wonder if there's a better way of finding the cacert. This will break when we upgrade python. Perhaps look in `venv/lib` for the most recent python and use that. Would prob need to be done outside of this function though as I think only the fakefs is visible inside the function.

antroy-madetech (commented):
Only one comment.

antroy-madetech (approved):

PR #222, opened 2023-11-08
DBTP-559 Rename BucketNameMap to EnvironmentConfigMap

WillGibson:
I am adding the deletion policy to it so a more open name makes sense, but I want to keep this noise out of my looming pull request.


WillGibson:
Changed mind about the noise factor

PR #223, opened 2023-11-08
DBTP-557 Tag and Branch based deployments

acodeninja:
- Replace `cfn.patch.yml` with cdk based override that reads directly from `manifest.yml` and `pipelines.yml`.
- Add nodejs 18 as a requirement for this tool.
- Make `mkfile` create non-existent directories if needed.
- Add extra validation for `codebases` property on pipeline config to ensure each environment is targeted once per codebase and that either `environments` or `codebases` is present.


yusufsheiqh (approved):

tests/copilot_helper/test_command_pipeline.py
```diff
@@ -45,39 +46,95 @@ def test_pipeline_generate_with_git_repo_creates_the_pipeline_configuration(
...
+    mocked_boto3_client, tmp_path, switch_to_tmp_dir_and_copy_fixtures
+):
+    mock_codestar_connections_boto_client(mocked_boto3_client, ["test-app"])
+    setup_git_repository()
+    environments_files = setup_output_file_paths_for_environments(tmp_path)
+    codebases_files = setup_output_file_paths_for_codebases(tmp_path)
+
+    result = CliRunner().invoke(generate)
+    for file in environments_files + codebases_files:
+        assert_file_created_in_stdout(file, result, tmp_path)
```
WillGibson, line 132: I wonder if it's worth slipping an extra file in after this and asserting that it does not exist after the next generate?

acodeninja: We've not implemented "clean up before adding files" as yet, so leaving this out.

dbt_copilot_helper/templates/pipelines/codebase/overrides/buildspec.deploy.yml
```diff
@@ -0,0 +1,40 @@
+version: 0.2
+
+env:
+  git-credential-helper: yes
+  variables:
+    COLOR: false
```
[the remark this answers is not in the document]
acodeninja: Because the codebuild job doesn't parse ansi escape codes for colors so we would end up with `\u001b[31mSome warning text\u001b[0m` or similar in the logs.

WillGibson (approved):

PR #224, opened 2023-11-09
DBTP-432 `copilot-helper` commands should not take credentials as arguments

gabelton:
https://uktrade.atlassian.net/browse/DBTP-432

Prevent users from entering secret credentials as command line options.

Documentation PR to follow.


acodeninja (approved):
Will need a version bump otherwise good work!

PR #227, opened 2023-11-09
DBTP-336 Allow overriding of default Postgres version

gabelton:
Already done for aurora, this PR adds the same logic for RDS postgres


yusufsheiqh (commented):
How do we determine which PostgreSQL versions are supported? Are there accompanying docs or will we add which versions of PostgreSQL we want to support as part of a validation check?

gabelton:
> How do we determine which PostgreSQL versions are supported? Are there accompanying docs or will we add which versions of PostgreSQL we want to support as part of a validation check?

`aws rds describe-db-engine-versions --engine postgres`  gives a list of supported versions. We could use that as part of a validation check, although I imagine that the cloudformation will raise that same validation error on `copilot env deploy` anyway. (Probs better to validate before it gets to that point and has to rollback though).

None of this exists for current aurora logic either though, so would propose doing that validation work as a separate ticket.

yusufsheiqh:
> > How do we determine which PostgreSQL versions are supported? Are there accompanying docs or will we add which versions of PostgreSQL we want to support as part of a validation check?
> 
> `aws rds describe-db-engine-versions --engine postgres` gives a list of supported versions. We could use that as part of a validation check, although I imagine that the cloudformation will raise that same validation error on `copilot env deploy` anyway. (Probs better to validate before it gets to that point and has to rollback though).
> 
> None of this exists for current aurora logic either though, so would propose doing that validation work as a separate ticket.

@gabelton I am happy for that to be another ticket. I think as a Platform, we can determine which versions are supported, not supported, and deprecated - this means we can support a subset of the versions supported by AWS potentially.

yusufsheiqh:
> @yusufsheiqh makes sense. Is that something we want to do now? I haven't got a sense of which versions we'd want to support

@gabelton Probably not. I would assume that would be something to discuss with the team as part of the new ticket.

gabelton:
> aws rds describe-db-engine-versions --engine postgres

@yusufsheiqh I've created a ticket. Feel free to edit or label as you see fit https://uktrade.atlassian.net/browse/DBTP-567

yusufsheiqh (approved):

PR #228, opened 2023-11-10
DBTP-295 Only add the required parameters to `addons.parameters.yml`

WillGibson:
Addresses https://uktrade.atlassian.net/browse/DBTP-295

Includes some other refactorings, which are mostly commented on inline.


tests/copilot_helper/test_command_copilot.py
```diff
@@ -67,66 +67,76 @@
 
 
 class TestMakeAddonCommand:
+    @patch(
+        "dbt_copilot_helper.utils.versioning.running_as_installed_package",
+        new=Mock(return_value=True),
+    )
+    def test_validate_version_called_once(self, validate_version):
```
WillGibson, line 74: It felt like it was worth separating this out to it's own test and not worrying about it in the rest of them.

tests/copilot_helper/test_command_copilot.py
```diff
@@ -67,8 +67,18 @@
...
+        result = CliRunner().invoke(copilot, ["make-addons"])
+
+        assert result.exit_code == 1
+        validate_version.assert_called_once()
+
     @pytest.mark.parametrize(
-        "addon_file,expected_env_addons,expected_service_addons,expect_db_warning",
+        "addon_file, expected_env_addons, expected_service_addons, expect_db_warning",
```
WillGibson, line 81: More readable with the spaces.

tests/copilot_helper/test_command_copilot.py
```diff
@@ -290,28 +308,13 @@ def test_make_addons_s3_deletion_policy(
...
-                        "deletion-policy": deletion_policy_override,
-                    },
-                },
-            }
-        }
-        if not deletion_policy:
-            del addon_file_contents["my-s3-bucket"]["deletion-policy"]
-        if not deletion_policy_override:
-            del addon_file_contents["my-s3-bucket"]["environments"]["development"][
+        addon_file_contents = yaml.safe_load(S3_STORAGE_CONTENTS)
```
WillGibson, line 311: Realised we could use this and add to it instead of what I did before.

dbt_copilot_helper/commands/copilot.py
```diff
@@ -262,7 +274,7 @@ def _generate_env_addons(
     for addon in addon_template_map[addon_type].get("env", []):
         template = templates.get_template(addon["template"])
 
-        contents = template.render({"service": environment_addon_config, "addons": addons})
+        contents = template.render({"addon_config": environment_addon_config, "addons": addons})
```
WillGibson, line 277: This has been bugging me for a while, been meaning to do a separate pull request for it, but I couldn't leave it any more.

tests/copilot_helper/test_command_copilot.py
```diff
@@ -478,16 +480,17 @@ def test_exit_if_no_local_copilot_environments(self, fakefs):
...
         result = CliRunner().invoke(copilot, ["make-addons"])
 
         assert result.exit_code == 0
-        assert (
-            "File copilot/environments/addons/addons.parameters.yml created" in result.output
-        ), f"addons.parameters.yml should be included for {addon_type}"
+        if has_postgres_addon:
+            assert (
+                "File copilot/environments/addons/addons.parameters.yml created" in result.output
+            ), f"addons.parameters.yml should be included for Postgres addons"
```
antroy-madetech, line 493: This assertion (and the ones below) could perhaps be written: assert ("File...created" in result.output == has_postgres_addon, "addons...")

Then it will be testing the presence of the notification (or content in the ones below) if there's a postgres addon, and the absence if not.

antroy-madetech (approved):

tests/copilot_helper/test_command_copilot.py
WillGibson: Actually, that if above should not be there, commit incoming for removing that one and kind of taking the other in via a helper method to keep the test output clearer.

antroy-madetech (approved):

PR #232, opened 2023-11-10
DBTP-362 minor bug check domain.py should only check dev or prod

gabelton:
https://uktrade.atlassian.net/browse/DBTP-362

Filter environments in check-domain command based on `--domain-profile` flag

### DIFF
--- a/dbt_copilot_helper/commands/dns.py
+++ b/dbt_copilot_helper/commands/dns.py
@@ -497,6 +497,16 @@ def check_domain(domain_profile, project_profile, base_domain, env):
                         click.secho("Domains listed in manifest file", fg="cyan", underline=True)
 
                         environments = conf["environments"].items()
+
+                        if domain_profile == "live":
+                            environments = [
+                                e for e in environments if e[0] in ["prod", "production"]
+                            ]
+                        else:
+                            environments = [
+                                e for e in environments if e[0] not in ["prod", "production"]
+                            ]
+
                         if env:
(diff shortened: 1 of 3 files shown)


dbt_copilot_helper/commands/dns.py
```diff
@@ -497,6 +497,16 @@ def check_domain(domain_profile, project_profile, base_domain, env):
...
                         environments = conf["environments"].items()
+
+                        if domain_profile == "live":
+                            environments = [
+                                e for e in environments if e[0] in ["prod", "production"]
+                            ]
+                        else:
+                            environments = [
+                                e for e in environments if e[0] not in ["prod", "production"]
+                            ]
```
acodeninja, line 508: How do we account for environments in the prod account that are no called `prod` or `production`? Thinking about `hotfix` and `training` from the digital workspace side of things. Though maybe they do run on the dev profile for domains. Just something to think about.

gabelton: Fair point. Do we capture the separation of accounts anywhere in the deploy repo or do we rely on dev team knowledge / pipeline setup to manage that?

acodeninja: So you can see which account an environment is deployed to from the parameter store variable `/copilot/applications/<application>/environments/*` in the property `accountID`. Though I'm happy with this going out as is, will approve.

acodeninja (approved):

PR #236, opened 2023-11-13
Rename `service` to `addon_config` in the addon templates

WillGibson:
Because it's been bugging me for for a while.

No need to publish this one, it doesn't actually change the behaviour of `copilot-helper`, just helps make the code make more sense.


antroy-madetech (approved):
Looks reasonable.

yusufsheiqh (approved):

PR #241, opened 2023-11-17
Extra prod domain bug

antroy-madetech:
Ticket is: [DBTP-444](https://uktrade.atlassian.net/browse/DBTP-444)

Copied over from the ticket (agreed on a thread in slack):

- The dev base domain will always be uktrade.digital
- Prod base domains will always be one of: prod.uktrade.digital, great.gov.uk, trade.gov.uk
- For dev, subdomains will always be <name>.<env>.uktrade.digital
- For prod, subdomains will always be <name>.<base_domain. Deeper names will be added as records (such as `v2.app.trade.gov.uk`)
- base_domain can be inferred from the subdomain if we keep a dict of base_domains vs domain_profiles in the codebase.
- We don’t need the functionality to be able to pass in an entire domain_profile, so domain_profile can also be inferred from the subdomain. 
  - Note: this means we can remove both --base-domain and --domain-profile options from the command.
- For the certificates, we need to generate a cert for the largest hosted zone we have created (or already existed)

[DBTP-444]: https://uktrade.atlassian.net/browse/DBTP-444?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dbt_copilot_helper/commands/dns.py
+++ b/dbt_copilot_helper/commands/dns.py
@@ -339,17 +339,17 @@ class InvalidDomainException(Exception):
     pass
 
 
-def _get_subdomains_from_base(base: list[str], subdomain: list[str]):
+def _get_subdomains_from_base(base: list[str], subdomain: list[str]) -> list[list[str]]:
     if base == subdomain:
         return []
     return [subdomain] + _get_subdomains_from_base(base, subdomain[1:])
 
 
-def get_required_subdomains(base_domain: str, subdomain: str):
+def get_required_subdomains(base_domain: str, subdomain: str) -> list[str]:
     """
     We only want to get subdomains up to 4 levels deep.
 
-    Prod base domains should be 3 levels deep so we should create up to one
+    Prod base domains should be 3 levels deep, so we should create up to one
     additional subdomain. Dev base domain is always "uktrade.digital" and should
     have the app and env subdomains created, so 4 levels in either case.
     """
--- a/dbt_copilot_helper/commands/dns.py
+++ b/dbt_copilot_helper/commands/dns.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 import os
+import re
 import time
 from typing import Tuple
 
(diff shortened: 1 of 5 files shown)


dbt_copilot_helper/commands/dns.py
```diff
@@ -24,8 +25,14 @@
 # (run a test before removing)
 
 # Base domain depth
-MAX_DOMAIN_DEPTH = 2
+AVAILABLE_DOMAINS = {
+    "great.gov.uk": "live",
+    "trade.gov.uk": "live",
+    "prod.uktrade.digital": "live",
```
yusufsheiqh, line 31: Has this been moved over to `live`? This was still in `dev` the last time I used this.

dbt_copilot_helper/commands/dns.py
```diff
@@ -247,140 +237,200 @@ def add_records(client, records, subdom_id, action):
...
+    zone_ids = {}
 
-    # We only want to create domains max 2 deep so remove all sub domains in excess
-    parts_to_remove = len(parts) - base_len - MAX_DOMAIN_DEPTH
-    domain_to_create = parts[parts_to_remove:]
-
-    # Walk back domain name
-    for x in reversed(range(len(domain_to_create) - (len(start_domain.split(".")) - 1))):
-        subdom = ".".join(domain_to_create[(x):]) + "."
+    for dom in domains_to_create:
```
yusufsheiqh, line 255: I reckon I would favour `domain` or `subdomain` instead of `dom`.

dbt_copilot_helper/commands/dns.py
```diff
@@ -247,140 +237,200 @@ def add_records(client, records, subdom_id, action):
...
-    domain_client = domain_session.client("route53")
-    acm_client = project_session.client("acm", region_name=AWS_CERT_REGION)
+def _get_subdomains_from_base(base: list[str], subdomain: list[str]) -> list[list[str]]:
+    """
+    Recurse over domain to get a list of subdomains.
 
-    # create the certificate
-    response = domain_client.list_hosted_zones_by_name()
-    hosted_zones = {hz["Name"]: hz for hz in response["HostedZones"]}
+    Note these are ordered smallest to largest so the domains can be created in
```
yusufsheiqh, line 324:
```suggestion
    These are ordered from smallest to largest, so the domains can be created in
```

dbt_copilot_helper/commands/dns.py
antroy-madetech: Yeah, I don't like dom really, but domain shadows the domain() function at the top level. Which is only really an irritation as the IDE complains. I'll rename it to something sensible.

dbt_copilot_helper/commands/dns.py
antroy-madetech: Don't know. I'll ask in the slack channel.

antroy-madetech: Actually, I looked in the dev and live accounts and it is in the dev account.

yusufsheiqh: Yup, shall we change this back to `dev` for the time being?

dbt_copilot_helper/commands/dns.py
```diff
@@ -58,56 +65,34 @@ def _wait_for_certificate_validation(acm_client, certificate_arn, sleep_time=10,
...
-        + click.style(f"{domain}\n", fg="white", bold=True)
+        + click.style(f"{domain_name}\n", fg="white", bold=True)
         + click.style("Do you want to continue?", fg="yellow"),
     ):
         exit()
 
-    domain_to_create = _get_domain_to_create(base_len, domain)
-
-    arn = _request_cert(client, domain)
+    arn = _request_cert(client, domain_name)
```
WillGibson, line 81: Maybe we can rename this to `certificate_arn` for disambiguation?

dbt_copilot_helper/commands/dns.py
```diff
@@ -247,140 +237,202 @@ def add_records(client, records, subdom_id, action):
...
         response = client.create_hosted_zone(
-            Name=subdom,
+            Name=domain_name,
             # Timestamp is on the end because CallerReference must be unique for every call
-            CallerReference=f"{subdom}_from_code_{int(time.time())}",
+            CallerReference=f"{domain_name}_from_code_{int(time.time())}",
         )
         ns_records = response["DelegationSet"]
-        subdom_id = response["HostedZone"]["Id"]
+        dom_zone_id = response["HostedZone"]["Id"]
```
WillGibson, line 273: Could maybe rename `dom_zone_id`s to `subdomain_zone_id` for consistency with other names.

WillGibson (commented):
Looks great at a first cursory pass. Happy to give it a slower pass after lunch.

WillGibson (commented):
A few small wording suggestions

WillGibson (approved):
So good to see this done @antroy-madetech. Nice work.

Only a few small non blocking suggestions from me so approving.

yusufsheiqh (approved):

PR #243, opened 2023-11-20
DBPT-589 conduit across accounts

acodeninja:
- Allow conduit to connect to addons provisioned in accounts other than the main application account (production for example).
- Implementation of an `Application` class and loader allowing easy and lazy-loading session access.
- Minor tweak to `fakefs` fixture to lazy-load real directories.

### DIFF
--- a/dbt_copilot_helper/utils/application.py
+++ b/dbt_copilot_helper/utils/application.py
@@ -1,11 +1,80 @@
+import json
 from pathlib import Path
+from typing import Dict
 
+import boto3
 import yaml
 from yaml.parser import ParserError
 
+from dbt_copilot_helper.utils.aws import get_aws_session_or_abort
+from dbt_copilot_helper.utils.aws import get_profile_name_from_account_id
 from dbt_copilot_helper.utils.files import load_and_validate_config
 from dbt_copilot_helper.utils.messages import abort_with_error
 from dbt_copilot_helper.utils.validation import BOOTSTRAP_SCHEMA
+
+
+class Environment:
+    name: str
+    account_id: str
+    sessions: Dict[str, boto3.Session]
+
+    def __init__(self, name: str, account_id: str, sessions: Dict[str, boto3.Session]):
+        self.name = name
+        self.account_id = account_id
+        self.sessions = sessions
+
+    @property
+    def session(self):
+        if self.account_id not in self.sessions:
+            self.sessions[self.account_id] = get_aws_session_or_abort(
+                get_profile_name_from_account_id(self.account_id),
+            )
+
+        return self.sessions[self.account_id]
+
+
+class Application:
+    name: str
+    environments: Dict[str, Environment]
+
+    def __init__(self, name: str):
+        self.name = name
+        self.environments = {}
+
+    def __str__(self):
+        output = f"Application {self.name} with"
+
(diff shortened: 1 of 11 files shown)


tests/copilot_helper/conftest.py
```diff
@@ -30,20 +31,53 @@
...
+        FIXTURES_DIR / "dummy_aws_config.ini", True, Path.home().joinpath(".aws/config")
+    )
+
     return fs
 
 
+@pytest.fixture(scope="function", autouse=True)
+def mock_application():
+    with patch("dbt_copilot_helper.utils.application.load_application") as app_patch:
+        from dbt_copilot_helper.utils.application import Application
```
acodeninja, line 63: ## Why are we using local imports?

If we used an import at the top of the file then the `load_application` patch would not be enabled for the scope of those imports. We do this in other tests to ensure that the global patching of `dbt_copilot_helper.utils.application.load_application` works, otherwise we would have to locally patch each time using `dbt_copilot_helper.commands.conduit.load_application`.

## TL;DR

Because it just works, innit.

yusufsheiqh (approved):

dbt_copilot_helper/utils/application.py
```diff
@@ -1,13 +1,82 @@
...
+    def session(self):
+        if self.account_id not in self.sessions:
+            self.sessions[self.account_id] = get_aws_session_or_abort(
+                get_profile_name_from_account_id(self.account_id),
+            )
+
+        return self.sessions[self.account_id]
+
+
+class Application:
```
gabelton, line 36: Is it worth adding docstrings to these two classes? Maybe saying how they map or don't map onto AWS concepts

acodeninja: Probably not worth the additional maintenance. I'd be interested to understand what is missing in code and accompanying tests that might help convey how these classes apply to copilot.

gabelton (approved):

tests/copilot_helper/test_command_conduit.py
```diff
@@ -238,11 +251,16 @@ def test_addon_client_is_running_when_no_client_agent_running(
     "addon_name",
     ["postgres", "redis", "opensearch", "rds-postgres"],
 )
-def test_add_stack_delete_policy_to_task_role(mock_stack, addon_name):
+@patch("time.sleep", return_value=None)
```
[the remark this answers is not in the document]
acodeninja: Indeed, by a second from my checks.

tests/copilot_helper/utils/test_application.py
```diff
@@ -26,3 +35,94 @@ def test_getting_an_application_name_from_workspace(fakefs):
...
+    def test_environment_session(self, get_aws_session_or_abort, get_profile_name_from_account_id):
+        environment = Environment("test", "999999999", {})
+
+        self.assertEqual(environment.name, "test")
+        self.assertEqual(environment.account_id, "999999999")
+
+        get_aws_session_or_abort.assert_not_called()
+        get_profile_name_from_account_id.assert_not_called()
+
+        self.assertEqual(environment.session, boto3)
```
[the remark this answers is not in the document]
acodeninja: I think keeping it as a `@property` definition is alright, it is technically a property of the environment even if it does some behind the scenes stuff I think.

PR #246, opened 2023-11-21
DBTP-464 Fix no accounts in pipelines config raising error

acodeninja:
- Fix an error when no accounts are configured (for situations where there is only one account in the application).
- Remove repetition when creating required pipeline stages for existing and newly created pipelines.
- Push up a codebase configuration including information about all pipelines that deploy that codebase.


yusufsheiqh (approved):
🚀

PR #250, opened 2023-11-23
DBTP-597 - Retain conduit logs after session ends

JohnStainsby:
- Add `DeletionPolicy: Retain` to the log group in the conduit task CloudFormation stack to prevent it being deleted when the session ends
- Append a random string to the end of the task name to ensure the log group is created with a unique name. The log group name is derived from the copilot task name. This prevents "resource already exists" errors in subsequent conduit sessions
- Retrieve task name to connect to existing conduit session


dbt_copilot_helper/commands/conduit.py
```diff
@@ -185,19 +202,52 @@ def add_stack_delete_policy_to_task_role(app: Application, env: str, addon_name:
...
+
+    conduit_stack_name = f"task-{task_name}"
+    template = cloudformation_client.get_template(StackName=conduit_stack_name)
+    template_yml = load_yaml(template["TemplateBody"])
+    template_yml["Resources"]["LogGroup"]["DeletionPolicy"] = "Retain"
+    template_yml["Resources"]["TaskNameParameter"] = load_yaml(
+        f"""
+        Type: AWS::SSM::Parameter
+        Properties:
+          Name: /copilot/{app.name}/{env}/secrets/{addon_name}_CONDUIT_TASK_NAME
```
acodeninja, line 217: Maybe separating these parameters from the `secrets` as they are meant for services. Something like `/copilot/{app.name}/{env}/conduits/{addon_name}_CONDUIT_TASK_NAME` might be better.

PR #251, opened 2023-11-24
Fixed a bug where hosted zones were created multiple times

antroy-madetech:
Introduced during refactoring work in DBTP-444. Test added around this case.

### DIFF
--- a/dbt_copilot_helper/commands/dns.py
+++ b/dbt_copilot_helper/commands/dns.py
@@ -137,15 +137,6 @@ def _create_resource_records(cert_record, domain_client, zone_id):
     )
 
 
-def _get_domain_id(domain_client, domain_to_create):
-    click.secho(f"Looking for ID of domain {domain_to_create}...", fg="yellow")
-    hosted_zones = domain_client.list_hosted_zones_by_name()["HostedZones"]
-
-    for hz in hosted_zones:
-        if hz["Name"] == domain_to_create:
-            return hz["Id"]
-
-
 def _certs_for_domain(client, domain):
     # Check if cert is present.
     cert_list = client.list_certificates(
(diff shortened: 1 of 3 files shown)


dbt_copilot_helper/commands/dns.py
```diff
@@ -137,15 +137,6 @@ def _create_resource_records(cert_record, domain_client, zone_id):
...
-def _get_domain_id(domain_client, domain_to_create):
-    click.secho(f"Looking for ID of domain {domain_to_create}...", fg="yellow")
-    hosted_zones = domain_client.list_hosted_zones_by_name()["HostedZones"]
-
-    for hz in hosted_zones:
-        if hz["Name"] == domain_to_create:
-            return hz["Id"]
-
-
 def _certs_for_domain(client, domain):
```
antroy-madetech, line 140: This function is no longer used anywhere, so removed it.

dbt_copilot_helper/commands/dns.py
```diff
@@ -251,42 +242,49 @@ def create_hosted_zones(client, base_domain, subdomain):
...
-            click.style("About to create domain: ", fg="cyan")
-            + click.style(f"{domain_name}\n", fg="white", bold=True)
-            + click.style("Do you want to continue?", fg="cyan"),
-        ):
-            exit()
-
-        click.secho(f"Creating hosted zone for {domain_name}...", fg="yellow")
+        if domain_zone in hosted_zones:
+            click.secho(f"Hosted zone '{domain_zone}' already exists", fg="yellow")
+            zone_ids[domain_name] = hosted_zones[domain_zone]["Id"]
```
antroy-madetech, line 254: I'm assuming here that if we do not need to create the hosted zone because it already exists then we also do not need to copy the records to and from its parent.

WillGibson (approved):

dbt_copilot_helper/commands/dns.py
```diff
@@ -423,17 +421,14 @@ def create_required_zones_and_certs(domain_client, project_client, subdomain, ba
...
-    response = client.list_hosted_zones_by_name()
-    hosted_zones = {hz["Name"]: hz for hz in response["HostedZones"]}
     if zone not in hosted_zones:
         click.secho(
-            f"The base domain: {zone} does not exist in your AWS domain account {hosted_zones}",
+            f"The hosted zone: {zone} does not exist in your AWS domain account {hosted_zones}",
             fg="red",
         )
         exit()
-    click.secho(f"Found hosted zone {zone}", fg="green")
```
antroy-madetech, line 436: This was just noise in the output and didn't really add anything, so removed it.

yusufsheiqh (approved):

PR #258, opened 2023-12-05
Added support for Aurora RDS snapshot recovery

antroy-madetech:
In order to support Aurora recovery to a snapshot, I have added the option to specify a snapshot ID in the addons.yml that will allow this.


tests/copilot_helper/fixtures/make_addons/expected/environments/addons/my-aurora-db.yml
```diff
@@ -44,6 +44,10 @@ Mappings:
       DBMinCapacity: 0.5
       DBMaxCapacity: 8
       DeletionPolicy: Delete
+      SnapshotIdentifier: my-recent-snapshot
+
+Conditions:
+  UseDBSnapshot: !Not [!Equals [!FindInMap [myAuroraDbEnvironmentConfigMap, !Ref Env, SnapshotIdentifier], ""]]
```
acodeninja, line 50: I'm wondering if these conditionals need to be prefixed with the `myAuroraDb` bit, as the addons are merged into one template before being deployed. This could mean that the conditionals override each other. Worth testing to confirm in demodjango where we have both RDS and Aurora.

This is probably fine for now without the straight RDS template in place, but may as well address up front.

tests/copilot_helper/test_command_check_cloudformation.py
```diff
@@ -49,7 +49,7 @@ def application_under_test(copilot_directory):
     CliRunner().invoke(make_config)
     CliRunner().invoke(make_addons)
     yield
-    ensure_directory_does_not_exist(copilot_directory)
+    # ensure_directory_does_not_exist(copilot_directory)
```
acodeninja, line 52: Does this need to be deleted or kept in?

antroy-madetech: Kept in. I commented it out when debugging something so I could see the files...

tests/copilot_helper/fixtures/make_addons/expected/environments/addons/my-aurora-db.yml
antroy-madetech: No, hadn't groked what you were saying. Thought you meant that it would be the same as the postgresRDS DB version when we come to that, and that you couldn't have two conditionals in the same deployment the same name.  Fixed.

yusufsheiqh (approved):

dbt_copilot_helper/templates/addons/env/aurora-postgres.yml
```diff
@@ -45,10 +45,16 @@ Mappings:
       DBMinCapacity: {{ config.min_capacity }}
       DBMaxCapacity: {{ config.max_capacity }}
       DeletionPolicy: {{ config.deletion_policy }}
+{%- if config.snapshot_id %}
+      SnapshotIdentifier: {{ config.snapshot_id }}
+{%- else %}
+      SnapshotIdentifier: ""
+{%- endif %}
```
WillGibson, line 52: Maybe...
```suggestion
      SnapshotIdentifier: {{ config.snapshot_id if config.snapshot_id else "" }}
```

WillGibson (approved):
One small suggestion. Approving anyway as it's not a deal breaker.

Also, is the way this is intended to be used documented elsewhere?

dbt_copilot_helper/templates/addons/env/aurora-postgres.yml
yusufsheiqh: Yes, it is documented in the platform docs. I'm making a few tweaks to it as we speak.

The suggested change doesn't preserve the quotation marks which is needed, but I've committed and tweaked your change slightly in https://github.com/uktrade/copilot-tools/pull/258/commits/83564caceab58e581c5b901dfbe2d08fd187e63c.

PR #261, opened 2023-12-07
Add service level patching for renaming service log group name

ejayesh:
Small change to bootstrap to include service level cfn patching to rename log group name with hypens to /'s

### DIFF
--- a/dbt_copilot_helper/commands/bootstrap.py
+++ b/dbt_copilot_helper/commands/bootstrap.py
@@ -47,6 +47,14 @@ def get_paas_env_vars(client: CloudFoundryClient, paas: str) -> dict:
     return dict(env_vars)
 
 
+def _generate_svc_overrides(base_path, templates, name):
+    click.echo("\n>>> Generating service overrides\n")
+    overrides_path = base_path.joinpath(f"copilot/{name}/overrides")
+    overrides_path.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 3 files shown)


dbt_copilot_helper/templates/svc/overrides/cfn.patches.yml
```diff
@@ -0,0 +1,3 @@
+- op: replace
+  path: /Resources/LogGroup/Properties/LogGroupName
+  value: !Sub '/copilot/${AppName}/${EnvName}/${WorkloadName}'
```
acodeninja, line 3: No new line

dbt_copilot_helper/commands/bootstrap.py
```diff
@@ -94,6 +102,7 @@ def make_config(directory="."):
         )
         name = service["name"]
         (base_path / f"copilot/{name}/addons/").mkdir(parents=True, exist_ok=True)
+        _generate_svc_overrides(base_path, templates, name)
```
acodeninja, line 105: Not something to address here, but what about completely new applications not created with a bootstrap file.

ejayesh: yep will create a new ticket to pick that up

dbt_copilot_helper/templates/svc/overrides/cfn.patches.yml
ejayesh: added.

PR #263, opened 2023-12-11
DBTP-625 - Update ALB log bucket name

JohnStainsby:
Update ALB log bucket name to `dbt-access-logs` to match bucket in central logging account


lgarvey (approved):
LGTM 🚀

PR #266, opened 2023-12-11
Dbtp 632 fix loggroup bug

ejayesh:
Multiple fixes:
1.  Added dependencies for subscription filter creation.
2. Fix bug where new DB were not getting created due to incompatability with AWS::LanguageExtensions.
3. Rename Filtername to have same format as Loggroup name.


antroy-madetech (approved):
LGTM

PR #269, opened 2023-12-14
DBTP-620 load balancer maintenance pages

acodeninja:
Adds two commands:

`copilot-helper environment offline --app <app> --env <env> [--template (default|migration)]`

Take an environment offline by setting a fixed response at the load balancer.

`copilot-helper environment online --app <app> --env <env>`

Take an environment online by removing the currently set maintenance page.

### DIFF
--- a/dbt_copilot_helper/COMMANDS.md
+++ b/dbt_copilot_helper/COMMANDS.md
@@ -28,6 +28,9 @@
 - [copilot-helper domain](#copilot-helper-domain)
 - [copilot-helper domain configure](#copilot-helper-domain-configure)
 - [copilot-helper domain assign](#copilot-helper-domain-assign)
+- [copilot-helper environment](#copilot-helper-environment)
+- [copilot-helper environment offline](#copilot-helper-environment-offline)
+- [copilot-helper environment online](#copilot-helper-environment-online)
 - [copilot-helper pipeline](#copilot-helper-pipeline)
 - [copilot-helper pipeline generate](#copilot-helper-pipeline-generate)
 - [copilot-helper svc](#copilot-helper-svc)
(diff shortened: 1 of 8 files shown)


dbt_copilot_helper/templates/svc/maintenance_pages/migration.html
```diff
@@ -0,0 +1,23 @@
...
+<header class="govuk-header " role="banner" data-module="govuk-header">
+    <div class="govuk-header__container govuk-width-container">
+        <div class="govuk-header__logo"></div>
+    </div>
+</header>
+<div class="govuk-width-container">
+    <main class="govuk-main-wrapper" id="main-content" role="main">
+        <h1 class="govuk-heading-l">Scheduled Maintenance</h1>
+        <p class="govuk-body-l">
+            The service is currently unavailable. Don't worry, we'll be back by the end of the day.
```
yusufsheiqh, line 18: Would be nice if the body message could be configurable in some way.

dbt_copilot_helper/COMMANDS.md
```diff
@@ -706,6 +710,72 @@ copilot-helper domain assign --app <app> --env <env> --svc <svc>
...
+copilot-helper environment offline --app <app> --env <env> [--template (default|migration)] 
+```
+
+## Options
+
+- `--app <text>`
+
+- `--env <text>`
+
+- `--template <choice>` _Defaults to default._
```
yusufsheiqh, line 753: `Defaults to default` doesn't read well, but I am not sure if we can do anything about that.

dbt_copilot_helper/templates/svc/maintenance_pages/migration.html
acodeninja: Yeah that would be nice, but this and the default page cover our current use cases.

dbt_copilot_helper/COMMANDS.md
acodeninja: Other than change `default` to something else, but I think this makes sense.

yusufsheiqh (approved):
I am glad that we are using the full word `environment` otherwise it can be confusing with `copilot env...`!

PR #270, opened 2023-12-14
Bump copilot cli version in generated templates

acodeninja:
So we avoid the downgrade errors in ci.

### DIFF
--- a/dbt_copilot_helper/templates/pipelines/environments/buildspec.yml
+++ b/dbt_copilot_helper/templates/pipelines/environments/buildspec.yml
@@ -3,7 +3,7 @@
 version: 0.2
 env:
   variables:
-    COPILOT_VERSION: 1.31.0
+    COPILOT_VERSION: 1.32.0
     DYFF_VERSION: 1.5.8
 phases:
   install:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ line-length = 100
 
 [tool.poetry]
 name = "dbt-copilot-tools"
-version = "0.1.110"
+version = "0.1.112"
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -3,7 +3,7 @@ line-length = 100
 
 [tool.poetry]
 name = "dbt-copilot-tools"
-version = "0.1.110"
+version = "0.1.112"
```
yusufsheiqh, line 6: Why has this been bumped up two versions? 🤔 

```suggestion
version = "0.1.111"
```

dbt_copilot_helper/templates/pipelines/environments/buildspec.yml
```diff
@@ -3,7 +3,7 @@
 version: 0.2
 env:
   variables:
-    COPILOT_VERSION: 1.31.0
+    COPILOT_VERSION: 1.32.0
```
yusufsheiqh, line 6: Are we keeping this up to date in terms of minor versions only? The latest version is `1.32.1` otherwise.

pyproject.toml
acodeninja: Because I have another PR in front of this one...

dbt_copilot_helper/templates/pipelines/environments/buildspec.yml
acodeninja: This matches the ci-image-builder version currently in use. We do need to come up with a better strategy for this and ensure we are all running the same version.

yusufsheiqh: That makes sense. Would be easier if we merge the `ci-image-builder` into this repo!

yusufsheiqh (approved):

PR #272, opened 2023-12-14
move prod domain to live aws

ejayesh:
very quick PR to move the prod.uktrade.digital domain to live, as this is where the R53 domain is now.


acodeninja (approved):
🚀

PR #273, opened 2023-12-15
style: DBTP-620 Add a nicer styled copilot-bootstrap landing page

acodeninja:
Adds a consistent style to the bootstrap landing page.


images/copilot-bootstrap/index.html
yusufsheiqh, line 1: Since it's the bootstrap image, it won't have the GOV.UK Design System, right? Not sure if all of styling would work...

yusufsheiqh: Although I do see it's linking the SSO stylesheet. 🤔

acodeninja: Yeup, I could not find a CDN for GOV.UK styles so just went for the service that should have the highest uptime.

WillGibson (approved):

PR #278, opened 2024-01-08
DBTP-575 Add snapshot ability to the RDS Postgres addon

antroy-madetech:
Adds in logic to the templates to support restoring a snapshot DB to RDS postgres

### DIFF
--- a/dbt_copilot_helper/templates/addons/env/rds-postgres.yml
+++ b/dbt_copilot_helper/templates/addons/env/rds-postgres.yml
@@ -48,10 +48,12 @@ Mappings:
       MaxAllocatedStorage: {{ (config.volume_size * 1.26)|round|int }}
       StorageType: {% if env_name == "prod" %}io1{% else %}gp3{% endif %}
       DeletionPolicy: {{ config.deletion_policy }}
+      SnapshotIdentifier: {{ config.snapshot_id if config.snapshot_id else '""' }}
 {%- endfor %}
 
 Conditions:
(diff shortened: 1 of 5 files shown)


dbt_copilot_helper/templates/addons/env/rds-postgres.yml
```diff
@@ -211,28 +213,57 @@ Resources:
       AllowMajorVersionUpgrade: false
       AutoMinorVersionUpgrade: true
       BackupRetentionPeriod: 8
-      EnablePerformanceInsights: true
+      EnablePerformanceInsights:
+        !If [
+          {{ addon_config.prefix }}UseSnapshot,
+          !Ref AWS::NoValue,
+          true
+        ]
```
WillGibson, line 221: Why has this become conditional?

WillGibson: And the stuff further down, why are we using `AWS::NoValue` if we are using snapshots?

antroy-madetech: Certain fields (MasterPassword, MasterUsername and the others in here) are mutually exclusive with DBSnapshotId. However the definitions are global to all envs, so cannot be removed for just one env.

Hence conditionals. AWS::NoValue makes CloudFormation treat the field as if it isn't there (unlike setting it to an empty string for example)

(It was the same for the Aurora DR if you take a look, except for slightly different fields due to it being a DB rather than a cluster)

WillGibson (approved):

PR #280, opened 2024-01-10
DBTP-659 - Add deletion-protection property to copilot-tools

JohnStainsby:
- Add deletion-protection property for aurora and rds postgres addons

### DIFF
--- a/addons-example.yml
+++ b/addons-example.yml
@@ -11,11 +11,13 @@
 #       some-setting: x
 #       another-setting: y
 #       plan: a-valid-plan  # some addon types support plans/t-shirt sizes.  These are listed in addon-plans.yaml in this repo.
+#       deletion-protection: X # true|false rds-postgres and aurora-postgres only
 #
 #     production:  # you can override defaults on a per environment basis
 #       plan: some-other-plan  # Any of the keys defined in a addon plan can also be overridden
 #       volume-size: 200
 #       deletion-policy: Retain
+#       deletion-protection: true
 
 
 ##### s3 bucket example:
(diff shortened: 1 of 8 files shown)


addons-example.yml
```diff
@@ -90,8 +92,11 @@ my-rds-db:
   environments:
     default:
       plan: small-ha
-      deletion-policy: Retain # Delete|Retain|Snapshot
```
yusufsheiqh, line 93: Do we still need this policy? Or does protection take priority?

yusufsheiqh: I can see both are set for the `production` env.

JohnStainsby: The `deletion-policy` is specific to preventing CloudFormation deleting a resource. Presumably `deletion-protection` would also stop that scenario but I'm not sure. There may be situations where you'd want one vs the other.

addons-example.yml
```diff
@@ -11,11 +11,13 @@
 #       some-setting: x
 #       another-setting: y
 #       plan: a-valid-plan  # some addon types support plans/t-shirt sizes.  These are listed in addon-plans.yaml in this repo.
+#       deletion-protection: X # true|false rds-postgres and aurora-postgres only
```
WillGibson, line 14: If this is an example, perhaps...
```suggestion
#       deletion-protection: false # true|false rds-postgres and aurora-postgres only
```

addons-example.yml
WillGibson: It does feel a bit weird having different settings overlapping like this, but I think we can thank AWS for that.

JohnStainsby: It is strange but I don't think there's much we can do about it. I tested it and `deletion-protection` does stop CloudFormation from deleting the database but it gets stuck in _update-cleanup-in-progress_ for ages.

Also `deletion-protection` is only an option for databases so `deletion-policy` would still have to be set on the other addons anyway.

WillGibson (approved):

PR #281, opened 2024-01-12
DBTP-648 must specify `--env` when migrating secrets

acodeninja:
No tests removed as none exist for running the command without `--env` set.


gabelton (approved):
Not a blocker, but I would've expected a test to fail with this kind of change. Perhaps worth a review in another ticket, to check that tests include minimal arguments for a command.

PR #282, opened 2024-01-12
Use tag-latest rather than latest as the default tag

antroy-madetech:
Changed behaviour to reject latest tag and to default to the tag-latest tag.

Reject the tag-latest tag if there is no corresponding commit- tag

### DIFF
--- a/dbt_copilot_helper/commands/svc.py
+++ b/dbt_copilot_helper/commands/svc.py
@@ -24,9 +24,16 @@ def svc():
 @svc.command()
 @click.option("--env", type=str, required=True)
 @click.option("--name", type=str, required=True)
-@click.option("--image-tag", type=str, required=False, show_default=True, default="latest")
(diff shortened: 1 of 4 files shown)


dbt_copilot_helper/commands/svc.py
```diff
@@ -24,19 +24,26 @@ def svc():
...
 @click.option("--env", type=str, required=True)
 @click.option("--name", type=str, required=True)
-@click.option("--image-tag", type=str, required=False, show_default=True, default="latest")
+@click.option("--image-tag", type=str, required=False, show_default=True, default="tag-latest")
 def deploy(env, name, image_tag):
     """Deploy image tag to a service, defaults to image tagged latest."""
 
+    if image_tag == "latest":
+        abort_with_message(
+            f'Releasing tag "latest" is not supported. Use tag "tag-latest" instead.'
```
WillGibson, line 33: ❤️

WillGibson (approved):

PR #283, opened 2024-01-15
DBTP-642 Missing log group name YAML patch override

gabelton:
Once a `subscription-filter` env addon has been created, each service in the app will need the log group name override file to be created. Although the override lives at svc level, it relies on an env addon existing and, therefore, it makes more sense to create the override as part of the make-addons command.


tests/copilot_helper/test_command_bootstrap.py
```diff
@@ -115,6 +115,8 @@ def test_make_config(tmp_path):
     with open(str(tmp_path / "copilot/test-service/manifest.yml")) as service:
         assert service.read() == test_service_manifest
 
+    assert os.path.exists(str(tmp_path / "copilot/test-service/addons"))
```
gabelton, line 118: This test didn't make any assertions about the addons or overrides folders being created

gabelton:
Think would also solve https://uktrade.atlassian.net/browse/DBTP-631

dbt_copilot_helper/commands/copilot.py
```diff
@@ -186,6 +194,11 @@ def make_addons(directory="."):
     _cleanup_old_files(config, output_dir, env_addons_path)
     custom_resources = _get_custom_resources()
 
+    svc_names = list_copilot_local_services()
+    base_path = Path(directory)
+    for svc_name in svc_names:
+        _generate_svc_overrides(base_path, templates, svc_name)
```
WillGibson, line 200: Could this be called from `_generate_service_addons()`?

tests/copilot_helper/fixtures/make_addons/expected/web/overrides/cfn.patches.yml
```diff
@@ -0,0 +1,4 @@
+- op: replace
+  path: /Resources/LogGroup/Properties/LogGroupName
+  value: !Sub '/copilot/${AppName}/${EnvName}/${WorkloadName}'
+
```
WillGibson, line 4:
```suggestion
```

dbt_copilot_helper/commands/copilot.py
gabelton: I did look at that option, but I think we will miss some services, unless they happen to be defined under an addon in addons.yml. e.g.

```
dw-s3-bucket:
  type: s3
  services:
    - "s3proxy"
    - "beat"
    - "web"
    - "worker"
```
Not all applications will have a setup like DW

WillGibson: Oh yeah, I think I slightly misunderstood what was going on down below.

WillGibson (approved):

antroy-madetech (approved):

PR #285, opened 2024-01-16
DBTP-660 Add OpenTelemetry variables to manifest files

yusufsheiqh:
## Context

- Add OTel env vars to manifest template files
  - This is to setup observability (X-Ray) for services.
  - The sampling ratio is set at a default of 5%.
- Add `Dockerfile.debian` file
  - This is currently used for `copilot-python` as the Alpine Dockerfile does not support installing gRPC within a `poetry` context. This image should also be used for this repo in the near future to avoid issues like this from reoccurring.

### DIFF
--- a/Dockerfile.debian
+++ b/Dockerfile.debian
@@ -1,0 +1,36 @@
+FROM debian:stable-slim
+
+ARG PYTHON_VERSIONS="3.9 3.10 3.11"
+
+ENV DEBIAN_FRONTEND noninteractive
+ENV PYENV_ROOT="$HOME/.pyenv"
+ENV PATH="$PYENV_ROOT/shims:$PYENV_ROOT/bin:$PATH"
+
+RUN apt-get update \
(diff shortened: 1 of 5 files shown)


acodeninja (commented):
Looking good, my only wonder is if this might get expensive if we enable for everything?

yusufsheiqh:
> Looking good, my only wonder is if this might get expensive if we enable for everything?

@acodeninja Would you suggest enabling on the `web`/public facing services as a default?

acodeninja:
> @acodeninja Would you suggest enabling on the `web`/public facing services as a default?

I reckon we should to start off with, from my understanding though this will not be enabled unless the process is prefixed with the otel collector application.

acodeninja (approved):

WillGibson (approved):

ejayesh (approved):
lgtm

PR #286, opened 2024-01-17
DBTP-675 make-addons generates addons and overrides for scheduled jobs

gabelton:
Add an `is_service` check to `list_copilot_local_services`, so that `make-addons` can distinguish between services and scheduled jobs.

### DIFF
--- a/dbt_copilot_helper/commands/copilot.py
+++ b/dbt_copilot_helper/commands/copilot.py
@@ -44,8 +44,11 @@ def list_copilot_local_environments():
 def is_service(path: PosixPath) -> bool:
     with open(path) as manifest_file:
         data = yaml.safe_load(manifest_file)
-        if not data:
-            return False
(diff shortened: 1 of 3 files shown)


dbt_copilot_helper/commands/copilot.py
```diff
@@ -25,15 +26,36 @@
...
     return [
         path.parent.parts[-1] for path in Path("./copilot/environments/").glob("*/manifest.yml")
     ]
 
 
+def is_service(path: PosixPath) -> bool:
+    with open(path) as manifest_file:
+        data = yaml.safe_load(manifest_file)
+        if not data:
+            return False
```
WillGibson, line 48: Somethings gone wrong rather than it's just not a service. Feels like it should exit the command with an error to me.

WillGibson (approved):

PR #287, opened 2024-01-22
DBTP-672 - Domain configure error creating certificate

JohnStainsby:
- Add test for creating a certificate when a record already exists
- Fix error in `domain configure` command when re-creating a certificate and a record already exists

### DIFF
--- a/dbt_copilot_helper/commands/dns.py
+++ b/dbt_copilot_helper/commands/dns.py
@@ -117,24 +117,34 @@ def _get_existing_cert_if_one_exists_and_cleanup_bad_certs(client, domain):
 
 
 def _create_resource_records(cert_record, domain_client, zone_id):
-    domain_client.change_resource_record_sets(
-        HostedZoneId=zone_id,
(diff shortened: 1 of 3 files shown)


dbt_copilot_helper/commands/dns.py
```diff
@@ -117,24 +117,34 @@ def _get_existing_cert_if_one_exists_and_cleanup_bad_certs(client, domain):
...
-                            {"Value": cert_record["Value"]},
-                        ],
+    resource_records = domain_client.list_resource_record_sets(HostedZoneId=zone_id)[
+        "ResourceRecordSets"
+    ]
+
+    exists = False
+    for record in resource_records:
+        if record["Type"] == cert_record["Type"] and record["Name"] == cert_record["Name"]:
+            exists = True
```
yusufsheiqh, line 127: Why not do an early return here? You can avoid the boolean logic.

yusufsheiqh (approved):

PR #289, opened 2024-01-25
DBTP-280 - Storage validation

JohnStainsby:
- Add opensearch storage volume validation per plan type
- Remove properties that shouldn't be overridden
- Fix invalid options and inconsistencies in addon plans


dbt_copilot_helper/addon-plans.yml
```diff
@@ -1,187 +1,186 @@
...
     master: false
     instance: m6g.2xlarge.search
 
 # RDS/Postgres Instances
 rds-postgres:
-  # 5GB Storage, DB Instance Class: db.t3.micro
+  # 2v CPU, 1GB RAM, 5GB Storage
   tiny:
     volume-size: 5
     replicas: 0
```
JohnStainsby, line 133: Not sure why a "replicas" integer is used here. In the Cloudformation template ouput it enables MultiAZ if the value is greater than 0, the actual number doesn't matter. It could probably be `multi-az: true|false` for more clarity.

WillGibson (approved):

antroy-madetech (commented):
Don't merge this just yet. My branch has a bunch of changes to validation and no longer uses the json-schema file. We should collaborate on this next week.

WillGibson (approved):

PR #290, opened 2024-01-25
Better validation errors for make_addons

antroy-madetech:
I have changed the validation from json-schema validation to python schema module validation in order to get better error messages out of the tool and to make adding new validation easier. This will bring it into line with other schema module validation that we are using elsewhere.


dbt_copilot_helper/utils/validation.py
```diff
@@ -17,6 +18,62 @@ def validator(string):
...
+            errors[addon_name] = f"Error in {addon_name}: {ex.code}"
+
+    return errors
+
+
+def int_between(lower, upper):
+    def is_between(value):
+        if isinstance(value, int) and lower <= value <= upper:
+            return True
+        raise SchemaError(f"should be an int between {lower} and {upper}")
```
JohnStainsby, line 54:
```suggestion
        raise SchemaError(f"Should be an integer between {lower} and {upper}")
```

dbt_copilot_helper/utils/validation.py
```diff
@@ -17,6 +18,62 @@ def validator(string):
...
+
+
+def float_between_with_halfstep(lower, upper):
+    def is_between(value):
+        is_number = isinstance(value, int) or isinstance(value, float)
+        is_half_step = re.match(r"^\d+(\.[05])?$", str(value))
+
+        if is_number and is_half_step and lower <= value <= upper:
+            return True
+        raise SchemaError(f"should be a number between {lower} and {upper} in half steps")
```
JohnStainsby, line 66:
```suggestion
        raise SchemaError(f"Should be a number between {lower} and {upper} in increments of 0.5")
```

dbt_copilot_helper/utils/validation.py
```diff
@@ -142,3 +199,197 @@ def validator(string):
         ],
     },
 )
+
+NUMBER = Or(int, float)
+DELETION_POLICY = Or("Delete", "Retain", "Snapshot")
```
JohnStainsby, line 204: The "Snapshot" deletion policy is only applicable to redis and postgres addons. It's still valid Cloudformation but defaults to delete if the resource doesn't support it. Might be worth protecting against any assumptions that snapshot would work for s3/opensearch?

```suggestion
DELETION_POLICY = Or("Delete", "Retain")
DELETION_POLICY_DATABASE = Or("Delete", "Retain", "Snapshot")
```

dbt_copilot_helper/utils/validation.py
```diff
@@ -142,3 +199,197 @@ def validator(string):
...
+    }
+)
+
+REDIS_SCHEMA = Schema(
+    {
+        "type": "redis",
+        Optional("deletion-policy"): DELETION_POLICY,
+        Optional("environments"): {
+            ENV_NAME: {
+                Optional("plan"): REDIS_PLANS,
```
JohnStainsby, line 340: Probably outside the scope of this PR but some properties like "plan" and "engine" are required but not for every environment.

Maybe we need to enforce the "default" environment with the required properties?

JohnStainsby (commented):
Nice, certainly easier to read than the json schema!

Some comments/suggestions, I will add in the schema changes in my PR once this is merged.

dbt_copilot_helper/utils/validation.py
```diff
@@ -17,6 +18,62 @@ def validator(string):
...
+            return True
+        raise SchemaError(f"should be a number between {lower} and {upper} in half steps")
+
+    return is_between
+
+
+ENV_NAME = Regex(
+    r"^[a-zA-Z][a-zA-Z0-9]*$",
+    error="Environment name {} is invalid: names must only contain alphanumeric characters."
+    # It would be nice if this worked, but it doesn't seem to for dictionary keys.
```
WillGibson, line 74: Is this comment still relevant? It looks like you've got it working downstairs.

tests/copilot_helper/utils/fixtures/addons_files/opensearch_addons_bad_data.yml
```diff
@@ -0,0 +1,62 @@
...
+  type: opensearch
+  environments:
+    dev:
+      volume_size: 512
+
+my-opensearch-11:
+  type: opensearch
+  environments:
+    dev:
+      deletion-policy: Snapeshot
```
WillGibson, line 59:
```suggestion
      deletion-policy: Snapshot
```

tests/copilot_helper/utils/test_validation.py
```diff
@@ -22,3 +35,203 @@ def test_validate_string(regex_pattern, valid_string, invalid_string):
...
+                "my-opensearch-2": r"deletion-policy.*does not match 27",
+                "my-opensearch-3": r"environments.*False should be instance of 'dict'",
+                "my-opensearch-4": r"environments.*Wrong key 'opensearch-plan'",
+                "my-opensearch-5": r"environments.*dev.*plan.*does not match 'largish'",
+                "my-opensearch-6": r"environments.*dev.*replicas.*should be an int between 0 and 5",
+                "my-opensearch-7": r"environments.*dev.*instance.*does not match 'medium'",
+                "my-opensearch-8": r"environments.*dev.*engine.*does not match 7.3",
+                "my-opensearch-9": r"environments.*dev.*volume_size.*should be an int between 10 and 511",
+                "my-opensearch-10": r"environments.*dev.*volume_size.*should be an int between 10 and 511",
+                "my-opensearch-11": r"environments.*dev.*deletion-policy.*does not match 'Snapeshot'",
```
WillGibson, line 154:
```suggestion
                "my-opensearch-11": r"environments.*dev.*deletion-policy.*does not match 'Snapshot'",
```

tests/copilot_helper/utils/fixtures/addons_files/s3_addons_policy.yml
```diff
@@ -0,0 +1,13 @@
+##### s3-policy example:
+
+my-s3-policy:
+  type: s3-policy  # gives the listed services permissions to access an existing bucket
+  readonly: true
+  deletion-policy: Retain
+  services:  # services that require access to the bucket.
+    - "web"
+    - - "other"
```
WillGibson, line 9:
```suggestion
    - "other"
```

WillGibson (commented):
Looking good. A few small comments.

dbt_copilot_helper/utils/validation.py
antroy-madetech: I think we should raise another ticket for this.

dbt_copilot_helper/utils/validation.py
antroy-madetech: Yes, it is still relevant. For values the "error" parameter works and outputs the custom text. For keys the custom text doesn't get reported in the exception for some reason.

tests/copilot_helper/utils/fixtures/addons_files/opensearch_addons_bad_data.yml
antroy-madetech: This is deliberate bad data for the test

tests/copilot_helper/utils/fixtures/addons_files/s3_addons_policy.yml
antroy-madetech: That's interesting. I'd have expected that to fail the test since it isn't a string.

copilot_helper.py
```diff
@@ -46,6 +46,4 @@ def copilot_helper():
 copilot_helper.add_command(waf_commands)
 
 if __name__ == "__main__":
-    # TODO: We should probably wrap this to catch a botocore.exceptions.UnauthorizedSSOTokenError to
```
gabelton, line 49: Does this not need to be done anymore?

dbt_copilot_helper/utils/validation.py
antroy-madetech: I'll keep the "s" of "should" lower case to be consistent with the built in errors.

dbt_copilot_helper/utils/validation.py
WillGibson:
```suggestion
    # For values the "error" parameter works and outputs the custom text. For keys the custom text doesn't get reported in the exception for some reason.
```

tests/copilot_helper/utils/test_validation.py
WillGibson: Or we could just use "BadData" so it's clear it's not just a typo?

dbt_copilot_helper/commands/copilot.py
```diff
@@ -100,10 +97,11 @@ def _normalise_keys(source: dict):
...
         return {}
 
-    try:
-        validate_json(instance=config, schema=schema)
-    except jsonschema.exceptions.ValidationError as err:
-        click.echo(click.style(err, fg="red"))
+    errors = validate_addons(config)
+    if errors:
+        click.echo(click.style(f"Errors found in {config_file}:", fg="red"))
+        for addon, error in errors.items():
```
gabelton, line 103: might be worth adding / updating a test for this new logic for outputting multiple errors

JohnStainsby (approved):

copilot_helper.py
antroy-madetech: There's a ticket for doing this in a slightly different way, leveraging Lawrence's code.

WillGibson (approved):

gabelton (approved):

PR #293, opened 2024-01-29
DBTP-448 - Remove addon type parameter from conduit command

JohnStainsby:
- Get addon type from config parameter rather than requiring it as a conduit command argument
- Swap to using CDK overrides for environments
- Add addons config to parameter store

New conduit command:
`copilot-helper conduit <addon-name> --app <app> --env <env>`


dbt_copilot_helper/commands/conduit.py
```diff
@@ -277,34 +276,52 @@ def start_conduit(app: str, env: str, addon_type: str, addon_name: str = None):
...
 @click.option("--addon-name", help="Name of custom addon", required=True)
-def conduit(addon_type: str, app: str, env: str, addon_name: str):
+def conduit(app: str, env: str, addon_name: str):
     """Create a conduit connection to an addon."""
     check_copilot_helper_version_needs_update()
 
-    try:
-        start_conduit(app, env, addon_type, addon_name)
-    except InvalidAddonTypeConduitError:
+    with open("addons.yml", "r") as fd:
```
acodeninja, line 286: I'm not sure about this, it now requires the command be run from the deploy repo...

Maybe an alternative is to use the parameter store / secrets manager paths as they include the addon name.

Alternatively, as per the pipeline setup [here](https://github.com/uktrade/copilot-tools/blob/main/dbt_copilot_helper/templates/pipelines/codebase/overrides/stack.ts#L447) you could upload the addons file to the environment and use that to [load data](https://github.com/uktrade/copilot-tools/blob/main/dbt_copilot_helper/commands/codebase.py#L101) for the conduit command.

The upside to this is that you could also add a `copilot conduit list --app <app> --env <env>` command or similar.

dbt_copilot_helper/commands/conduit.py
```diff
@@ -277,34 +276,52 @@ def start_conduit(app: str, env: str, addon_type: str, addon_name: str = None):
 
 
 @click.command(cls=ClickDocOptCommand)
-@click.argument("addon_type", type=click.Choice(CONDUIT_ADDON_TYPES))
 @click.option("--app", help="AWS application name", required=True)
 @click.option("--env", help="AWS environment name", required=True)
 @click.option("--addon-name", help="Name of custom addon", required=True)
```
acodeninja, line 281: Addon name as an option doesn't quite feel right. Maybe it will look nicer as an argument.

`copilot-helper conduit <addon-name> --app <app> --env <env>`

acodeninja (commented):
A few thoughts

dbt_copilot_helper/commands/conduit.py
JohnStainsby: Didn't really think about that command not being tied to the deploy repo 🫤

Getting the config from a parameter is a nicer solution anyway, I'll add a separate ticket for a `conduit list` command.

dbt_copilot_helper/templates/env/overrides/stack.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    private uploadAddonConfiguration() {
+        const deployRepoRoot = path.join(process.cwd(), '..', '..', '..');
+
+        const addonConfig = parse(readFileSync(
+            path.join(deployRepoRoot, 'addons.yml'),
+        ).toString('utf-8'));
+
+        new cdk.aws_ssm.CfnParameter(this, 'AddonConfig', {
+            name: `/copilot/applications/${this.appName}/addons`,
```
acodeninja, line 46: Something like this seems better, otherwise you'll run into race conditions when deploying/destroying environments.

```suggestion
            name: `/copilot/applications/${this.appName}/environments/${this.envName}/addons`,
```

dbt_copilot_helper/templates/env/overrides/.gitignore
```diff
@@ -0,0 +1,12 @@
...
+
+# NodeJS artifacts.
+node_modules
+*.js
+!jest.config.js
+*.d.ts
+
+# CDK asset staging directory.
+.cdk.staging
+cdk.out
```
acodeninja, line 12: EOF Newline

dbt_copilot_helper/templates/env/overrides/bin/override.ts
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env node
+import * as cdk from 'aws-cdk-lib';
+import { TransformedStack } from '../stack';
+
+const app = new cdk.App();
+new TransformedStack(app, 'Stack', {
+    appName: process.env.COPILOT_APPLICATION_NAME || "",
+    envName: process.env.COPILOT_ENVIRONMENT_NAME || "",
+});
```
acodeninja, line 9: EOF Newline

dbt_copilot_helper/templates/env/overrides/cdk.json
```diff
@@ -0,0 +1,20 @@
...
+      "**/*.d.ts",
+      "**/*.js",
+      "tsconfig.json",
+      "package*.json",
+      "yarn.lock",
+      "node_modules",
+      "test"
+    ]
+  }
+}
```
acodeninja, line 20: EOF Newline

acodeninja (commented):
Few bits to get started, will dig further on Monday.

dbt_copilot_helper/commands/conduit.py
```diff
@@ -40,25 +39,27 @@ class CreateTaskTimeoutConduitError(ConduitError):
...
+    "rds-postgres",
+    "aurora-postgres",
     "redis",
 ]
 
 
-def normalise_string(to_normalise: str) -> str:
-    output = re.sub("[^0-9a-zA-Z]+", "-", to_normalise)
-    return output.lower()
+def get_secret_name(addon_name: str) -> str:
```
WillGibson, line 48: I'm not sure about the rename, it's still normalising a string, not actually getting a secret name from anyway. Maybe this if you want it to be specific to secret names...
```suggestion
def normalise_secret_name(addon_name: str) -> str:
```

dbt_copilot_helper/utils/files.py
```diff
@@ -44,3 +44,31 @@ def ensure_cwd_is_repo_root():
             bg="red",
         )
         exit(1)
+
+
+def generate_files(base_path, file_path, output_dir):
```
WillGibson, line 49: Should we have a/some test(s) for this?

dbt_copilot_helper/utils/files.py
```diff
@@ -44,3 +44,31 @@ def ensure_cwd_is_repo_root():
             bg="red",
         )
         exit(1)
+
+
+def generate_override_files(base_path, file_path, output_dir):
+    for file in file_path.glob("*"):
```
WillGibson, line 50: These two bits of code are identical apart from the glob.

Do we need the two globs because we want all the files in the root of the directory and all the files from the `bin` directory?

We could perhaps extract the duplicated code to a helper method and just pass in the glob.

JohnStainsby: Yes, just the root and bin directories. I did have a go at refactoring this by trying to do it all in one glob pattern, but then it required excepting some other directories.

WillGibson (approved):

dbt_copilot_helper/commands/conduit.py
```diff
@@ -37,28 +37,68 @@ class CreateTaskTimeoutConduitError(ConduitError):
...
+
+    if addon_name not in addon_config.keys():
+        raise AddonNotFoundConduitError
+
+    for name, config in addon_config.items():
+        if name == addon_name:
+            addon_type = config["type"]
+
+    if not addon_type or addon_type not in CONDUIT_ADDON_TYPES:
+        raise InvalidAddonTypeConduitError(addon_type)
```
acodeninja, line 83: Maybe in a subsequent PR, but how about producing an error that says something like: "Addon dwPostgres not found, use one of the following: dwRedis, dwOpensearch".

dbt_copilot_helper/commands/conduit.py
```diff
@@ -277,19 +313,32 @@ def start_conduit(app: str, env: str, addon_type: str, addon_name: str = None):
...
+    except AddonNotFoundConduitError:
+        click.secho(
+            f"""Addon "{addon_name}" does not exist.""",
+            fg="red",
+        )
+        exit(1)
+    except InvalidAddonTypeConduitError as err:
         click.secho(
-            f"""Addon type "{addon_type}" does not exist, try one of {", ".join(CONDUIT_ADDON_TYPES)}.""",
+            f"""Addon type "{err.addon_type}" does not exist, try one of {", ".join(CONDUIT_ADDON_TYPES)}.""",
```
acodeninja, line 341: Maybe something more along the lines of `Addon type "{addon_type}" is not supported, we support: {", ".join(CONDUIT_ADDON_TYPES)}.`

acodeninja (approved):
Few small points on UX, otherwise great!

PR #299, opened 2024-01-30
Reverse the order in .python-version so latest is used by default

WillGibson:
After playing around with it locally, it seems that `pyenv` uses the first version in `.python-version` by default.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ line-length = 100
 
 [tool.poetry]
 name = "dbt-copilot-tools"
-version = "0.1.132"
+version = "0.1.133"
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -3,7 +3,7 @@ line-length = 100
 
 [tool.poetry]
 name = "dbt-copilot-tools"
-version = "0.1.132"
+version = "0.1.133"
```
WillGibson, line 6: So we can check it behaves in the pypi pipeline.

acodeninja (approved):
Nice use of `tac`, not seen that in a while!

WillGibson:
> Nice use of `tac`, not seen that in a while!

I didn't know it existed until yesterday ;-)

PR #300, opened 2024-01-30
DBTP-509 Use better error handling for copilot-helper

antroy-madetech:
Use the get_aws_session_or_abort function in the aws module in preference to the raw boto3.client call to get better errors if not logged in.

### DIFF
--- a/dbt_copilot_helper/commands/codebase.py
+++ b/dbt_copilot_helper/commands/codebase.py
@@ -4,7 +4,6 @@ import stat
 import subprocess
 from pathlib import Path
 
-import boto3
 import click
 import requests
 import yaml
@@ -12,6 +11,7 @@ import yaml
 from dbt_copilot_helper.utils.application import Application
 from dbt_copilot_helper.utils.application import ApplicationNotFoundError
 from dbt_copilot_helper.utils.application import load_application
+from dbt_copilot_helper.utils.aws import get_aws_session_or_abort
 from dbt_copilot_helper.utils.click import ClickDocOptGroup
 from dbt_copilot_helper.utils.files import mkfile
 from dbt_copilot_helper.utils.template import setup_templates
@@ -96,8 +96,9 @@ def prepare():
 def list(app, with_images):
     """List available codebases for the application."""
     application = load_application_or_abort(app)
-    ssm_client = boto3.client("ssm")
-    ecr_client = boto3.client("ecr")
+    session = get_aws_session_or_abort()
+    ssm_client = session.client("ssm")
+    ecr_client = session.client("ecr")
     parameters = ssm_client.get_parameters_by_path(
         Path=f"/copilot/applications/{application.name}/codebases",
         Recursive=True,
@@ -161,7 +162,8 @@ def build(app, codebase, commit):
         )
         raise click.Abort
 
-    codebuild_client = boto3.client("codebuild")
+    session = get_aws_session_or_abort()
+    codebuild_client = session.client("codebuild")
     build_url = start_build_with_confirmation(
         codebuild_client,
         f'You are about to build "{app}" for "{codebase}" with commit "{commit}". Do you want to continue?',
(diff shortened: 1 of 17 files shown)


pyproject.toml
```diff
@@ -3,7 +3,7 @@ line-length = 100
 
 [tool.poetry]
 name = "dbt-copilot-tools"
-version = "0.1.136"
+version = "0.1.138"
```
WillGibson: Why the double bump?

WillGibson (approved):
Looks legit :-)

dbt_copilot_helper/commands/codebase.py
```diff
@@ -96,8 +96,9 @@ def prepare():
 def list(app, with_images):
     """List available codebases for the application."""
     application = load_application_or_abort(app)
-    ssm_client = boto3.client("ssm")
-    ecr_client = boto3.client("ecr")
+    session = get_aws_session_or_abort()
```
acodeninja, line 99: This works, but also we're calling `get_aws_session_or_abort()` inside the `load_application_or_abort(app)` call. Not a major thing, but many calls...

dbt_copilot_helper/commands/codebase.py
```diff
@@ -189,11 +191,11 @@ def build(app, codebase, commit):
 def deploy(app, env, codebase, commit):
     """Trigger a CodePipeline pipeline based deployment."""
     application = load_application_with_environment(app, env)
-
     check_codebase_exists(application, codebase)
     check_image_exists(application, codebase, commit)
 
-    codebuild_client = boto3.client("codebuild")
+    session = get_aws_session_or_abort()
```
acodeninja, line 197: As we have the environment loaded, we can access the session for the environment via `application.environments.get(env).session`

dbt_copilot_helper/commands/copilot.py
```diff
@@ -423,7 +423,8 @@ def _cleanup_old_files(config, output_dir, env_addons_path):
 def get_env_secrets(app, env):
     """List secret names and values for an environment."""
 
-    client = boto3.client("ssm")
+    session = get_aws_session_or_abort()
```
acodeninja, line 426: Again here we could use `load_application_with_environment(app, env)` then `application.environments.get(env).session.client('ssm')`

acodeninja (commented):
Few bits on efficiency and using the `load_application_or_abort()` to get cached sessions in place.

pyproject.toml
antroy-madetech: I think it was when I merged in main. Probably manually bumped the version after the hook had already bumped.

dbt_copilot_helper/commands/copilot.py
antroy-madetech: I don't think there's any need to call load_application here as it would introduce some unnecessary calls to aws and make it more complicated to get the session back out.

tests/copilot_helper/test_command_codebase.py
```diff
@@ -134,18 +143,19 @@ def test_codebase_prepare_generates_an_executable_image_build_run_file(self, tmp
 
 
 class TestCodebaseBuild:
-    @patch("boto3.client")
     @patch("click.confirm")
     @patch("subprocess.run")
-    @patch("dbt_copilot_helper.utils.application.get_aws_session_or_abort", return_value=boto3)
+    @patch("dbt_copilot_helper.commands.codebase.get_aws_session_or_abort")
```
acodeninja, line 148: Its a little annoying that you've had to switch back to patching the command file rather than maintaining the patch on the dependency.

acodeninja (approved):
Approved, though the change to patching the import inside the file under test rather than patching the actual import we want to patch is a little annoying 😢

tests/copilot_helper/test_command_codebase.py
antroy-madetech: This is because the function is imported as `from dbt_copilot_helper.utils.aws import get_aws_session_or_abort`. See: https://docs.python.org/3/library/unittest.mock.html#where-to-patch for details. I think it was probably working incidentally before because of the additional "@mock_ssm" annotation.

PR #302, opened 2024-01-31
DBTP 630 domain http fix

ejayesh:
Fix to produce an error if no domain is specified in manifest.yml.  Plus addtional check that your sso token is set when running assign.


dbt_copilot_helper/utils/aws.py
```diff
@@ -52,6 +52,13 @@ def get_aws_session_or_abort(aws_profile: str = None) -> boto3.session.Session:
             fg="red",
         )
         exit()
+    except botocore.exceptions.TokenRetrievalError:
+        click.secho(
+            "The SSO Token associated with this profile has expired.  "
+            "To refresh this SSO session run aws sso login with the corresponding profile",
```
acodeninja, line 58: Just a bit to make it more obvious this is a command.
```suggestion
            "To refresh this SSO session run `aws sso login` with the corresponding profile",
```

dbt_copilot_helper/commands/dns.py
```diff
@@ -526,11 +526,24 @@ def get_load_balancer_domain_and_configuration(
...
+            if env in conf["environments"]:
+                for domain in conf["environments"].items():
+                    if domain[0] == env:
+                        if (
+                            domain[1] is None
+                            or domain[1]["http"] is None
+                            or domain[1]["http"]["alias"] is None
+                        ):
+                            click.secho(
+                                "No domains found, please check the manifest file", fg="red"
```
acodeninja, line 538: Check which manifest file? Might be useful to show the path here, just for better UX.

dbt_copilot_helper/commands/dns.py
```diff
@@ -526,11 +526,24 @@ def get_load_balancer_domain_and_configuration(
...
+                            )
+                            exit()
+                        domain_name = domain[1]["http"]["alias"]
+            else:
+                click.secho(
+                    f"Environment {env} not found, please check the manifest file", fg="red"
```
acodeninja, line 544: Again, check which manifest file?

acodeninja (commented):
Few little UX bits, otherwise nice!

dbt_copilot_helper/commands/dns.py
ejayesh: sure, will add path.

dbt_copilot_helper/commands/dns.py
ejayesh: adding full path

acodeninja (approved):
Noice

PR #305, opened 2024-02-05
Better bucket name validation

antroy-madetech:
Validation rules in here implemented: https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html


gabelton:
Are all the rules in that list implemented? I couldn't see:

- Bucket names must begin and end with a letter or number.
- Bucket names must not contain two adjacent periods.

Is it possible to validate the uniqueness of the name at this stage do you think?
- Bucket names must be unique across all AWS accounts in all the AWS Regions within a partition. A partition is a grouping of Regions. AWS currently has three partitions: aws (Standard Regions), aws-cn (China Regions), and aws-us-gov (AWS GovCloud (US)).
- A bucket name cannot be used by another AWS account in the same partition until the bucket is deleted.

acodeninja:
> Is it possible to validate the uniqueness of the name at this stage do you think?
> 

Yes, it is! From: https://platform.readme.trade.gov.uk/components/addons/#new-s3-bucket

Buckets must have [globally unique names](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html) and need to be defined for each environment. Check if a bucket exists by running:

```
aws s3api head-bucket --bucket <bucket>
```

A `404` response means the bucket name is available. `200`, `403` or `400` means the bucket already exists.

antroy-madetech:
The double-period rule is already validated. Missed the start/end with a letter/number. I'll also add the validatation that the name is free.

dbt_copilot_helper/utils/validation.py
```diff
@@ -18,10 +23,82 @@ def validator(string):
...
+)
+BUCKET_NAME_IN_USE_TEMPLATE = "Bucket name '{}' is already in use."
+
+
+def s3_bucket_name_is_available(name: str):
+    """
+    We try to find the bucket name in AWS.
+
+    The validation logic is:
+    True: if the response is a 200 (it exists and you have access - this bucket has probably already been deployed)
```
acodeninja, line 38: Will this work cross account? Thinking your current session is the non-prod account but you have deployed an S3 bucket into the prod account.

dbt_copilot_helper/utils/application.py
```diff
@@ -87,6 +87,7 @@ def load_application(app: str = None, default_session: Session = None) -> Applic
     application.environments = {
         env["name"]: Environment(env["name"], env["accountID"], sessions)
         for env in [json.loads(p["Value"]) for p in response["Parameters"]]
+        if "name" in env
```
acodeninja, line 90: Is this here to allow for the `/copilot/applications/{application}/environments/{environment}/addons` paths? Can we filter that out based on the parameter name in the line above, something along the lines of `[json.loads(p["Value"]) for p in response["Parameters"] if not p['Name'].endswith('addons')]`.

acodeninja (commented):
Maybe something to address on the cross account nature of S3 buckets.

dbt_copilot_helper/utils/validation.py
antroy-madetech: It won't, but it's still better to accept here, as there's no way to tell if it is the bucket for your deployment or for another account. It will have to fail on deployment in that case, which is better than stopping people rebuilding addons if they have already deployed their bucket once.

dbt_copilot_helper/utils/application.py
acodeninja: Yeah thats the addons config for the environment, if someone adds a new one with a name property that doesn't describe an environment this will break. How about the following instead:

```python
{
    env["name"]: Environment(env["name"], env["accountID"], sessions)
    for env in [
        json.loads(p["Value"]) for p in response["Parameters"] 
        if len(p.replace(f"/copilot/applications/{application.name}/environments/", "").split('/')) == 1
    ]
}
```

This will mean if we add extra parameters namespaced to environments that do have a name prop, we're not suddenly failing here.

dbt_copilot_helper/utils/validation.py
acodeninja: My point with this is that if the bucket already exists and is in the prod account, the call will get a 40x, therefore raising a validation error.

antroy-madetech: Oh, I see. I did have this check just throw up a warning rather than failing the check. I'll revert to that then I think. Unless you can see a better solution?

dbt_copilot_helper/utils/application.py
antroy-madetech: Do you mean check that the parameter name is in the correct format? i.e.:
```
if len(p['Name'].replace(f"/copilot/applications/{application.name}/environments/", "").split('/')) == 1
```

acodeninja: Yeah checking the parameter name rather than checking the parameter value has a name field.

dbt_copilot_helper/utils/validation.py
acodeninja: Pre-deployment I don't see a better way, post deployment the `application.environments` bit could be used to check each env account, but that just feels like overkill. I think just a warning saying that the bucket may be taken is enough, and indicate that it may be a problem, but check your aws accounts.

acodeninja (approved):

PR #307, opened 2024-02-06
Remove redundant X-Ray addon template file

yusufsheiqh:
## Context

- Remove X-Ray addon template file
  - This is redundant as we now using AWS Copilot's built-in observability tracing with AWS X-Ray.

## Additional information

Reference: https://aws.github.io/copilot-cli/docs/developing/observability/


samuele-mattiuzzo (approved):

WillGibson (approved):
It would be nice if all the distracting formatting changes were in a different commit and not distracting from the actual changes.

I wonder if we could automate that happening on our main branch on a schedule?

yusufsheiqh:
> It would be nice if all the distracting formatting changes were in a different commit and not distracting from the actual changes.
> 
> I wonder if we could automate that happening on our main branch on a schedule?

@WillGibson Good point. I didn't realise it would have formatted code that I didn't touch, which suggests the pre-commit hook was either bypassed before or didn't run against this file.

PR #309, opened 2024-02-08
test: DBTP-681 Pull request pipeline part 1

WillGibson:
This is part one, but it adds value and I have had to put this down for now because of other demands.

I would like to get it in now and myself or someone else can pick it up and go further in due course.


buildspec.test.yml
```diff
@@ -3,16 +3,18 @@ version: 0.2
 phases:
   install:
     commands:
-      - echo Setting local Python versions
+      - echo -e "\nSetting local Python versions"
```
WillGibson, line 6: Dropped some line break around to make it easier to see where the different parts begin and end in the nasty lofi CodeBuild output.

antroy-madetech (approved):
LGTM

PR #310, opened 2024-02-08
feat: task-level metrics dashboard [DBTP-698]

acodeninja:
- Adds a feature so bumping feature number in version
- Updates test suite to indicate which additional tools are required to use it
- Adds Task level metrics dashboard, example can be found in intranet-prod account [here](https://eu-west-2.console.aws.amazon.com/cloudwatch/home?region=eu-west-2#dashboards/dashboard/Test-2?start=PT24H)


yusufsheiqh (approved):
Looks good for a first iteration! Unfortunate that we cannot graph per task, but it is what it is.

ejayesh (approved):
nice work

PR #312, opened 2024-02-08
fix: all addon resource names must be globally unique [DBTP-734]

acodeninja:
- Fix so bumping patch version number
- All resources must be prefixed with `{{ addon_config.prefix }}` otherwise they will conflict
- Haven't added an overall test to check this due to time constraint, though maybe one could be implemented in future.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ line-length = 100
 
 [tool.poetry]
 name = "dbt-copilot-tools"
-version = "1.1.1"
+version = "1.1.2"
 description = "Set of tools to help transfer applications/services from GOV.UK PaaS to DBT PaaS augmenting AWS Copilot."
 authors = ["Department for Business and Trade Platform Team <sre-team@digital.trade.gov.uk>"]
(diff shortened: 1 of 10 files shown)


yusufsheiqh (approved):

WillGibson (approved):
Approving because I can't see it doing any harm, but apart from putting the prefix where it's name indicates it belongs, how does moving it from the end of the name to the beginning make it any more unique?

pyproject.toml
```diff
@@ -3,7 +3,7 @@ line-length = 100
 
 [tool.poetry]
 name = "dbt-copilot-tools"
-version = "1.1.1"
+version = "1.1.2"
```
WillGibson, line 6: Is there a chance that this could be backwards breaking?

acodeninja: Doubtful it would break anything, it will at most delete the existing log subscription filters and create them again.

To your other comment, I moved the prefixes to be prefixes because it was touching similar code. You know scouting rule and all that, leave things better than how you found them.

PR #320, opened 2024-02-13
DBTP-708 Remove all of the `codebuild` commands from copilot-helper

gabelton:
Proposing a minor change along the same lines as [this PR](https://github.com/uktrade/copilot-tools/pull/317)


pyproject.toml
```diff
@@ -3,7 +3,7 @@ line-length = 100
 
 [tool.poetry]
 name = "dbt-copilot-tools"
-version = "1.2.2"
+version = "1.3.0"
```
antroy-madetech, line 6: Strictly speaking this should be a major version bump as it removes functionality. I think given that it's not being used and doesn't break any current config the minor bump is probably fine.

gabelton: I thought it was maybe minor because the pipeline commands offer a workaround, so the functionality isn't completely removed, even if the command is

WillGibson: At this stage I would be comfortable with calling the removal of cruft we are confident no-one is using a bug fix.

yusufsheiqh: I think this is a minor version bump and that is because this command is not used, hence it's not really a breaking change. 😄

acodeninja: Agreed on the patch version segment for this one, we are removing something no one is using. Lets just make sure  we remove all references to the command from the platform docs.

gabelton: Docs change here https://github.com/uktrade/platform-documentation/pull/150

I guess it would be good to document a few common scenarios for major, minor, and patch as part of [this ticket]( https://uktrade.atlassian.net/browse/DBTP-744), given that we've each made the case for three different types of version bump here.

gabelton:
Think this ticket can be deleted now https://uktrade.atlassian.net/browse/DBTP-428

pyproject.toml
antroy-madetech: Agreed, because strictly speaking if it's removing functionality it should be a major version change. Even if it's removing stuff that's not used.

gabelton: Updated to patch

antroy-madetech: So the consensus in slack is to just go with strict semver rather than waste time discussing this sort of thing. So this should be a major version bump.

WillGibson (approved):

PR #322, opened 2024-02-14
DBTP-747 Make sure log resource policy is applied

WillGibson:
Fixes the bug in https://uktrade.atlassian.net/browse/DBTP-747

I don't think this is the end of the ticket, we need to think about testing, and I'll keep the ticket open for that, but this does fix an actual bug, so I would like to get it merged.

I have tested it by deploying a temporary environment with OpenSearch. I watched it fail, then ran `make-addons` from this code and watch it succeed.


tests/copilot_helper/fixtures/make_addons/expected/environments/overrides/cfn.patches.yml
WillGibson, line 1: This is no longer how it's done. I deleted the file, no tests failed.

WillGibson, line 1: This file is not relevant any more.

dbt_copilot_helper/templates/env/overrides/stack.ts
```diff
@@ -1,9 +1,9 @@
 import * as cdk from 'aws-cdk-lib';
+import { aws_cloudformation as cloudformation, aws_logs as logs, Fn } from 'aws-cdk-lib';
 import * as path from 'path';
-import {aws_logs as logs, aws_cloudformation as cloudformation} from 'aws-cdk-lib';
 import resource_policy from './log_resource_policy.json';
-import {readFileSync} from "fs";
-import {parse} from 'yaml';
+import { readFileSync } from "fs";
+import { parse } from 'yaml';
```
WillGibson, line 6: I let my IDE tidy up the imports.

dbt_copilot_helper/templates/env/overrides/stack.ts
```diff
@@ -32,7 +32,7 @@ export class TransformedStack extends cdk.Stack {
         const logResourcePolicy = this.template.getResource("LogResourcePolicy") as logs.CfnResourcePolicy;
 
         addonStack.addDependency(logResourcePolicy);
-        logResourcePolicy.policyDocument = JSON.stringify(resource_policy);
+        logResourcePolicy.policyDocument = Fn.sub(JSON.stringify(resource_policy));
```
WillGibson, line 35: This is the fix.

ejayesh (approved):
lgtm

PR #323, opened 2024-02-14
DBTP-729 Add S3 Retention policies

antroy-madetech:
This PR gives the ability to add retention policies to your S3 buckets.

It also removes the unnecessary top level deletion_policy parameter.


samuele-mattiuzzo (approved):

addons-example.yml
```diff
@@ -37,6 +36,10 @@ my-s3-bucket:
     production:
       bucket_name: my-bucket-prod
       deletion_policy: Retain
+      retention_policy: # Optional
+        mode: GOVERNANCE # or COMPLIANCE
```
WillGibson, line 40: What is the difference? Is that documented somewhere?

pyproject.toml
```diff
@@ -3,7 +3,7 @@ line-length = 100
 
 [tool.poetry]
 name = "dbt-copilot-tools"
-version = "2.0.2"
+version = "3.0.0"
```
WillGibson, line 6: 🦾

tests/copilot_helper/utils/fixtures/addons_files/opensearch_addons_bad_data.yml
```diff
@@ -3,52 +3,48 @@ my-opensearch-1:
...
 my-opensearch-2:
-  type: opensearch
-  deletion_policy: 27
-
-my-opensearch-3:
   type: opensearch
   environments: False
 
-my-opensearch-4:
+my-opensearch-3:
```
WillGibson, line 9: Could these have meaningful suffixes and remove the need to adjust them all is we add or remove one?

addons-example.yml
antroy-madetech: Governance means the bucket objects can be modified by an admin with the appropriate perms. Compliance means not even admins can remove or modify objects in the bucket.

antroy-madetech: FYI: https://aws.amazon.com/blogs/storage/protecting-data-with-amazon-s3-object-lock/

WillGibson: I think this might be helpful...
```suggestion
        mode: GOVERNANCE # or COMPLIANCE (See https://aws.amazon.com/blogs/storage/protecting-data-with-amazon-s3-object-lock/)
```

WillGibson (approved):
One small suggestion, but approving.

PR #324, opened 2024-02-14
DBTP-742 `copilot-helper domain configure` errors when prod env override does not contain http alias

samuele-mattiuzzo:


tests/copilot_helper/conftest.py
```diff
@@ -57,31 +57,13 @@ def fakefs(fs):
...
-
-  prod2:
-    http:
-      alias: v2.app.great.gov.uk
-
-  prod3:
-    http:
-      alias: app.trade.gov.uk
-""",
+def create_test_service_manifest(fakefs):
```
WillGibson, line 60: Pedantry...
```suggestion
def create_test_service_manifests(fakefs):
```

PR #326, opened 2024-02-15
DBTP-742 No `http.alias` in manifest bugfix

samuele-mattiuzzo:
[Jira ticket](https://uktrade.atlassian.net/browse/DBTP-742) 

I've decided to unfurl and extract the code that extracts the http domain alias to make sure it can be debugged.
It now logs when a manifest is misconfigured and checks for both http and alias' presence (being far too nice here...)

Caveat: somehow, when creating a failing test, the real error gets swallowed and execution resumes. Had to figure it out manually eventually.

Co-Authored by @WillGibson


dbt_copilot_helper/commands/dns.py
```diff
@@ -501,20 +512,21 @@ def _get_subdomains_from_env_manifests(environment, manifests):
...
-                    "  " + "\n  ".join(aliases),
-                    fg="yellow",
-                    bold=True,
-                )
-                subdomains.extend(aliases)
+                aliases = _get_domain_aliases(conf, environment)
+                if not len(aliases):
+                    click.echo(
+                        click.style(
+                            "Misconfigured: no domain or alias present, skipping...", fg="cyan"
```
WillGibson, line 519: It may not be a misconfiguration, e.g. https://github.com/uktrade/demodjango-deploy/blob/main/copilot/celery-worker/manifest.yml#L44

Maybe just...
```suggestion
                            f"No http.alias present for {environment} environment in {manifest}, skipping...", fg="cyan"
```

WillGibson (approved):

PR #327, opened 2024-02-16
DBTP-681 Add AWS Copilot CLI to debian/python image

WillGibson:
To support our pull request and regression test pipelines.

Has already been published and tested.

### DIFF
--- a/images/debian-python/Dockerfile.debian
+++ b/images/debian-python/Dockerfile.debian
@@ -32,8 +32,10 @@ RUN apt-get update \
   && apt-get clean autoclean \
   && apt-get autoremove -y \
   && rm -rf /var/lib/apt/lists/* \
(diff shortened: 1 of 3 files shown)


images/debian-python/Dockerfile.debian
```diff
@@ -32,12 +32,20 @@ RUN apt-get update \
   && apt-get clean autoclean \
   && apt-get autoremove -y \
   && rm -rf /var/lib/apt/lists/* \
-  && rm -f /var/cache/apt/archives/*.deb \
-  && git clone https://github.com/pyenv/pyenv .pyenv \
+  && rm -f /var/cache/apt/archives/*.deb
```
WillGibson, line 35: Split these as they are both long and it's a bit tedious having to run the whole thing again if you only change one side.

antroy-madetech (approved):
LGTM

PR #328, opened 2024-02-16
DBTP-623 - Default postgres conduit to read only

JohnStainsby:
- Default postgres conduit connections to read only permissions
- Add `write` flag to conduit command to use application user permissions
- Add `admin` flag to conduit command to use postgres user permissions
- Update postgres user lambda code to allow creation of multiple users
- Add read only user to postgres addons


tests/copilot_helper/test_command_conduit.py
```diff
@@ -116,52 +114,89 @@ def test_get_connection_secret_arn_when_secret_does_not_exist(mock_application):
...
-    get_connection_secret_arn, subprocess_call, mock_application
+@pytest.mark.parametrize(
+    "addon_type, addon_name",
+    [
+        ("postgres", "custom-name-postgres"),
+        ("postgres", "custom-name-rds-postgres"),
+        ("redis", "custom-name-redis"),
+        ("opensearch", "custom-name-opensearch"),
+    ],
+)
```
WillGibson, line 169: What's happening here with the two `paramerize`s?

Oh wow, today I learnt...

image

dbt_copilot_helper/commands/conduit.py
```diff
@@ -316,14 +354,16 @@ def start_conduit(application: Application, env: str, addon_type: str, addon_nam
 @click.argument("addon_name", type=str, required=True)
 @click.option("--app", help="AWS application name", required=True)
 @click.option("--env", help="AWS environment name", required=True)
-def conduit(addon_name: str, app: str, env: str):
+@click.option("--write", is_flag=True, help="Allow write access to database addons")
+@click.option("--admin", is_flag=True, help="Allow admin access to database addons")
```
WillGibson, line 358: At the moment `--write` overrides `--admin` if both are passed.

I feel like passing both is a mistake and it might be better to exit with an error.

JohnStainsby: Hmm admin should be taking precedence if both are provided but I agree having both doesn't feel right

tests/copilot_helper/test_command_conduit.py
JohnStainsby: You can stack multiple parametrize to get combinations of all of the arguments

dbt_copilot_helper/commands/conduit.py
antroy-madetech: How about if this was just "--access <read|write|admin>" rather than boolean flags?

dbt_copilot_helper/commands/conduit.py
```diff
@@ -88,14 +88,29 @@ def get_addon_type(app: Application, env: str, addon_name: str) -> str:
...
-def get_or_create_task_name(app: Application, env: str, addon_name: str) -> str:
+def get_parameter_name(
+    app: Application, env: str, addon_type: str, addon_name: str, write: bool, admin: bool
+) -> str:
+    if addon_type == "postgres":
+        permission = "READ_ONLY"
+        if admin:
+            permission = "ADMIN"
+        elif write:
+            permission = "WRITE"
```
antroy-madetech, line 99: If the cli option was "--access write" for example, this could be tidied up into a dictionary.

antroy-madetech: Or even just capitalized.

dbt_copilot_helper/commands/conduit.py
JohnStainsby: Yeah I'll have a look at that, maybe with read as the default so you don't always have to specific the access

WillGibson (approved):

PR #329, opened 2024-02-19
Run tests in parallel using xdist

markWarr:
These changes extend pytest to run tests in parallel, speeding up the unit test execution. 
Locally, I have seen the time to run tests reduce from 50s to 29s.
*UPDATE* since further updates have been made, this is now down to 25s.

The [pytest-xdist](https://pytest-xdist.readthedocs.io/en/stable/#pytest-xdist) plugin extends pytest with new test execution modes, the most used being distributing tests across multiple CPUs to speed up test execution.

In pytest.ini, the options `-n auto --dist loadgroup` are used . 
This tells xdist to use all available processors and to ensure that all tests in a group run in the same process.

Several tests perform file access and this is a limiting factor since we don't want several threads performing file operations at the same time. 
These tests have been marked with the following decorator:
`@pytest.mark.xdist_group(name="fileaccess")`

This tells xdist to run all tests marked with this decorator in a single process, ensuring that file access tests don't run together. Without this, results can be unpredictable and we can see the situation where a different test fails each time.


pyproject.toml
```diff
@@ -3,7 +3,7 @@ line-length = 100
 
 [tool.poetry]
 name = "dbt-copilot-tools"
-version = "2.0.2"
+version = "2.0.3"
```
yusufsheiqh, line 6: Not sure this change/refactor requires a version upgrade.

pyproject.toml
```diff
@@ -38,6 +38,7 @@ prettytable = "^3.9.0"
 semver = "^3.0.2"
 tomlkit = "^0.12.2"
 checkov = "^3.1.67"
+pytest-xdist = "^3.5.0"
```
yusufsheiqh, line 41: This should be moved into the `dev` dependencies section which is right below this one.

tests/copilot_helper/test_command_config.py
```diff
@@ -1,13 +1,16 @@
...
 from dbt_copilot_helper.commands.config import validate
 
 
+@pytest.mark.xdist_group(name="fileaccess")
 def test_running_in_non_copilot_directory():
     result = CliRunner().invoke(validate)
     assert result.output == "Could not find a deployment repository, no checks to run.\n"
 
 
+@pytest.mark.xdist_group(name="fileaccess")
```
WillGibson, line 13: @lawrence can maybe confirm, but I wouldn't expect to need this flag when we are using `fakefs`, I thought it was unique to each test.

pytest.ini
```diff
@@ -1,4 +1,4 @@
 [pytest]
 env =
     AWS_DEFAULT_REGION=eu-west-2
-addopts = --tb=short
+addopts = --tb=short -n auto --dist loadgroup
```
WillGibson, line 4: For readability...
```suggestion
addopts = --tb=short --numprocesses auto --dist loadgroup
```

pyproject.toml
markWarr: I think I had to do this to get past the pre-commit hooks. I just tried to reverse this and it bumped up the version to 3.0.1

WillGibson: `git commit --no-verify` for today.

Then we maybe need a ticket to not bump the version if the only thing that has changed is the test files.

tests/copilot_helper/test_command_config.py
lawrence: @WillGibson Not sure what's going on, but sure looks cool.

WillGibson: @acodeninja Am I correct in thinking that `fakefs` is unique to each test case, and therefore the tests can be safely run in parallel?

markWarr: I've removed the decorator from tests using fakefs and the tests are running fine.

WillGibson (approved):
So much faster. I'm sure there is more we can do, but 😍

PR #330, opened 2024-02-21
chore!: DBTP-756 delete waf config

ejayesh:
Remove all code related to WAF, as this is no longer needed in the helper command


acodeninja (commented):
Is this technically a breaking change?

ejayesh:
> Is this technically a breaking change?

I wouldn't have thought so, as we havent use this anywhere.   Although the waf.yml file is present in the DW-deploy repo.

WillGibson:
It is technically a breaking change, even if no-one is using it, and we agreed to be strict about applying it to prevent time spent talking about it.

Also, we need to decide on our release process!

ejayesh:
Ok.

ejayesh:
Link to the Release that will contain this change:  https://github.com/uktrade/copilot-tools/releases

WillGibson (approved):

PR #333, opened 2024-02-22
DBTP-770 Run unit tests in parallel by Python version in GitHub Actions

WillGibson:
Approximately 6 times faster than running them in serial in CodeBuild, which we can save for running the regression tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,6 +23,6 @@ jobs:
           poetry install
 
       - name: Test against Python ${PYTHON_VERSION}
-        run: poetry run tox -e "py${PYTHON_VERSION/./}"
+        run: poetry run pytest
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,28 @@
...
+          - "3.10"
+          - "3.9"
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: |
+          pip install poetry
+          poetry install
```
gabelton, line 23: Could you get away with just installing dev dependencies?

WillGibson: It needs the others too for the production code to run.

antroy-madetech (approved):

PR #336, opened 2024-02-23
fix: DBTP-763 - Get copilot and copilot-helper version from file

JohnStainsby:
Read the copilot and copilot helper versions from version files

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ line-length = 100
 
 [tool.poetry]
 name = "dbt-copilot-tools"
-version = "3.1.1"
+version = "3.1.3"
 description = "Set of tools to help transfer applications/services from GOV.UK PaaS to DBT PaaS augmenting AWS Copilot."
 authors = ["Department for Business and Trade Platform Team <sre-team@digital.trade.gov.uk>"]
 license = "MIT"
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -3,7 +3,7 @@ line-length = 100
 
 [tool.poetry]
 name = "dbt-copilot-tools"
-version = "3.1.2"
+version = "3.1.3"
```
gabelton, line 6:
```suggestion
version = "4.0.0"
```

gabelton (approved):
There are merge conflicts around the version but otherwise lgtm

acodeninja (commented):
Love this change, will help with some issues, quick thought though, does this need to go out with a change to stop the following type of error message when running `copilot-helper copilot make-addons` or equivilent?

```
You are running copilot-helper v3.1.2, upgrade to v4.0.0 by running run `pip install --upgrade dbt-copilot-tools`.
```

acodeninja:
@JohnStainsby 

> Yeah that change is going to have to be released with this. Just realised that means that pinning the `copilot-helper` version won't work for earlier versions than this.

I also have another thought... what about using the `.tool-versions` file we already use for other things? An example can be found [here for ICMS](https://github.com/uktrade/icms-deploy/blob/main/.tool-versions)

JohnStainsby:
@acodeninja 
> I also have another thought... what about using the `.tool-versions` file we already use for other things? An example can be found [here for ICMS](https://github.com/uktrade/icms-deploy/blob/main/.tool-versions)

Hmm, that would be cleaner in the deploy repos. The only thing is that the `.copilot-version` and `.copilot-helper-version` file convention has already made it's way into a few things

PR #337, opened 2024-02-23
fix!: DBTP-689 custom postgres application user blocks env delete

gabelton:
Implements first solution recommended by copilot team [here](https://github.com/aws/copilot-cli/issues/5665#issuecomment-1942076324), moving lambda from env addon to its own svc addon. More context on the ticket.

Draft release including upgrade path https://github.com/uktrade/platform-tools/releases/tag/untagged-cb99a891e658f57aff3e


tests/copilot_helper/fixtures/make_addons/expected/web/addons/my-aurora-db.yml
```diff
@@ -0,0 +1,265 @@
...
+    Description: The environment name your service, job, or workflow is being deployed to.
+  Name:
+    Type: String
+    Description: The name of the service, job, or workflow being deployed.
+
+Resources:
+  myAuroraDbLambdaFunction:
+    Type: AWS::Lambda::Function
+    Properties:
+      FunctionName: !Sub "${App}-${Env}-myAuroraDb-aurora-create-user"
```
acodeninja, line 18: Function names must be unique. What about a situation where there are multiple databases of the same type? Should be  namespaced with service name as well.

tests/copilot_helper/fixtures/make_addons/expected/web/addons/my-aurora-db.yml
```diff
@@ -0,0 +1,265 @@
...
+    UpdateReplacePolicy: Retain
+    Properties:
+      ServiceToken: !GetAtt 'myAuroraDbLambdaFunction.Arn'
+      CopilotApplication: !Sub "${App}"
+      CopilotEnvironment: !Sub "${Env}"
+      MasterUserSecret: 
+        Fn::ImportValue:
+          !Sub "${App}-${Env}-myAuroraDb-AuroraSecretName"
+      SecretDescription: !Sub  application user secret for ${AWS::StackName}
+      SecretName: !Sub '/copilot/${App}/${Env}/secrets/MY_AURORA_DB_APPLICATION_USER'
```
acodeninja, line 192: Should be service namespaced, if multiple services connect to the DB, thinking a worker service and a beat service, both will attempt to overwrite the same secret.

tests/copilot_helper/test_command_copilot.py
```diff
@@ -94,60 +94,68 @@ def test_validate_version_called_once(self, validate_version):
     @pytest.mark.parametrize(
         "addon_file, expected_env_addons, expected_service_addons, expect_db_warning",
         [
+            # (
```
acodeninja, line 97: This needs uncommenting...

dbt_copilot_helper/templates/addons/svc/postgres-lambda.yml
```diff
@@ -0,0 +1,265 @@
...
+      PublicRouteTable: 
+        Fn::ImportValue:
+          !Sub "${App}-${Env}-{{ addon_config.prefix }}-PublicRouteTableName"
+      PublicSubnet1RouteTableAssociation:
+        Fn::ImportValue: 
+          !Sub "${App}-${Env}-{{ addon_config.prefix }}-PublicSubnet1RouteTableAssociationName"
+      PublicSubnet2RouteTableAssociation: 
+        Fn::ImportValue: 
+          !Sub "${App}-${Env}-{{ addon_config.prefix }}-PublicSubnet2RouteTableAssociationName"
+      Snapshot: ""
```
acodeninja, line 211: Where has the username property gone?

acodeninja (changes requested):
I'm really not sure about this change working. There are two main issues:

- You aren't setting the required parameters `Username` and `Permissions` on the custom resource.
- You've left a block on addons commented out.

In addition to this I do also think this is a major version bump, but you've not included a zero downtime upgrade path. I'm also thinking there might be a few other fixes that should be merged in before forcing people to upgrade.

I also feel that now might be a good time to make this database custom resource testable. See the S3 Objects custom resource for an example: [Template code](https://github.com/uktrade/copilot-tools/blob/main/dbt_copilot_helper/templates/addons/env/s3.yml#L144-L157) | [Lambda code](https://github.com/uktrade/copilot-tools/blob/main/dbt_copilot_helper/custom_resources/s3_object.py) | [Tests](https://github.com/uktrade/copilot-tools/blob/main/tests/copilot_helper/custom_resources/test_s3_object.py)

gabelton:
> * You aren't setting the required parameters `Username` and `Permissions` on the custom resource.
> * You've left a block on addons commented out.

It was working when I tested it, but probs not any more because it needs John's changes from last week (which was when `Username` and `Permissions` properties were added. Will add those now. Just uncommented the test parameters and got them passing again

dbt_platform_helper/custom_resources/app_user.py
WillGibson, line 1: Is it possible to signpost anything which has actually changed in this?

dbt_platform_helper/custom_resources/app_user.py
```diff
@@ -0,0 +1,204 @@
...
+                Name=user_secret_name,
+                Description=user_secret_description,
+                Value=json.dumps(user_secret_string),
+                Overwrite=True,
+            )
+
+    return user_secret
+
+
+def send(event, body, logger, headers):
```
gabelton, line 71: This method is new, though closely modelled on logic in `dbt_platform_helper/custom_resources/s3_object.py`. Wasn't able to make it a shared util because it won't import in lambda env

dbt_platform_helper/custom_resources/app_user.py
```diff
@@ -0,0 +1,204 @@
...
+            break
+        except HTTPError as ex:
+            if count < 5:
+                logger.warning(f"{ex} [{ex.url}] - Retry {count}")
+            else:
+                logger.error(f"{ex} [{ex.url}]")
+            time.sleep(count * 5)
+
+
+# borrowed from https://github.com/awslabs/aws-cloudformation-templates/blob/master/aws/services/CloudFormation/MacrosExamples/StackMetrics/lambda/cfnresponse.py
```
gabelton, line 89: This method is new

dbt_platform_helper/custom_resources/app_user.py
```diff
@@ -0,0 +1,204 @@
...
+    user_secret_string = {
+        "username": username,
+        "password": user_password,
+        "engine": master_user["engine"],
+        "port": master_user["port"],
+        "dbname": master_user["dbname"],
+        "host": master_user["host"],
+    }
+
+    if "dbClusterIdentifier" in master_user.keys():
```
gabelton, line 150: These if blocks are new because the svc addon is shared between the two types of postgres addon

dbt_platform_helper/custom_resources/app_user.py
```diff
@@ -0,0 +1,204 @@
...
+                    "Data": {"Error": f"""Invalid requestType of '${event["RequestType"]}'"""},
+                }
+    except Exception as e:
+        response = {"Status": "FAILED", "Data": {"Error": str(e)}}
+
+    cursor.close()
+    conn.close()
+
+    print(json.dumps(response, default=str))
+    send_response(event, context, response["Status"], response["Data"], event["LogicalResourceId"])
```
gabelton, line 204: Using new method instead of cfnresponse module

gabelton (commented):
@WillGibson I've commented anything that's changed in the lambda code. There isn't too much. Everything else that github says is new (unit tests, templates, template map etc.) is actually new.

tests/platform_helper/fixtures/make_addons/expected/web/addons/my-rds-db.yml
```diff
@@ -0,0 +1,347 @@
...
+        SecurityGroupIds:
+          - !ImportValue
+            'Fn::Sub': "${App}-${Env}-myRdsDb-DBClusterSecurityGroupName"
+        SubnetIds: 
+          !Split 
+            - ","
+            - !ImportValue
+              'Fn::Sub': "${App}-${Env}-myRdsDb-PrivateSubnetsName" 
+      Code:
+        ZipFile: |
```
antroy-madetech, line 48: Feels like there must be a better way of doing this rather than copy-pasting the lambda function from the source file every time there's a change. Maybe either making this bit a regex wildcard, or replacing a placeholder with the actual file content or something.

antroy-madetech (commented):
Apart from the one comment, this looks good to me.

tests/platform_helper/fixtures/make_addons/expected/web/addons/my-rds-db.yml
WillGibson: Are you talking about the Terraform version, which will supersede this shortly?

gabelton: @antroy-madetech I don't mind the copy-pasting leg work because I want a unit test to fail when someone makes a change to the lambda code and to think what the change means while they're copying it over.

WillGibson: Going forward we will be using https://github.com/uktrade/terraform-platform-modules/blob/main/postgres/manage_users.py, which I am about to create a ticket to add unit tests for if we don't have them already.

gabelton:
closing. terraform solves and DW will be migrated soon, so not worth any more work

PR #338, opened 2024-02-27
docs: DBTP-735 Expand common command parameter placeholders to full words

WillGibson:
Addresses a couple of bullets from https://uktrade.atlassian.net/browse/DBTP-735


utils/create_command_docs.py
```diff
@@ -90,7 +90,16 @@ def get_cmd_metadata(
...
         subcommands=subcommands,
         # Strip "Usage: " (7 characters) and re-indent each line to match
-        usage="\n".join(map(lambda x: x[7:], cmd.get_usage(context).split("\n"))),
+        usage="\n".join(
+            map(
+                lambda x: x[7:],
+                cmd.get_usage(context)
+                .replace("<app>", "<application>")
+                .replace("<env>", "<environment>")
+                .replace("<svc>", "<service>")
```
WillGibson, line 99: If the list of replacements gets any longer, we may want to extract this out to a method, but that doesn't feel like a win to me as it stands today.

JohnStainsby (approved):

PR #340, opened 2024-02-28
Run black formatting on all files

WillGibson:
To keep some noise out of the pull request that coming next.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/psf/black
-    rev: 23.7.0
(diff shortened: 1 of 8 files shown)


.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/psf/black
-    rev: 23.7.0
+    rev: 24.2.0
```
WillGibson, line 3: Noticed we were a bit behind the curve

gabelton (approved):

yusufsheiqh (approved):

PR #342, opened 2024-02-28
fix: ensure opensearch domain is created before param

gabelton:
https://uktrade.atlassian.net/browse/DBTP-781

Slack discussion / context [here](https://ditdigitalteam.slack.com/archives/C046VC4HTKN/p1709126769784399)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ line-length = 100
 
 [tool.poetry]
 name = "dbt-copilot-tools"
-version = "3.1.2"
+version = "3.1.3"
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -3,7 +3,7 @@ line-length = 100
 
 [tool.poetry]
 name = "dbt-copilot-tools"
-version = "3.1.2"
+version = "3.1.3"
```
WillGibson, line 6: This should only change when we decide to do a release, which we probably should once this is merged.

WillGibson (approved):
We should probably do a release once this has been merged.

PR #344, opened 2024-02-29
chore: Remove -d / --directory arguments and default param from make_addons

samuele-mattiuzzo:
Small PR to suggest the removal of what doesn't seem to be a used param / argument for make_addons


JohnStainsby (approved):
I've never used this and can't see why we would need it

WillGibson (approved):

PR #352, opened 2024-03-05
fix: addon plan attribute inheritance

acodeninja:
Due to the alteration to config files where property names had to have `_` word separators, addons were no longer generated with the correct plan settings when their default plan included `-` separated properties.

This PR addresses this issue by including new tests to cover multiple environments when generating addons.

The specific instance of the problem was with the following config in `addons.yml`:

```yaml
database:
  type: rds-postgres
  version: 16.1
  environments:
    default:
      plan: small
    prod:
      plan: small-ha
```

Which generated the following output for the prod environment, 

```yaml
    prod:
      DBInstanceClass: 'db.t3.small'
      MultiAZ: False
```

When it should have produced the following:

```yaml
    prod:
      DBInstanceClass: 'db.t3.small'
      MultiAZ: True
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ line-length = 100
 
 [tool.poetry]
 name = "dbt-copilot-tools"
-version = "4.0.1"
+version = "4.1.0"
 description = "Set of tools to help transfer applications/services from GOV.UK PaaS to DBT PaaS augmenting AWS Copilot."
 authors = ["Department for Business and Trade Platform Team <sre-team@digital.trade.gov.uk>"]
 license = "MIT"


acodeninja:
> This will require a version bump btw.

I've bumped, I thought we were doing the version number only when we decide to do a release?

yusufsheiqh:
> I've bumped, I thought we were doing the version number only when we decide to do a release?

@acodeninja Not sure about the release process, so I will wait for a member of the Platform team to confirm. 😅

JohnStainsby (approved):
@acodeninja @yusufsheiqh 
We are agreeing the version number when we tag a release, based on the work it contains, so it doesn't really matter on each PR.

antroy-madetech (approved):
LGTM

PR #359, opened 2024-03-07
feat: add `iops` property for RDS addons

yusufsheiqh:
## Context

- Update `validations.py` to accommodate the `iops` property as optional ranging from a minimum of 1000 to 2000
  - This property has to be set if the storage type is set to `io1` which is set for production environments: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbinstance.html#cfn-rds-dbinstance-iops
- Update RDS addons template file
  - Change storage type to `io1` for `prod` and `production` named envrionments.
  - Add the `iops` property as part of the environment configuration map with a condition.
  - Use the `Iops` property within the RDS instance resource if the above condition is met otherwise it is set to `AWS:NoValue`.

### DIFF
--- a/dbt_copilot_helper/templates/addons/env/rds-postgres.yml
+++ b/dbt_copilot_helper/templates/addons/env/rds-postgres.yml
@@ -47,7 +47,7 @@ Mappings:
       AllocatedStorage: {{ config.volume_size }}
       MaxAllocatedStorage: {{ (config.volume_size * 1.26)|round|int }}
       StorageType: {% if env_name == "prod" or env_name == "production" %}io1{% else %}gp3{% endif %}
-      Iops: {{ config.iops if config.iops else '""' }}
+      Iops: {{ config.iops if env_name == "prod" or env_name == "production" else '""' }}
       DeletionPolicy: {{ config.deletion_policy }}
       DeletionProtection: {{ config.deletion_protection if config.deletion_protection else false }}
       SnapshotIdentifier: {{ config.snapshot_id if config.snapshot_id else '""' }}
--- a/dbt_copilot_helper/utils/validation.py
+++ b/dbt_copilot_helper/utils/validation.py
@@ -376,7 +376,7 @@ RDS_SCHEMA = Schema(
             ENV_NAME: {
                 Optional("plan"): RDS_PLANS,
                 Optional("volume_size"): int_between(20, 10000),
-                Optional("iops"): int_between(1000, 2000),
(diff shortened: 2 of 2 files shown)


dbt_copilot_helper/utils/validation.py
```diff
@@ -376,6 +376,7 @@ def is_between(value):
             ENV_NAME: {
                 Optional("plan"): RDS_PLANS,
                 Optional("volume_size"): int_between(20, 10000),
+                Optional("iops"): int_between(1000, 2000),
```
JohnStainsby, line 379: Unfortunately I think the validation needs to be a bit more complex than this, there are some extra constraints listed under the [iops property here.](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbinstance.html)

Constraints:
- For RDS for Db2, MariaDB, MySQL, Oracle, and PostgreSQL - Must be a multiple between .5 and 50 of the storage amount for the DB instance.

dbt_copilot_helper/templates/addons/env/rds-postgres.yml
```diff
@@ -46,7 +46,8 @@ Mappings:
       MultiAZ: {{ config.multi_az }}
       AllocatedStorage: {{ config.volume_size }}
       MaxAllocatedStorage: {{ (config.volume_size * 1.26)|round|int }}
-      StorageType: {% if env_name == "prod" %}io1{% else %}gp3{% endif %}
+      StorageType: {% if env_name == "prod" or env_name == "production" %}io1{% else %}gp3{% endif %}
+      Iops: {{ config.iops if config.iops else '""' }}
```
JohnStainsby, line 50: Could we set a default value for iops if the env is prod, or update the validation so enforce the iops property for prod.

JohnStainsby (commented):
Couple of comments, I know this is getting in the way of some migration activities so maybe these can be addressed in a separate ticket?

antroy-madetech (commented):
A comment to look at and address, but other than that looks good.

dbt_copilot_helper/utils/validation.py
PaulWheatcroft: I'm just picking this up in Yusuf's absence. As @antroy-madetech suggests 9950 is still well below the upper limit boundary so am I OK to amend to this or also as suggested just remove the upper limit?

antroy-madetech (approved):

JohnStainsby (approved):

PR #368, opened 2024-03-12
chore: Ensure python versions are installed and used when running tests

samuele-mattiuzzo:
Me and @JohnStainsby noticed that currently all 4 test runs are run with python version `3.10.12` so effectively, 3 are redundant.

This PR addresses this.

### DIFF
--- a/.github/workflows/quality-checks.yml
+++ b/.github/workflows/quality-checks.yml
@@ -9,14 +9,18 @@ jobs:
 
     strategy:
       matrix:
-        python-version:
-          - "3.12"
-          - "3.11"
-          - "3.10"
-          - "3.9"
+        python-version: ["3.12", "3.11", "3.10", "3.9"]
 
     steps:
       - uses: actions/checkout@v4
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Display Python version
+        run: python -c "import sys; print(sys.version)"
 
       - name: Install dependencies
         run: |
--- a/.github/workflows/quality-checks.yml
+++ b/.github/workflows/quality-checks.yml
@@ -20,7 +20,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Display Python version
-        run: python -c "import sys; print(sys.version)"
+        run: python --version
 
       - name: Install dependencies
         run: |
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,5 +1,5 @@
 [pytest]
 env =
     AWS_DEFAULT_REGION=eu-west-2
-;addopts = --tb=short --numprocesses auto --dist loadgroup
-addopts = --tb=short
+addopts = --tb=short --numprocesses auto --dist loadgroup
+;addopts = --tb=short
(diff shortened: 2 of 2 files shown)


JohnStainsby (approved):
The regression tests are failing, but I think Will is fixing it in his PR

pytest.ini
```diff
@@ -1,5 +1,5 @@
 [pytest]
 env =
     AWS_DEFAULT_REGION=eu-west-2
-;addopts = --tb=short --numprocesses auto --dist loadgroup
-addopts = --tb=short
+addopts = --tb=short --numprocesses auto --dist loadgroup
+;addopts = --tb=short
```
WillGibson, line 5:
```suggestion
addopts = --tb=short --numprocesses auto --dist loadgroup
; You may need to use the addopts setting below to unbreak debugging in your IDE
; addopts = --tb=short
```

.github/workflows/quality-checks.yml
```diff
@@ -9,15 +9,19 @@ jobs:
...
     steps:
       - uses: actions/checkout@v4
 
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Display Python version
+        run: python -c "import sys; print(sys.version)"
```
WillGibson, line 23: How does this install the correct version?

For my old mind, can we have this too...
```suggestion
        run: python -c "import sys; print(sys.version)"
```
```suggestion
        run: |
          python -c "import sys; print(sys.version)"
          python --version
```

.github/workflows/quality-checks.yml
```diff
@@ -9,15 +9,19 @@ jobs:
...
-          - "3.9"
+        python-version: ["3.12", "3.11", "3.10", "3.9"]
 
     steps:
       - uses: actions/checkout@v4
 
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
```
samuele-mattiuzzo, line 20: @WillGibson install wise, this is where it happens ☝🏻

.github/workflows/quality-checks.yml
samuele-mattiuzzo: I'm happy to give it a shot, however the changes I've made are as per GitHub's official documentation which I have linked in the issue's description

samuele-mattiuzzo: Either way, we need one only between `--version` and `-c "import..."` as they'd do the same thing. Considering using the `-c` flag is Git's official recommendation, I would maybe stick with it, or get rid of lines 22-23 alltogether as it's just a verbose way to display the correct python version in use (which is going to show up muddled somewhere atop each pytest run)

WillGibson: It may be unfounded, but I have more confidence that `python --version` will display the version that will be used in the following steps.

I do prefer to explicitly have this output.

.github/workflows/quality-checks.yml
WillGibson: I am very slow sometimes 😂

WillGibson (approved):

PR #369, opened 2024-03-13
feat: DBTP-786 file based platform-tools version checks

samuele-mattiuzzo:
Replaces https://github.com/uktrade/platform-tools/pull/355 to clean up after a lot of wip commits and the rename to `platform-tools`

Jira ticket: https://uktrade.atlassian.net/browse/DBTP-786

Differences:

- part of required a global flag of sorts to ensure a call to `generate()` wouldn't print the same version-mismatched message 2 more times (after `pipeline.generate` and `make_addons` get called); `Click` doesn't allow that and using python's `globals()` causes GitHub Actions to fail intermittently, probably due to aggressive caching. Spoke with @JohnStainsby briefly and looked into it, and we decided to discard it as it's a minor feature and pipeline_generate / make_addons should not be used anymore as a standalone command

- part of the quality checks has not been implemented as I've realised it's not 100% clear to me where and how I should implement it so I will need to chat to @WillGibson further about it, but I don't want to hold this any longer as it's been a while since I've picked it up


dbt_platform_helper/utils/versioning.py
```diff
@@ -103,6 +108,17 @@ def validate_version_compatibility(
...
 
+def check_version_on_file_compatibility(
+    app_version: Tuple[int, int, int], file_version: Tuple[int, int, int]
+):
+    app_major, app_minor, app_patch = app_version
+    file_major, file_minor, file_patch = file_version
+
+    if app_major != file_major or app_minor != file_minor or app_patch != file_patch:
+        return False
+    return True
```
WillGibson, line 119: Maybe...
```suggestion
    return app_major == file_major and app_minor == file_minor and app_patch == file_patch
```

tests/platform_helper/utils/test_versioning.py
```diff
@@ -117,6 +119,28 @@ def test_validate_template_version(template_check: Tuple[str, Type[BaseException
...
+@patch("dbt_platform_helper.utils.versioning.get_file_app_versions")
+def test_validate_platform_helper_file_version(
+    get_file_app_versions, template_check: Tuple[str, Type[BaseException], str]
+):
+    get_file_app_versions.return_value = (1, 0, 0), (1, 0, 0)
+    template_name, raises, message = template_check
+
+    with pytest.raises(raises) as exception:
+        template_path = str(Path(f"{FIXTURES_DIR}/version_validation/{template_name}").resolve())
+        validate_platform_helper_file_version(template_path)
```
WillGibson, line 138: I think setting `template_path` is part of the "arrange".
```suggestion
    template_name, raises, message = template_check
    template_path = str(Path(f"{FIXTURES_DIR}/version_validation/{template_name}").resolve())

    with pytest.raises(raises) as exception:
        validate_platform_helper_file_version(template_path)
```

WillGibson (approved):

PR #373, opened 2024-03-13
chore: update nginx to use new image

acodeninja:
Nginx moved from public.ecr.aws/uktrade/nginx-reverse-proxy to public.ecr.aws/uktrade/nginx-dbt-platform but we never updated it in the templates.


acodeninja:
Needs https://github.com/uktrade/platform-tools/pull/366 to merge first.

WillGibson (approved):

antroy-madetech (approved):

PR #380, opened 2024-03-27
fix: update aws-cdk-lib package to fix semver package vulnerability

tony-griffin:


gabelton:
I don't this change works with copilot, unless something's changed since https://github.com/uktrade/platform-tools/pull/244

PR #383, opened 2024-04-03
feat: DBTP-778 Slack notification on publish

samuele-mattiuzzo:


utils/notify/publish_notification.py
```diff
@@ -0,0 +1,53 @@
+import os
+
+from slack_sdk import WebClient
+from slack_sdk.models import blocks
+
+
+class Settings:
+    channel: str
+    workspace: str
```
gabelton, line 9: Does this get used?

utils/notify/publish_notification.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class PublishNotify:
+    settings: Settings
+
+    def __init__(self, send_notifications: bool = True):
+        self.settings = Settings()
+        self.send_notifications = send_notifications
+        self.release_notes_url = "https://github.com/uktrade/platform-tools/releases/latest"
```
gabelton, line 18: Should this be a constant at the top of the file?

utils/notify/publish_notification.py
samuele-mattiuzzo: it's used a bit further down to set the values for each when instantiating the notifier object; I've kept it just to stay aligned with `ci-image-builder` in case we bring it into `platform-tools`

otherwise no, not strictly needed

utils/notify/publish_notification.py
samuele-mattiuzzo: yes, fair point; separated it just because it's a long url, but we can move into a const

utils/notify/publish_notification.py
gabelton: Perhaps we could remove for now? Up to you. Not a blocker

samuele-mattiuzzo: @gabelton sorted, fancy giving it a quick re-review?

PR #395, opened 2024-04-08
DBTP-777 - Automate release version

JohnStainsby:


WillGibson (commented):
I thought the plan was to automatically bump the version in the code when we publish a GitHub release.

I don't think we would want this to create a release on every merge to main.

JohnStainsby:
@WillGibson 
> I thought the plan was to automatically bump the version in the code when we publish a GitHub release.
> 
> I don't think we would want this to create a release on every merge to main.

It won't. On merge to main this will bump the code version and generate release note.
A release will only be created when the automatically generated release-pr is merged.

Example release-pr:
Screenshot 2024-04-09 at 16 53 08

samuele-mattiuzzo:
> * Uses [release-please](https://github.com/marketplace/actions/release-please-action) GitHub action to create and update a release PR when changes are merged to main

Is this for all PRs going from `feature-branch` to `main`? What's the behaviour like when we do have test/transient pieces of work that require us to merge multiple times into master and trigger Codebuild (as an example)? Can we override and bump back down if we need to?

JohnStainsby:
@samuele-mattiuzzo 
> Is this for all PRs going from `feature-branch` to `main`? What's the behaviour like when we do have test/transient pieces of work that require us to merge multiple times into master and trigger Codebuild (as an example)? Can we override and bump back down if we need to?

There are config options to manually set the next release version. Or you could just close the suggested release-pr and manually create a GitHub release

WillGibson (approved):

samuele-mattiuzzo (approved):

PR #404, opened 2024-04-17
test: fix pytest warnings

gabelton:
https://uktrade.atlassian.net/browse/DBTP-906

Bumping moto silences CryptographyDeprecationWarning - requires replacing decorators with all-purpose `mock_aws`

Running tests with 3.12 gives hundreds of warnings - silenced a few of these in `pytest.ini`

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -147,6 +147,17 @@ files = [
 ]
 
 [[package]]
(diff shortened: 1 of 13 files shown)


yusufsheiqh (approved):
I like the general purpose `mock_aws` decorator. 🚀

PR #410, opened 2024-04-23
chore!: DBTP-910 - Remove environment overrides

JohnStainsby:
Remove env overrides for log resource policies which have now moved to Terraform


dbt_platform_helper/templates/env/overrides/stack.ts
```diff
@@ -23,18 +21,9 @@ export class TransformedStack extends cdk.Stack {
         this.appName = props.appName;
         this.envName = props.envName;
 
-        this.addLogResourcePolicy();
```
WillGibson, line 26: I think to support existing 100% AWS Copilot applications,  we need to keep this, but make it conditional on whether there is a `terraform` directory?

lgarvey:
This is handled by this PR: 

https://github.com/uktrade/platform-tools/commit/8f6af32532e112a0bd550884acee98bc0ca1f654

PR #412, opened 2024-04-24
feat: DBTP-860 Add a command to generate/update environment manifest files with VPC/ALB config

gabelton:
I've added the `ruamel.yaml` dependency because `pyyaml` doesn't preserve comments and spacing when loading. 

I still need to create a docs PR to indicate where in the process we expect this command to be run.


samuele-mattiuzzo (approved):

dbt_platform_helper/commands/copilot.py
```diff
@@ -467,3 +468,57 @@ def _rmdir(path):
         for f in svc_addons_dir.iterdir():
             if f.is_file():
                 f.unlink()
+
+
+@copilot.command
+def vpc_generate():
```
lgarvey, line 474: Is this being added to the general `platform-helper generate` command?

dbt_platform_helper/commands/copilot.py
```diff
@@ -467,3 +468,57 @@ def _rmdir(path):
...
+            ]
+            new_data["network"] = {
+                "vpc": {"id": vpc["VpcId"], "subnets": {"public": public, "private": private}}
+            }
+
+            with open(env_manifest_path, "w") as manifest_file:
+                yuamel.dump(new_data, manifest_file)
+        else:
+            click.echo(
+                f"{env_name} environment manifest file not found. You may need to run `copilot env init --name {env_name}` to generate this file."
```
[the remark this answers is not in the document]
WillGibson: I am in favour of generating the configuration fresh each time we need to use it, no it needs to contain details which can be changed outside of AWS Copilot land.

Maybe we need to talk it through with @rossmiller and @benn-mt and actually make a decision.

dbt_platform_helper/commands/copilot.py
gabelton: I think it's related to the question of whether we generate env manifest files as part of this command. If we roll it into this `generate` command, then we need to either run `copilot env init` as part of `generate` or document that `copilot env init` will need to be run before `generate`

WillGibson: Maybe we can move this command into an environment space...

`platform-helper environment update-vpc-config <environment>`

I agree that it should not be dong `copilot env init`. In fact, we will not need to run that to generate the environment manifests.

I would expect this command to be fired later from `platform-helper environment generate <environment>`.

dbt_platform_helper/commands/copilot.py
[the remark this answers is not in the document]
WillGibson: I'm definitely seeing an `environments.yml` in our future.

dbt_platform_helper/commands/environment.py
```diff
@@ -134,6 +138,60 @@ def online(app, env):
...
+def update_vpc_config():
+    """Update or add VPC configuration in copilot environment manifest files."""
+
+    ensure_cwd_is_repo_root()
+    session = get_aws_session_or_abort()
+    envs = list_copilot_local_environments()
+    vpcs = session.client("ec2").describe_vpcs()["Vpcs"]
+
+    for vpc in vpcs:
+        # skip unnamed vpcs
```
WillGibson, line 151: Pedantry...
```suggestion
        # skip unnamed VPCs
```

dbt_platform_helper/commands/environment.py
```diff
@@ -134,6 +138,60 @@ def online(app, env):
         raise click.Abort
 
 
+@environment.command()
+def update_vpc_config():
+    """Update or add VPC configuration in copilot environment manifest files."""
```
WillGibson, line 143: Part 1 of little refactor suggestion
```suggestion
    """Update or add VPC configuration in copilot environment manifest files."""

    def filter_subnets(subnets, visibility):
        tag = {"Key": "subnet_type", "Value": visibility}
        return [
            {"id": subnet["SubnetId"]} for subnet in subnets if tag in subnet["Tags"]
        ]
```

dbt_platform_helper/commands/environment.py
```diff
@@ -134,6 +138,60 @@ def online(app, env):
...
+            public = [
+                {"id": subnet["SubnetId"]} for subnet in subnets if public_tag in subnet["Tags"]
+            ]
+            private_tag = {"Key": "subnet_type", "Value": "private"}
+            private = [
+                {"id": subnet["SubnetId"]} for subnet in subnets if private_tag in subnet["Tags"]
+            ]
+            new_data["network"] = {
+                "vpc": {"id": vpc["VpcId"], "subnets": {"public": public, "private": private}}
+            }
```
WillGibson, line 185: Part 2 of little refactor suggestion
```suggestion
            
            new_data["network"] = {
                "vpc": {
                    "id": vpc["VpcId"],
                    "subnets": {
                        "public": (filter_subnets(subnets, "public")),
                        "private": (filter_subnets(subnets, "private")),
                    },
                }
            }
```

tests/platform_helper/test_command_environment.py
```diff
@@ -309,6 +315,104 @@ def test_online_an_environment_when_load_balancer_not_found(
         remove_maintenance_page.assert_not_called()
 
 
+class TestUpdateVpcConfig:
+    @mock_aws
+    def test_vpc_generate(self, fakefs):
```
WillGibson, line 320:
```suggestion
    def test_update_vpc_config(self, fakefs):
```

tests/platform_helper/test_command_environment.py
```diff
@@ -309,6 +315,104 @@ def test_online_an_environment_when_load_balancer_not_found(
...
+        )["Vpc"]
+        addons_dir = FIXTURES_DIR / "make_addons"
+        fakefs.add_real_directory(
+            addons_dir / "config/copilot", read_only=False, target_path="copilot"
+        )
+
+        result = CliRunner().invoke(update_vpc_config)
+
+        assert (
+            "envwithnomanifestfile environment manifest file not found. You may need to run `copilot env init --name envwithnomanifestfile` to generate this file."
```
WillGibson, line 411:
```suggestion
            "envwithnomanifestfile environment manifest file not found. You may need to run `copilot env init --name envwithnomanifestfile` to update the VPC details."
```

Also, `environment_with_no_manifest` might be easier to read.

gabelton:
I've moved all the logic that was in the `copilot` command file and grouped it under one `environment` generate command, creating the file from scratch instead of updating the existing. I've stopped short of using an `environments.yml` because I felt that this needed some discussion and maybe an ADR. So at this point the command expects one or more env names to be passed as arguments.

ejayesh:
> I've moved all the logic that was in the `copilot` command file and grouped it under one `environment` generate command, creating the file from scratch instead of updating the existing. I've stopped short of using an `environments.yml` because I felt that this needed some discussion and maybe an ADR. So at this point the command expects one or more env names to be passed as arguments.

I think we should include a list of envs somewhere though, maybe it could go in the extensions.yml file as a list.  But we can stick with it being passed in for now.  But this looks good to me

ejayesh (approved):
approved, changes lookg good to me

PR #426, opened 2024-05-07
fix: env generate vpc name not found bug

gabelton:
Non-prod envs all use the same vpc and expect the vpc name to match AWS account name e.g. `intranet`, whereas prod env expects format `intranet-prod`

Also removes commented out crap from previous PR


ejayesh (approved):
lovely approved

PR #428, opened 2024-05-10
feat: DBTP-933 Add option to Platform-helper to manage .aws/config file

gabelton:
I would've added more unit tests for the helper methods, but `moto` doesn't support any of the methods used and the tests seemed fairly trivial by the time I'd patched that moto logic.


tony-griffin (approved):
Looks good. Wondering how you got around the sessions permissions. I couldn't seem to get access to the correct session to be allowed to call create_token()

samuele-mattiuzzo (approved):

PR #429, opened 2024-05-15
test: DBTP-990 Use platform-helper generate in regression tests

WillGibson:


WillGibson:
Doing it under https://uktrade.atlassian.net/browse/DBTP-989 now.

PR #430, opened 2024-05-16
feat: Validation usability improvements

antroy-madetech:
Validation catches all errors now rather than it being a manual process to iterate over failures
Undeprecated the make-addons and pipeline generate commands, as they are used in different places in the new process.


dbt_platform_helper/commands/pipeline.py
```diff
@@ -28,13 +28,9 @@ def pipeline():
     check_platform_helper_version_needs_update()
 
 
-@pipeline.command(deprecated=True, hidden=True)
+@pipeline.command()
```
gabelton, line 31: What's the scenario where this gets run by itself?

gabelton:
Is it possible to update / add a unit test to capture the change to validation capturing more than one error?

antroy-madetech:
> Is it possible to update / add a unit test to capture the change to validation capturing more than one error?

Thought I had, but it was just multiple bad environments I'd added. Adding a separate test now.

WillGibson (approved):
Lovely

dbt_platform_helper/commands/pipeline.py
gabelton: makes sense. thanks!

gabelton (approved):

PR #432, opened 2024-05-17
feat: DBTP-969 Allow HTTPS GitHub clones

benn-dbt:
Includes creating unit tests for git.py
But git.py now supports both SSL and HTTPS github URLs.


dbt_platform_helper/utils/git.py
```diff
@@ -6,10 +6,13 @@ def git_remote():
     git_repo = subprocess.run(
         ["git", "remote", "get-url", "origin"], capture_output=True, text=True
     ).stdout.strip()
+    return extract_repository_name(git_repo)
```
WillGibson, line 9: Is there value in adding a test to verify that `git_remote()` calls `extract_repository_name()`?

Or does some other test fail in a way that makes the problem obvious if it's delated?

dbt_platform_helper/utils/git.py
```diff
@@ -6,10 +6,13 @@ def git_remote():
...
 
-    if not git_repo:
+
+def extract_repository_name(repository_url):
+    if not repository_url:
         return
 
-    _, repo = git_repo.split("@")[1].split(":")
+    repo = re.search(r"[^/:]*?/[^/]*?\.git\Z", repository_url).group()
```
WillGibson, line 16: Maybe this could be more readable if we did it one way for repository URLs starting with "https" and the previous way for the rest?

dbt_platform_helper/utils/git.py
benn-dbt: I couldn't work out what I needed to do for "@patch" to make that work. 
What I'd *love* to do is look at where this is called and see if we can move the subprocess.run bit out of this entirely. Or even fetch the github url from somewhere else entirely?

dbt_platform_helper/utils/git.py
benn-dbt: Only difference between the regexes would be in the [^/:] section. The rest of the regex would be the same.

WillGibson: It's still not pretty, but is this easier to grok than the regex?
```suggestion
    if repository_url.startswith("https"):
        repo = repository_url.split("//")[1].split("/", maxsplit=1)[1]
    else:
        repo = repository_url.split(":")[1]
```

benn-dbt: I've refactored the entire method to a single line, using regex. Would be interested in thoughts on the update.

WillGibson: Looking now. I was thinking we could use a capturing group to ditch that extra step in the return line, but was waiting to see whether we settled on Regex or not.

dbt_platform_helper/utils/git.py
```diff
@@ -6,10 +6,11 @@ def git_remote():
...
-    if not git_repo:
-        return
 
-    _, repo = git_repo.split("@")[1].split(":")
+def extract_repository_name(repository_url):
+    if not repository_url:
+        return
 
-    return re.sub(r".git$", "", repo)
+    return re.search(r"([^/:]+?/[^/]+?)\.git", repository_url).group(1)
```
WillGibson, line 16: I am OK with using a regular expression here, but because they are hard to understand, to help the next person and to validate if I have managed to understand it...
```suggestion
    # In the capturing group (the bit in the brackets)...
    # [^/:]+? Capture the organisation part (matches any character except : or / and captures as few characters as possible)
    # / Capture the slash after the organisation
    # [^/]+? Capture the repository name
    # After the capturing group...
    # Match, but don't capture the .git part at the end
    return re.search(r"([^/:]+?/[^/]+?)\.git", repository_url).group(1)
```

benn-dbt: Decided after discussion not to include the comment - would become out of date with the regex, makes the method significantly longer, and there are tools to help understand the regex (which we've made simpler)

WillGibson (approved):

PR #438, opened 2024-05-21
fix: platform-helper generate ALB bugfixes

samuele-mattiuzzo:
https://uktrade.atlassian.net/browse/DBTP-945

- [x] validation does not allow for `additional_address_list` under `ALB` configuration
- [x] validation does not allow for an empty `prod` environment (without configuration)


dbt_platform_helper/commands/copilot.py
```diff
@@ -168,6 +168,9 @@ def _normalise_keys(source: dict):
             normalised_environments[env] = _normalise_keys(initial)
 
         for env_name, env_config in environments.items():
+            env_config = (
+                {} if env_config is None else env_config
+            )  # ensures an env config can be nullable
```
WillGibson, line 173: Would this be more obvious and not need the comment as...
```suggestion
            if env_config is None:
                env_config = {}
```

tests/platform_helper/test_command_copilot.py
```diff
@@ -1050,6 +1050,94 @@ def test_appconfig_ip_filter_policy_is_applied_to_each_service_by_default(self,
...
+        """ALB validation should allow additional address list (optional)"""
+        fakefs.create_file(
+            EXTENSION_CONFIG_FILENAME,
+            contents="""
+    alb:
+      type: alb
+      environments:
+        development:
+          cdn_domains_list: 
+            test.domain.uktrade.digital: "test.domain.uktrade.digital"
```
WillGibson, line 1076: I think this would normally be like this...
```suggestion
            test.domain.uktrade.digital: "domain.uktrade.digital"
```

tests/platform_helper/test_command_copilot.py
```diff
@@ -1050,6 +1050,94 @@ def test_appconfig_ip_filter_policy_is_applied_to_each_service_by_default(self,
...
+          additional_address_list: ["another.domain"]
+    """,
+        )
+        fakefs.add_real_file(FIXTURES_DIR / "valid_workspace.yml", False, "copilot/.workspace")
+
```
WillGibson, line 1081: Can we drop these line breaks to keep the arrange parts grouped?

tests/platform_helper/test_command_copilot.py
```diff
@@ -1050,6 +1050,94 @@ def test_appconfig_ip_filter_policy_is_applied_to_each_service_by_default(self,
...
+        """Prod should be allowed to be defined without a config specified."""
+        fakefs.create_file(
+            EXTENSION_CONFIG_FILENAME,
+            contents="""
+    alb:
+      type: alb
+      environments:
+        default:
+          cdn_domains_list: 
+            test.domain.uktrade.digital: "test.domain.uktrade.digital"
```
WillGibson, line 1119:
```suggestion
            test.domain.uktrade.digital: "domain.uktrade.digital"
```

tests/platform_helper/test_command_copilot.py
```diff
@@ -1050,6 +1050,94 @@ def test_appconfig_ip_filter_policy_is_applied_to_each_service_by_default(self,
...
+        development:
+         # empty config
```
WillGibson, line 1121: Indenting...
```suggestion
          # empty config
```

tests/platform_helper/test_command_copilot.py
```diff
@@ -1050,6 +1050,94 @@ def test_appconfig_ip_filter_policy_is_applied_to_each_service_by_default(self,
...
+        fakefs.create_file(
+            "copilot/web/manifest.yml",
+            contents=" ".join([yaml.dump(yaml.safe_load(WEB_SERVICE_CONTENTS))]),
+        )
+
+        result = CliRunner().invoke(copilot, ["make-addons"])
+
+        assert "Key 'development' error" not in result.output
+        assert "None should be instance of 'dict'" not in result.output
+        assert result.exit_code == 0
```
WillGibson, line 1139: I approve of the additional test for this, but I'm wondering whether we should be asserting on the behaviour we are expecting here instead of the errors that triggered the adding of the new test/fix.

Perhaps we could paramaterise the happy path test and pass in a test case with an empty config?

tests/platform_helper/test_command_copilot.py
samuele-mattiuzzo: I think that's the formatter doing its thing, I'll tell it to ignore it

ejayesh (approved):
Lovely, lgtm.

samuele-mattiuzzo:
> Lovely, lgtm.

@ejayesh I'm working on adding a more comprehensive test for the empty `alb` config, to check that the file gets generated without issues, when that's been re-reviewed, I'll merge this in 👍🏻

tests/platform_helper/test_command_copilot.py
WillGibson: I think there's been a miscommunication on the second paragraph here. Probably my fault.

If you feel there should be distinct test cases, which I am inclined to agree with on second pass, there is no need to parameterise the tests as they only test one scenario each.

samuele-mattiuzzo: Happy to discuss this.
We have a fix for the regression tests so this is still blocked while the fix gets reviewed + merged

WillGibson (approved):

PR #439, opened 2024-05-21
feat: Cleaning out pipelines config before regenerating

antroy-madetech:
Change ensures that any previously generated or manually modified config is cleaned out of the copilot directory before generating the new config.


tests/platform_helper/fixtures/unnecessary_file.yml
```diff
@@ -0,0 +1 @@
+Not required
```
WillGibson, line 1:
```suggestion
Not required

```

dbt_platform_helper/commands/pipeline.py
```diff
@@ -49,20 +50,44 @@ def generate():
...
+                codestar_connection_arn,
+                git_repo,
+                codebase,
+                base_path,
+                pipelines_dir,
+                templates,
             )
 
 
+def _clean_pipeline_config(pipelines_dir):
```
WillGibson, line 85: If you declared this in the method above, it would actually kind of be private, certainly not callable from elsewhere.

tests/platform_helper/test_command_pipeline.py
```diff
@@ -167,11 +166,14 @@ def test_pipeline_generate_with_empty_pipelines_yml_does_nothing(
...
 @patch("dbt_platform_helper.commands.pipeline.git_remote", return_value="uktrade/test-app-deploy")
-def test_pipeline_generate_overwrites_any_existing_config_files(
+def test_pipeline_generate_deletes_any_existing_config_files_and_writes_new_ones(
     git_remote, get_aws_session_or_abort, fakefs
 ):
     mock_codestar_connections_boto_client(get_aws_session_or_abort, ["test-app"])
     setup_fixtures(fakefs)
+    fakefs.add_real_file(
+        FIXTURES_DIR / "unnecessary_file.yml", False, "copilot/pipelines/unnecessary_file.yml"
+    )
```
WillGibson, line 176: Can we create a file without needing the actual `unnecessary_file.yml` fixture?

dbt_platform_helper/commands/pipeline.py
antroy-madetech: Python convention is that a preceding underscore means the method is "private". I'm not a big fan of nested method declarations in general. May as well inline it in this case. Extracting it is for readability.

tests/platform_helper/test_command_pipeline.py
antroy-madetech: I'll look. Not massively familiar with the fakefs.

samuele-mattiuzzo (approved):

tests/platform_helper/test_command_pipeline.py
samuele-mattiuzzo: `fakefs_create_file('filename', contents=""" some content """)` might be what you're after here, it's used in several tests

dbt_platform_helper/commands/pipeline.py
samuele-mattiuzzo: extracting it also makes the method testable directly with fair ease, the `_function()` convention is just a way developers convey intention between each other, as there isn't anything at compiler level preventing this being called.

WillGibson: Nothing outside of `generate()` needs it, so it probably shouldn't be accessible outside of `generate()`. I would prefer to inline it that make a "not really private" method.

samuele-mattiuzzo: Personally I'm happier with the `_methodname` + extraction version than putting it inside something else, as I do know what this implies in Python. Maybe a comment that reiterates the internal only purpose?

WillGibson (approved):

PR #442, opened 2024-05-22
fix: Update get_s3_kms_alias_arns to check for environment name

tony-griffin:
- Add check that the current environment was in the list of copilot environments
- Create bucket name from environment in config and add it to kms key alias

### DIFF
--- a/dbt_platform_helper/commands/copilot.py
+++ b/dbt_platform_helper/commands/copilot.py
@@ -211,14 +211,15 @@ def is_terraform_project() -> bool:
     return Path("./terraform").is_dir()
 
 
-def _get_s3_kms_alias_arns(session, application_name, extension_name):
+def _get_s3_kms_alias_arns(session, application_name, config):
     application = load_application(application_name, session)
 
     arns = {}
     for environment_name, environment in application.environments.items():
-        client = environment.session.client("kms")
-
-        alias_name = f"alias/{application_name}-{environment_name}-{extension_name}-key"
+        if environment_name in config:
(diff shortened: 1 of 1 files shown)


gabelton (approved):

dbt_platform_helper/commands/copilot.py
```diff
@@ -211,14 +211,15 @@ def is_terraform_project() -> bool:
...
 
     arns = {}
     for environment_name, environment in application.environments.items():
-        client = environment.session.client("kms")
-
-        alias_name = f"alias/{application_name}-{environment_name}-{extension_name}-key"
+        if environment_name in config:
+            bucket_name = config[environment_name]["bucket_name"]
+            client = environment.session.client("kms")
+            alias_name = f"alias/{application_name}-{environment_name}-{bucket_name}-key"
```
WillGibson, line 222: Looks to me like the `try except else` might belong under the if statement as well?

Otherwise, I think for an environment which is not in config, they would just execute with the `alias_name` from the pervious environment. Or we could just skip it for `environment_name not in config`...

```suggestion
        if environment_name not in config:
            continue
        
        bucket_name = config[environment_name]["bucket_name"]
        client = environment.session.client("kms")
        alias_name = f"alias/{application_name}-{environment_name}-{bucket_name}-key"
```

I would be inclined to move `client = environment.session.client("kms")` above the for loop too.

tony-griffin: @WillGibson I'll create a new branch and implement this

PR #448, opened 2024-05-30
feat: Add slack notification commands

antroy-madetech:
Ported from ci-imagebuilder. Modified to be commands, but core code is broadly the same.

The main differences are that the command explicitly has all arguments and options supplied on the cli, rather than having variables coming in via env vars.

### DIFF
--- a/tests/platform_helper/test_command_notify.py
+++ b/tests/platform_helper/test_command_notify.py
@@ -1,4 +1,3 @@
-import unittest
 from unittest.mock import patch
 
 import pytest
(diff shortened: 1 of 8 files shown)


tests/platform_helper/test_command_notify.py
```diff
@@ -0,0 +1,261 @@
...
+@patch("dbt_platform_helper.commands.notify._get_slack_client")
+class TestNotify(unittest.TestCase):
+    def test_getting_build_url(self, webclient):
+        self.assertEqual(
+            get_build_url("arn:aws:codebuild:region:000000000000:build/project:example-build-id"),
+            "https://region.console.aws.amazon.com/codesuite/codebuild/000000000000"
+            "/projects/project/build/project%3Aexample-build-id",
+        )
+
+    def test_sending_progress_updates_with_no_optional_elements(self, webclient):
```
gabelton, line 20: Is it possible to parametrize some of these progress updates tests?

antroy-madetech: I had thought about it, but thought it'd get messy. I'll take a look and see if I can do it neatly.

gabelton (approved):

PR #452, opened 2024-06-04
test: DBTP-681 Run deploy environment pipeline in regression tests

WillGibson:
Adds the next part of https://uktrade.atlassian.net/browse/DBTP-681.

Includes breaking out the parts of the regression tests script to individual scripts covering discreet "stages" for easier grokking.


images/debian-python/README.md
```diff
@@ -1,13 +1,17 @@
-#### [`Dockerfile.debian`](Dockerfile.debian)
```
lgarvey, line 1: Is there any value in adding the SSM agent in case we want to debug inside it?

regression_tests/stages/clone_demodjango_deploy.sh
```diff
@@ -0,0 +1,14 @@
...
+
+set -e
+
+echo -e "\nGet CodeStar connection details"
+codestar_connections=$(aws codestar-connections list-connections --provider-type GitHub --query "Connections[? ConnectionStatus == 'AVAILABLE']")
+aws_account=$(echo "$codestar_connections" | jq -r ".[0].OwnerAccountId")
+codestar_arn=$(echo "$codestar_connections" | jq -r ".[0].ConnectionArn")
+codestar_connection_id=$(echo "${codestar_arn##*/}")
+
+echo -e "\nClone demodjango_deploy"
```
lgarvey, line 11: Question, isn't this the generic build image? If so, why is it working directly wtih demodjango-deploy? 🤔

samuele-mattiuzzo (approved):

images/debian-python/README.md
WillGibson: Almost certainly, but outside the scope of this pull request. Maybe we could consider adding this to all of the images we use in CodeBuild?

regression_tests/stages/clone_demodjango_deploy.sh
WillGibson: The regression tests test the combination of four codebases:

* `demodjango`
* `demodjango-deploy`
* `terraform-platform-modules`
* `platform-tools`

We will be triggering them on merge to main on any of those in due course.

PR #454, opened 2024-06-06
chore: Add DMAS migration maintenance template

garethpn:
Maintenance template added for DMAS migration based on wording provided by the service team.


acodeninja (approved):
Assuming you're able to fix the PR title, this looks ok. Maybe for a future PR we can make the text configurable with a reasonable default. I'm thinking something like:

```
platform-helper environment offline --template migration --text "Sorry, the service is unavailable.<br />You will be able to use the service from 10am on Tuesday 18 June."
```

PR #457, opened 2024-06-07
feat: DBTP-612 allow wildcard in branch names

samuele-mattiuzzo:


dbt_platform_helper/templates/pipelines/codebase/overrides/stack.ts
```diff
@@ -54,9 +54,17 @@ export class TransformedStack extends cdk.Stack {
         );
 
         for (const branch of watchedBranches) {
+            let sanitisedBranch = branch;
+
+            if (branch?.endsWith('*')){
+                sanitisedBranch?.replace('*', '.*');
+            } else {
+                sanitisedBranch = sanitisedBranch?.concat('$');
+            }
```
samuele-mattiuzzo, line 63: @acodeninja I think this is enough, but I also recall our conversation around the branch having to allow for `feat-whatever` and `feat` and I think this is what covers it

dbt_platform_helper/templates/pipelines/codebase/overrides/stack.ts
```diff
@@ -54,9 +54,17 @@ export class TransformedStack extends cdk.Stack {
         );
 
         for (const branch of watchedBranches) {
+            let sanitisedBranch = branch;
```
gabelton, line 57: Is there any way of extracting and unit testing this logic to check it covers the edge cases described on the ticket?

gabelton: Or maybe you can get away with just adding a few more scenarios / parameters to your existing `test_command_pipeline` tests?

samuele-mattiuzzo: I hope so, it's something I'm looking into now. None of `stack.ts` is tested here, so I'll see and assess, if it's a big task I might add a separate ticket for it

samuele-mattiuzzo: @gabelton yes valid suggestion, that would however test the regex not the changes made in `stack.ts`.

I still want to refactor the validation tests a little bit so I still have some work to do on this

acodeninja (approved):
LGTM

PR #465, opened 2024-06-17
chore: DBTP-1011 signpost autogenerated files

samuele-mattiuzzo:
Addresses [DBTP-1011](https://uktrade.atlassian.net/browse/DBTP-1011).

---
### Title:
### Description:
### Tasks:
- [ ] Manually run the commented out sections of the [pull request regression tests](https://github.com/uktrade/platform-tools/blob/main/regression_tests/pull_request_tests.sh) (unless it's not a code change)

[DBTP-1011]: https://uktrade.atlassian.net/browse/DBTP-1011?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


WillGibson:
This is looking good subject to the line breaks question and also including the following auto generated files...

* `copilot/web/overrides/cfn.patches.yml`
* `copilot/pipelines/application/*` Perhaps could be achieved simply by dropping a text file into the root of that directory if do all the files is tricky

samuele-mattiuzzo:
> This is looking good subject to the line breaks question and also including the following auto generated files...
> 
> * `copilot/web/overrides/cfn.patches.yml`
> * `copilot/pipelines/application/*` Perhaps could be achieved simply by dropping a text file into the root of that directory if do all the files is tricky

The other option would be to modify our `.yml` template files directly so that `{% version_tag %}` only prints out the version, and add the warning line manually in them.

Which, in hindsight, is probably the most tedious yet the best way forward

WillGibson:
> The other option would be to modify our `.yml` template files directly so that `{% version_tag %}` only prints out the version, and add the warning line manually in them.
> 
> Which, in hindsight, is probably the most tedious yet the best way forward

Wouldn't this suggestion result on a load of duplicated boiler plate?

samuele-mattiuzzo:
> > The other option would be to modify our `.yml` template files directly so that `{% version_tag %}` only prints out the version, and add the warning line manually in them.
> > Which, in hindsight, is probably the most tedious yet the best way forward
> 
> Wouldn't this suggestion result on a load of duplicated boiler plate?

@WillGibson can you give it another pair of eyes? I've added a new template tag that does the multiline trick.

@gabelton yes, maybe, however it's not straight forward to have a multiline template tag that is yaml-compliant, so for now I'm perfectly happy if this stays as is as I don't want it to take days

dbt_platform_helper/templates/svc/manifest-public.yml
```diff
@@ -1,3 +1,4 @@
+# {% extra_header %}
```
gabelton, line 1: Do we want this in this file and manifest-backend? We do manually edit those don't we? i.e. adding ip-filter or tweaking env vars etc.

gabelton:
> @gabelton yes, maybe, however it's not straight forward to have a multiline template tag that is yaml-compliant, so for now I'm perfectly happy if this stays as is as I don't want it to take days

Could you add something on the same line or does that break linting rules? Otherwise yeah not a biggy

samuele-mattiuzzo:
> > @gabelton yes, maybe, however it's not straight forward to have a multiline template tag that is yaml-compliant, so for now I'm perfectly happy if this stays as is as I don't want it to take days
> 
> Could you add something on the same line or does that break linting rules? Otherwise yeah not a biggy

I could yes, it's not a biggie at all as it's just a comment, it's just ugly to have a too long line. Ideally, the new template tag I added would be able to spit out content with line breaks, but that's a future enhancement I suppose, to keep this ticket's scope contained

dbt_platform_helper/templates/svc/manifest-public.yml
samuele-mattiuzzo: Yes, true, I think I went a bit too far with the find / replace. Will update

WillGibson (approved):
Beautifully simple solution in the end :-)

gabelton (approved):

PR #468, opened 2024-06-17
fix: add dmas-migration to available environment template options

lgarvey:
Addresses a direct request in slack.

The dmas-migration template option is missing and needs to be added to enable DMAS cutover tomorrow.

I'm not sure if tests are needed, or whether this option will be removed or made a bit more generic post migration. 

It is an additional fix that allows the changes in this PR to work correctly: https://github.com/uktrade/platform-tools/commit/8b271c9f6bfd04974c3d8794b2f646dafc4fa9ac\

Post DMAS go live we should probably look at a more general solution - maybe allowing a template to be supplied, or a parameter for a date that the service will unavailable. 

---
### Title:
### Description:
### Tasks:
- [ ] Manually run the commented out sections of the [pull request regression tests](https://github.com/uktrade/platform-tools/blob/main/regression_tests/pull_request_tests.sh) (unless it's not a code change)

### DIFF
--- a/dbt_platform_helper/COMMANDS.md
+++ b/dbt_platform_helper/COMMANDS.md
@@ -652,7 +652,7 @@ platform-helper environment (offline|online|generate)
 ## Usage
 
 ```
-platform-helper environment offline --app <application> --env <environment> [--template (default|migration)] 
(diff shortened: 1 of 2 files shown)


garethpn (approved):
Thank you.  As you mentioned a more generic solution with a template file would be preferable

PR #473, opened 2024-06-19
fix: DBTP-1044 as a developer when i enable a maintenance page i should still be able to access the service (PT. 2)

gabelton:
Addresses [DBTP-1044](https://uktrade.atlassian.net/browse/DBTP-1044)

Follows on from previous maintenance page work, addressing gaps highlighted by DMAS migration https://github.com/uktrade/platform-tools/pull/453

New logic only whitelists env egress ips, not individual users' ips. Instead we just create custom header values and expect developers to add these, while accepting that this means Cirrus users can't bypass holding pages. (This is because of how zscaler manages ips on cirrus machines)

Docs change here https://github.com/uktrade/platform-documentation/pull/318

---
### Title:
- [X] Scope included as per [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [X] Ticket reference included (unless it's a quick out of ticket thing)
### Description:
- [N/A] If the work includes user interface changes, before and after screenshots included in description
- [X] Includes any applicable changes to the documentation in this code base
- [X] Includes link(s) to any applicable changes to the documentation in the [DBT Platform Documentation](https://platform.readme.trade.gov.uk/) (can be to a pull request)
### Tasks:

[DBTP-1044]: https://uktrade.atlassian.net/browse/DBTP-1044?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dbt_platform_helper/COMMANDS.md
```diff
@@ -656,8 +656,7 @@ platform-helper environment (offline|allow-ips|online|generate)
 ```
 platform-helper environment offline --app <application> --env <environment> --svc <service> 
                                     [--template (default|migration|dmas-migration)] 
-                                    [--allowed-ip <allowed_ip>] 
-                                    [--ip-filter] 
+                                    [--vpc <vpc>]
```
WillGibson, line 659: The docs merged as linked in the pull request description still refer to `--allowed-ip`.

gabelton: Thanks - updated https://github.com/uktrade/platform-documentation/pull/318

dbt_platform_helper/commands/environment.py
```diff
@@ -572,50 +525,50 @@ def add_maintenance_page(
...
-        else:
-            bypass_value = "".join(random.choices(string.ascii_lowercase + string.digits, k=12))
-            create_header_rule(
-                lb_client,
-                listener_arn,
-                target_group_arn,
-                "Bypass-Key",
-                [bypass_value],
-                "BypassIpFilter",
+                (index + 1) * (ip_index + 100),
```
samuele-mattiuzzo, line 551: part of me would like to suggest extracting this to a variable for readability and debugging purposes, another part of me knows it's not going to make a massive difference, so it's just a small suggestion

dbt_platform_helper/commands/environment.py
```diff
@@ -572,50 +525,50 @@ def add_maintenance_page(
...
+
+        click.secho(
+            f"\nUse a browser plugin to add `Bypass-Key` header with value {bypass_value} to your requests. For more detail, visit https://platform.readme.trade.gov.uk/ ",
+            fg="green",
+        )
 
     lb_client.create_rule(
         ListenerArn=listener_arn,
-        Priority=2,
+        Priority=20000,
```
samuele-mattiuzzo, line 571: what does `20000` stand for? how does it compare to the argument `priority`? we had `1` and `2` before, now it went up x10k times, maybe worth adding a little comment?

samuele-mattiuzzo: or even extracting it into its own constant since it's reused in tests as well

samuele-mattiuzzo (approved):
LGTM beside 2 small comments

dbt_platform_helper/commands/environment.py
gabelton: I've added a comment. The number is a little bit arbitrary, so didn't think it was worth assigning to a constant

samuele-mattiuzzo (approved):

dbt_platform_helper/commands/environment.py
```diff
@@ -331,6 +251,21 @@ def get_cert_arn(session, env_name):
     raise click.Abort
 
 
+def get_env_ips(vpc: str, application_environment: Environment):
+    vpc_name = vpc if vpc else application_environment.account_id
```
WillGibson, line 255: Under what circumstance would the VPC be named with an AWS account ID?

dbt_platform_helper/commands/environment.py
```diff
@@ -572,50 +525,50 @@ def add_maintenance_page(
...
+            listener_arn,
+            target_group_arn,
+            "Bypass-Key",
+            [bypass_value],
+            "BypassIpFilter",
+            index + 1,
+        )
+
+        click.secho(
+            f"\nUse a browser plugin to add `Bypass-Key` header with value {bypass_value} to your requests. For more detail, visit https://platform.readme.trade.gov.uk/ ",
```
WillGibson, line 565: Please link to the actual location of the information.

dbt_platform_helper/commands/environment.py
```diff
@@ -572,50 +525,50 @@ def add_maintenance_page(
...
+            fg="green",
+        )
 
     lb_client.create_rule(
         ListenerArn=listener_arn,
-        Priority=2,
+        Priority=20000,  # big number because we create multiple higher priority "AllowedIps" rules for each allowed ip for each service above.
```
WillGibson, line 571: Could we increment a `priority` variable above and do the following here?
```suggestion
        Priority=priority + 1,  # big number because we create multiple higher priority "AllowedIps" rules for each allowed ip for each service above.
```

dbt_platform_helper/commands/pipeline.py
```diff
@@ -92,7 +92,12 @@ def generate():
 def _clean_pipeline_config(pipelines_dir):
     if pipelines_dir.exists():
         click.echo("Deleting copilot/pipelines directory.")
-        rmtree(pipelines_dir)
+        try:
+            rmtree(pipelines_dir)
+            click.echo("Deleted!")
+        except Exception as e:
+            click.echo(f"Failed to delete {pipelines_dir}: {e}")
```
WillGibson, line 99: I'm curious as to how you got here?

tests/platform_helper/test_command_environment.py
WillGibson, line 1: At the moment we are not testing the actual result of running the commands, just that they have called `add_maintenance_page()` with the right parameters. 

We could probably delete the some of all of internals of that method and the test would still pass.

Can we test the change has actually been applied here somehow?

Can we add test for this to the regression tests? Yes. We should be able to take the environment offline after the smoke tests, run another test to check it's offline, maybe run a test to check we can still access it with the header set, then bring it online and re-run the smoke tests.

dbt_platform_helper/commands/environment.py
```diff
@@ -331,6 +251,21 @@ def get_cert_arn(session, env_name):
     raise click.Abort
 
 
+def get_env_ips(vpc: str, application_environment: Environment):
```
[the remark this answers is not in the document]
gabelton: I think we should still have individual unit tests for the private methods, trying to keep tests and methods as small as possible. 

if we stopped testing the helper methods, the unit tests for the commands themselves would have to be rewritten because currently they're just checking the command output and that helper methods are called with expected args. With the size of the commands, I'd prefer to stick with the current approach, as the unit tests for commands using moto would be enormous.

Agree that it would be good to add the commands to the regression tests though.

dbt_platform_helper/commands/pipeline.py
gabelton: Sorry just some crap from when I was testing that intermittent actions test failure a few days ago

dbt_platform_helper/commands/environment.py
gabelton: I thought the var name was a misnomer here and referred to account name instead of ID, but had the wrong idea. Will amend

tests/platform_helper/test_command_environment.py
gabelton: Lots of the tests in this file already followed that pattern of patching helper method calls, so I just followed that. I think those tests are ok, but agree it'd be a good idea to run maintenance commands as part of regression test.

WillGibson: Let's do it then!

Happy to pair on that.

dbt_platform_helper/commands/environment.py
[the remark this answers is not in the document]
WillGibson: I think we should really be properly testing the expected behaviour of the commands. Like I say, I have some thoughts about this codebase, just wish there was time to do a proof of concept on one of the simpler commands.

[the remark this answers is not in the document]
gabelton: Yeah it'd be good to chat more about what we think best practice for unit testing looks like and try to get more consistency across this package. When you say "properly" do you mean using moto for all the api calls or actually running the command again a realistic AWS test service? Or just regression tests?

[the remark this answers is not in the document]
WillGibson: Possibly using moto, not testing against a real AWS service for unit tests.

I have a feeling that if we started extracting out the behaviours needed by the commands to classes, with discrete responsibilities, and unit tested their public methods, we might end up with something that is easier to unit test, understand and maintain.

acodeninja (commented):
Came across an error when we have more than twenty loadbalancer rules and are running the env online command:

```
  File "/python/3.12.4/lib/python3.12/site-packages/dbt_platform_helper/commands/environment.py", line 428, in get_maintenance_page
    rules = lb_client.describe_tags(ResourceArns=[r["RuleArn"] for r in rules])["TagDescriptions"]
  File "/python/3.12.4/lib/python3.12/site-packages/botocore/client.py", line 565, in _api_call
    return self._make_api_call(operation_name, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/python/3.12.4/lib/python3.12/site-packages/botocore/client.py", line 1021, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (ValidationError) when calling the DescribeTags operation: Cannot have more than 20 resources described
```

WillGibson:
@gabelton I think we'll need to iterate through the paginated response from [describe-tags](https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-tags.html) to get them all.

dbt_platform_helper/commands/environment.py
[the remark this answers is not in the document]
WillGibson: Let's park this thread. If I end up doing a new command sometime I might look at how to structure it.

regression_tests/stages/run_maintenance_page_tests.sh
```diff
@@ -0,0 +1,33 @@
...
+MAINTENANCE_PAGE_BYPASS_VALUE=$(echo "$OUTPUT" | grep -oP 'Bypass-Key` header with value \K[^\s]+')
+
+echo -e "\nCheck maintenance page is working and that we can view the site with Bypass-Key header"
+cd "${CODEBUILD_SRC_DIR}/demodjango"
+./smoke_tests.sh ${TARGET_ENVIRONMENT} maintenance_pages ${MAINTENANCE_PAGE_BYPASS_VALUE}
+
+#TODO check env ip is whitelisted https://uktrade.atlassian.net/browse/DBTP-1161
+
+echo -e "\nRunning online command"
+echo "y" | AWS_PROFILE=platform-sandbox PLATFORM_TOOLS_SKIP_VERSION_CHECK=true platform-helper environment online --app demodjango --env ${TARGET_ENVIRONMENT}
```
WillGibson, line 29: I think it might be worth adding a `--yes` flag to the command, it's a pretty common pattern for allowing commands to be used in pipelines.

dbt_platform_helper/commands/environment.py
```diff
@@ -572,50 +531,52 @@ def add_maintenance_page(
...
             continue
 
-        if not ip_filter:
-            user_ip = get_public_ip()
-            allowed_ips = list(allowed_ips) + [user_ip]
-
+        user_ip = get_public_ip()
+        allowed_ips = list(allowed_ips) + [user_ip]
+        for ip_index, ip in enumerate(allowed_ips):
+            forwarded_rule_priority = (index + 1) * (ip_index + 100)
```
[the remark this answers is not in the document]
gabelton: @acodeninja can you remember if this is still needed?

regression_tests/stages/run_maintenance_page_tests.sh
WillGibson: Let's see what other people think. It's not a deal breaker for me.

I wouldn't want to scope creep into all the other commands in this branch, if you do do it, just to the `offline`/`online` commands.

I totally appreciate that this work has been a long slog too, and adds plenty value already.

regression_tests/pull_request_tests.sh
```diff
@@ -3,16 +3,22 @@
 # exit early if something goes wrong
 set -e
 
+export TARGET_ENVIRONMENT=${TARGET_ENVIRONMENT:-toolspr}
```
WillGibson, line 6: The use of this via the AWS Console to fire off regression tests using a different target environment needs adding to the README.

tests/platform_helper/test_command_environment.py
```diff
@@ -1365,15 +1122,98 @@ def test_create_header_rule(self, capsys):
...
+        captured = capsys.readouterr()
+
+        assert "No parameter found with name: /vpc/EGRESS_IPS\n" in captured.out
+
+    @patch("boto3.client")
+    def test_get_rules_tag_descriptions(self, mock_boto_client):
+        from dbt_platform_helper.commands.environment import get_rules_tag_descriptions
+
+        mock_client = Mock()
+        mock_client.describe_tags.return_value = {"TagDescriptions": ["TagDescriptions"]}
```
WillGibson, line 1207: Might be worth using different strings so we can be confident it didn't just use the same one twice.

WillGibson: Forgive me if I'm being naive here, but should we have 21 tag descriptions?

regression_tests/pull_request_tests.sh
gabelton: Btw I ran into an issue with that because my env isn't part of a pipeline

tests/platform_helper/test_command_environment.py
gabelton: I think it's the number of rules that's important to replicating the bug, not the number of descriptions returned. So it's import that 21 rules means the method is called twice.

Although you're right I should make it return a different string the second time. Was being lazy

gabelton: I think it's the number of rules that's important to replicating the bug, not the number of descriptions returned. So it's important that 21 rules means the method is called twice.

Although you're right I should make it return a different string the second time. Was being lazy

dbt_platform_helper/commands/environment.py
```diff
@@ -511,18 +444,29 @@ def remove_maintenance_page(session: boto3.Session, listener_arn: str):
...
 
-def get_public_ip():
-    response = requests.get("https://api.ipify.org")
-    return response.text
+def get_rules_tag_descriptions(rules: list, lb_client):
+    tag_descriptions = []
+    chunk_size = 20
+
+    for i in range(0, len(rules), chunk_size):
+        chunk = rules[i : i + chunk_size]
```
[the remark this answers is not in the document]
gabelton: This is [slice notation](https://www.geeksforgeeks.org/python-list-slicing/). So on first iteration it will be 0:20, then 20:40 on second and so on. It will iterate through the rules taking slices of max 20 elements at a time

dbt_platform_helper/commands/environment.py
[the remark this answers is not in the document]
acodeninja: It might not be, this was put in to ensure enough space between priorities as two entries cannot have the same priority values. From the looks of it you are still doing a nested loop so the requirement is still needed.

An alternative could be to have a incrementing counter instead, which might be easier to grok.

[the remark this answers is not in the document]
WillGibson: Thanks @acodeninja. Definitely worth a refactor for grokability.

WillGibson (commented):
Nearly there :-)

dbt_platform_helper/commands/environment.py
```diff
@@ -594,50 +562,54 @@ def add_maintenance_page(
...
     services: List[Service],
     allowed_ips: tuple,
-    ip_filter: bool,
     template: str = "default",
 ):
     lb_client = session.client("elbv2")
     maintenance_page_content = get_maintenance_page_template(template)
+    bypass_value = "".join(random.choices(string.ascii_lowercase + string.digits, k=12))
+
+    counter = 1
```
WillGibson, line 571: I can't see this getting incremented anywhere

dbt_platform_helper/commands/environment.py
```diff
@@ -594,50 +562,54 @@ def add_maintenance_page(
...
         if not target_group_arn:
             continue
 
-        if not ip_filter:
-            user_ip = get_public_ip()
-            allowed_ips = list(allowed_ips) + [user_ip]
-
+        allowed_ips = list(allowed_ips)
+        for ip_index, ip in enumerate(allowed_ips):
+            forwarded_rule_priority = counter * (ip_index + 100)
```
WillGibson, line 582: I'm struggling to understand the reasoning behind this maths.

Can you write the criteria for `forwarded_rule_priority` in human facing terms?

gabelton: Each service needs a Bypass-Key rule created. This is assigned the counter priority. Then for each allowed ip for each service, we create an X-Forwarded-For rule.

So say we have two services and three allowed ips.

1st iteration:
- Create bypass rule with priority 1
- Then rules for each ip with priorities: 100, 101, 102

2nd:
- Create bypass rule with priority 2
- Then rules for each ip with priorities: 200, 202, 204

3rd:
- Create bypass rule with priority 3
- Then rules for each ip with priorities: 300, 303, 306

It's just to ensure that there's no chance of clashing priority numbers, even if we have loads of allowed ips

WillGibson: Thanks. That confirms the feeling I had on Friday.

I suggest we rename `counter` to `service_number` so it's more obvious what it represents.

Also, 100 is a magic number, please can we extract that to a variable `max_allowed_ips`.

Also, I'm not sure how many allowed IPs we might have, but is 100 per service definitely enough? @lgarvey would have a good feel for that I think.

As things stand, if the second service has >=50 allowed IPs, it will start clashing with the third service...

```
2 * (50 + 100) = 300
```

This will only get worse as the number of services increases.

Perhaps...

```suggestion
            forwarded_rule_priority = (service_number * max_allowed_ips) + ip_index
```

Then...
```
(2 * 100) + 50 = 250
```

WillGibson: `max_allowed_ips` could probably just be 1000 and I think we would be covered whilst still having the service number as the first digit.

dbt_platform_helper/commands/environment.py
```diff
@@ -594,50 +562,56 @@ def add_maintenance_page(
...
+        )
+
+        counter += 1
+
+        click.secho(
+            f"\nUse a browser plugin to add `Bypass-Key` header with value {bypass_value} to your requests. For more detail, visit https://platform.readme.trade.gov.uk/activities/holding-and-maintenance-pages/",
+            fg="green",
+        )
 
+    fixed_rule_priority = 20000
```
WillGibson, line 611: Just in case we had a squillion services...
```suggestion
    fixed_rule_priority = ((service_number + 5) * max_allowed_ips)
```

For the three services example, that would yield `800`, which should be obviously enough no related to one of the services.

dbt_platform_helper/commands/environment.py
gabelton: I think 100 is fine. The only ips we're allowing are the env ips and there won't be more than 100 of those

dbt_platform_helper/commands/environment.py
```diff
@@ -594,50 +562,57 @@ def add_maintenance_page(
...
             continue
 
-        if not ip_filter:
-            user_ip = get_public_ip()
-            allowed_ips = list(allowed_ips) + [user_ip]
-
+        allowed_ips = list(allowed_ips)
+        max_allowed_ips = 100
+        for ip_index, ip in enumerate(allowed_ips):
+            forwarded_rule_priority = service_number * (ip_index + max_allowed_ips)
```
WillGibson, line 583: For the reasons mentioned on https://github.com/uktrade/platform-tools/pull/473#discussion_r1686196863, this maths should still be 
```suggestion
            forwarded_rule_priority = (service_number * max_allowed_ips) + ip_index
```

WillGibson (approved):

PR #477, opened 2024-06-21
fix: DBTP-1093 - Add postgres version 16 to conduit image

JohnStainsby:
Addresses https://uktrade.atlassian.net/browse/DBTP-1093

---
### Title:
### Description:
### Tasks:

### DIFF
--- a/images/conduit/postgres/Dockerfile
+++ b/images/conduit/postgres/Dockerfile
@@ -2,7 +2,10 @@ FROM public.ecr.aws/docker/library/debian:12-slim
 
 RUN apt-get update && \
     apt-get upgrade -y && \
-    apt-get install -y jq procps postgresql-client && \
+    # Add repository for postgres version 16
+    apt-get install -y postgresql-common && \
+    /usr/share/postgresql-common/pgdg/apt.postgresql.org.sh -y && \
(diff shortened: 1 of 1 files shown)


images/conduit/postgres/Dockerfile
```diff
@@ -2,7 +2,10 @@ FROM public.ecr.aws/docker/library/debian:12-slim
 
 RUN apt-get update && \
     apt-get upgrade -y && \
-    apt-get install -y jq procps postgresql-client && \
+    # Add repository for postgres version 16
+    apt-get install -y postgresql-common && \
+    /usr/share/postgresql-common/pgdg/apt.postgresql.org.sh -y && \
+    apt-get install -y jq procps postgresql-client-16 && \
```
WillGibson, line 8: Is client 16 compatible with earlier versions of Postgres?

JohnStainsby: The client is backwards compatible with older server versions, at the moment digital workspace is running 14.4 and the client is 15.x

I'll do a test this morning against 14 and 15 to double check

WillGibson: Will approve subject to that working.

WillGibson (approved):

PR #490, opened 2024-07-03
feat: Added an 'account' parameter to the environment_pipelines config

antroy-madetech:
Addresses [DBTP-1136](https://uktrade.atlassian.net/browse/DBTP-1136)

In order to filter the pipelines to determine which to deploy to an account, we need to tag those pipelines with an "account" parameter. This PR adds that parameter.

Also moved some functions around so I could streamline some of the platform-config.yml validaton.
---
### Title:
### Description:
### Tasks:

[DBTP-1136]: https://uktrade.atlassian.net/browse/DBTP-1136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dbt_platform_helper/commands/pipeline.py
```diff
@@ -95,26 +88,6 @@ def _clean_pipeline_config(pipelines_dir):
...
-
-            unique_codebase_environments = sorted(list(set(codebase_environments)))
-
-            if sorted(codebase_environments) != sorted(unique_codebase_environments):
-                abort_with_error(
-                    f"The {PLATFORM_CONFIG_FILE} file is invalid, each environment can only be "
-                    "listed in a single pipeline per codebase"
-                )
-
-
```
antroy-madetech, line 117: Moved to validation.py with the other validation code.

dbt_platform_helper/utils/files.py
```diff
@@ -14,36 +14,6 @@
...
-
-
-def load_and_validate_config(path, schema):
-    conf = yaml.safe_load(Path(path).read_text())
-
-    schema.validate(conf)
-
-    return conf
-
-
```
antroy-madetech, line 46: Moved these functions into validation.py

dbt_platform_helper/commands/pipeline.py
```diff
@@ -190,10 +163,3 @@ def _create_file_from_template(
     ).render(template_data)
     message = mkfile(base_path, pipelines_dir / file_name, contents, overwrite=True)
     click.echo(message)
-
-
-def _safe_load_config(filename, schema):
-    try:
-        return load_and_validate_config(filename, schema)
-    except ParserError:
-        abort_with_error(f"The {filename} file is invalid")
```
antroy-madetech, line 199: Moved this exception handling into the validation.py function ' load_and_validate_platform_config' so it can be used by anything that wants to validate platform-config.yml

dbt_platform_helper/utils/validation.py
```diff
@@ -456,6 +465,100 @@ def is_between(value):
 )
 
 
+def validate_platform_config(config):
+    get_aws_session_or_abort()  # Ensure we have a valid session as validation requires it.
+    PLATFORM_CONFIG_SCHEMA.validate(config)
+    _validate_environment_pipelines(config)
+    _validate_codebase_pipelines(config)
+
```
antroy-madetech, line 473: New function that does a comprehensive validation.

tests/platform_helper/utils/test_files.py
```diff
@@ -2,38 +2,12 @@
...
-    """Test that, given the path to a valid yaml file, load_and_validate_config
-    returns the loaded yaml unmodified."""
-
-    path = FIXTURES_DIR / yaml_file
-    validated = load_and_validate_config(path, schema)
-
-    with open(path, "r") as fd:
-        conf = yaml.safe_load(fd)
-
-    assert validated == conf
```
antroy-madetech, line 36: Moved to test_validation.py

tests/platform_helper/utils/test_files.py
```diff
@@ -142,79 +116,3 @@ def test_is_terraform_project(fakefs, platform_config_content, expected_result):
...
-    for file in files:
-        fakefs.create_file(file)
-
-    with pytest.raises(SystemExit):
-        config_file_check()
-
-    console_message = capsys.readouterr().out
-
-    for expected_message in expected_messages:
-        assert expected_message in console_message
```
antroy-madetech, line 220: Moved to test_validation.py

dbt_platform_helper/utils/constants.py
WillGibson, line 1: I'm wondering if there is a more obvious "central" location for this file?

dbt_platform_helper/utils/validation.py
```diff
@@ -456,6 +465,100 @@ def is_between(value):
...
+            for env in pipeline.get("environments", {}).keys():
+                env_account = (
+                    enriched_config.get("environments", {})
+                    .get(env, {})
+                    .get("accounts", {})
+                    .get("deploy", {})
+                    .get("name")
+                )
+                if env_account and not env_account == account:
+                    bad_envs.append(env)
```
WillGibson, line 492: Maybe some renaming for clarity and defaulting an account name so the if statement reads clearer?
```suggestion
        pipeline_account = pipeline.get("account", None)
        if pipeline_account:
            for env in pipeline.get("environments", {}).keys():
                env_account = (
                    enriched_config.get("environments", {})
                    .get(env, {})
                    .get("accounts", {})
                    .get("deploy", {})
                    .get("name")
                )
                if not env_account == pipeline_account:
                    bad_envs.append(env)
```

dbt_platform_helper/utils/validation.py
```diff
@@ -456,6 +465,100 @@ def is_between(value):
...
+        message = "The following pipelines are misconfigured:"
+        for pipeline, detail in bad_pipelines.items():
+            envs = detail["bad_envs"]
+            acc = detail["account"]
+            message += f"  '{pipeline}' - these environments are not in the '{acc}' account: {', '.join(envs)}\n"
+        abort_with_error(message)
+
+
+def _validate_codebase_pipelines(pipeline_config):
+    if not (CODEBASE_PIPELINES_KEY in pipeline_config or ENVIRONMENTS_KEY in pipeline_config):
```
WillGibson, line 505: Wouldn't we need both of these?
```suggestion
    if not (CODEBASE_PIPELINES_KEY in pipeline_config and ENVIRONMENTS_KEY in pipeline_config):
```

WillGibson (commented):
About half way down, that's as much diff as I can meaningfully review on this pass.

dbt_platform_helper/utils/constants.py
antroy-madetech: That's what I was wondering, but there isn't really. Hence making one. Open to suggestions though.

dbt_platform_helper/utils/validation.py
antroy-madetech: Not sure tbh. I just moved this function. It's existing functionality.

antroy-madetech: Actually re-reading it, it's:
`not (A or B)` which is equivalent to `not A and not B` which is correct.

dbt_platform_helper/utils/constants.py
WillGibson: I think it could live in `dbt_platform_helper`, like `exceptions.py`.

dbt_platform_helper/utils/validation.py
WillGibson: It looks wrong to me, I think we need `CODEBASE_PIPELINES_KEY` and `ENVIRONMENTS_KEY` in the code that follows

Don't we want it to error if `not A or not B`?

Then we don't need the `if CODEBASE_PIPELINES_KEY in pipeline_config` which at the time of writing is on line 508.

dbt_platform_helper/utils/constants.py
antroy-madetech: Move it up a level? Makes sense.

WillGibson: Yes.

dbt_platform_helper/utils/validation.py
WillGibson: Not sure if you resolved this before my last comment, so unresolving.

antroy-madetech: Pretty sure the current logic is sound. The command generates *copilot* pipelines. So it needs at least one of codebase_pipelines (where it gets codebase pipelines config from) or environments (where it gets environment pipeline config from) but not both.

WillGibson: We have agreed verbally to remove this check as it should be possible to run `platform-helper generate` before you have pipelines configured.

tests/platform_helper/utils/test_validation.py
```diff
@@ -419,6 +423,208 @@ def test_validate_s3_bucket_name_multiple_failures():
...
+@patch("dbt_platform_helper.utils.validation.get_aws_session_or_abort", new=Mock())
+def test_validate_platform_config_success(valid_platform_config):
+    validate_platform_config(valid_platform_config)
     # No assertions - validate will error if config is invalid.
+
+
+@pytest.mark.parametrize(
+    "account, envs, exp_bad_envs",
+    [
+        ("non-prod", ["dev"], ["dev"]),
```
WillGibson, line 435: I'm not sure about this test case. As far as I can tell there is no `non-prod` account, so this is really testing a non existent account.

Maybe this would clarify intent...
```suggestion
        ("account-does-not-exist", ["dev"], ["dev"]),
```

tests/platform_helper/utils/test_validation.py
```diff
@@ -419,6 +423,208 @@ def test_validate_s3_bucket_name_multiple_failures():
...
+        ("non-prod-acc", ["prod"], ["prod"]),
+        ("prod-acc", ["dev", "prod"], ["dev"]),
+        ("prod-acc", ["dev", "staging", "prod"], ["dev", "staging"]),
+        ("non-prod-acc", ["dev", "prod"], ["prod"]),
```
WillGibson, line 439: I'm not convinced two of these scenarios add any value...
```suggestion
        ("prod-acc", ["dev", "staging", "prod"], ["dev", "staging"]),
        ("non-prod-acc", ["dev", "prod"], ["prod"]),
```

tests/platform_helper/utils/test_validation.py
```diff
@@ -419,6 +423,208 @@ def test_validate_s3_bucket_name_multiple_failures():
...
+    assert (
+        f"  'main' - these environments are not in the '{account}' account: {', '.join(exp_bad_envs)}"
+        in message
+    )
+
+
+@patch("dbt_platform_helper.utils.validation.warn_on_s3_bucket_name_availability", new=Mock())
+@patch("dbt_platform_helper.utils.validation.abort_with_error")
+@patch("dbt_platform_helper.utils.validation.get_aws_session_or_abort", new=Mock())
+def test_validate_platform_config_fails_if_pipeline_account_does_not_match_environment_accounts_multiple_pipelines(
```
WillGibson, line 471: Isn't this covered by the test above?

tests/platform_helper/utils/test_validation.py
```diff
@@ -419,6 +423,208 @@ def test_validate_s3_bucket_name_multiple_failures():
...
+
+    assert "The following pipelines are misconfigured:" in message
+    assert f"  'main' - these environments are not in the 'non-prod' account: dev" in message
+    assert f"  'prod' - these environments are not in the 'prod' account: dev, staging" in message
+
+
+@pytest.mark.parametrize(
+    "account, envs",
+    [
+        ("non-prod-acc", ["dev"]),
```
WillGibson, line 501: Not sure this scenario adds any value...
```suggestion
```

tests/platform_helper/utils/test_validation.py
```diff
@@ -10,16 +10,20 @@
 from moto import mock_aws
 from schema import SchemaError
 
+from dbt_platform_helper.constants import PLATFORM_CONFIG_FILE
```
WillGibson, line 13: Should we rely on this definition in the tests?

You could change it and I think the tests would still pass.

I'm wondering if the tests should have their own declaration of this?

tests/platform_helper/utils/test_validation.py
antroy-madetech: There's only a single pipeline in the tests above. That exercises multiple environment combinations in a single pipeline. This tests a multi-pipeline setup. Could merge them as one test but the mark.parametrize would get gnarly.

tests/platform_helper/utils/test_validation.py
antroy-madetech: My thinking is that these tests don't care what the filename is just the logic of what happens when that is validated. If we decide to change the filename we don't then have to work out if these tests broke because of logic, or broke due to just the filename change (which we don't care about here)

antroy-madetech: I mean, perhaps a separate test "test_filename_is_pipeline_config_yml" that hardcodes the name would give a single test that would challenge people to decide if this is really what they wanted to do (in case of introducing a typo or something by accident)

tests/platform_helper/utils/test_validation.py
WillGibson: Perhaps we can rename the tests to express the intent more clearly?

tests/platform_helper/utils/test_validation.py
```diff
@@ -419,6 +423,216 @@ def test_validate_s3_bucket_name_multiple_failures():
...
+    assert (
+        f"  'main' - these environments are not in the '{account}' account: {', '.join(exp_bad_envs)}"
+        in message
+    )
+
+
+@patch("dbt_platform_helper.utils.validation.warn_on_s3_bucket_name_availability", new=Mock())
+@patch("dbt_platform_helper.utils.validation.abort_with_error")
+@patch("dbt_platform_helper.utils.validation.get_aws_session_or_abort", new=Mock())
+def test_validate_platform_config_fails_if_pipeline_account_catches_all_errors_across_multiple_pipelines(
```
WillGibson, line 469:
```suggestion
def test_validate_platform_config_catches_all_errors_across_multiple_pipelines(
```

WillGibson (approved):
Last suggestion. I will let you decide. Approving.

PR #496, opened 2024-07-05
fix: DBTP-1148 platform-helper environment generate should exit it not logged in

simon106:
Addresses https://uktrade.atlassian.net/browse/DBTP-1148

Moved check into generate function

---
### Title:
### Description:
### Tasks:

### DIFF
--- a/dbt_platform_helper/commands/environment.py
+++ b/dbt_platform_helper/commands/environment.py
@@ -377,7 +377,7 @@ def generate_terraform(name):
     _generate_terraform_environment_manifests(conf["application"], name, env_config)
 
 
-def _generate_copilot_environment_manifests(name, env_config):
+def _generate_copilot_environment_manifests(name, env_config, session):
     env_template = setup_templates().get_template("env/manifest.yml")
     vpc_name = env_config.get("vpc", None)
     vpc_id = get_vpc_id(session, name, vpc_name)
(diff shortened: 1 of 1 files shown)


dbt_platform_helper/commands/environment.py
```diff
@@ -377,7 +378,6 @@ def generate_terraform(name):
 
 
 def _generate_copilot_environment_manifests(name, env_config):
```
gabelton, line 380: This function needs session arg adding

```suggestion
def _generate_copilot_environment_manifests(name, env_config, session):
```

simon106: Thanks updated as suggested will run pytest locally next time :)

gabelton (commented):
Could you update / add a unit tests that captures the fix here?

gabelton (approved):
It's not a blocker. The `get_aws_session_or_abort` is pretty thoroughly tested by itself

dbt_platform_helper/commands/environment.py
```diff
@@ -342,6 +342,7 @@ def generate(name, vpc_name):
         )
         raise click.Abort
 
+    session = get_aws_session_or_abort()
```
WillGibson, line 345: At risk of opening up a can of worms, because I've just realised we don't for any of the other places this method is used, should we have a test to cover this?

It could be just an assert that the method has been called and we rely on [the tests for the method](https://github.com/uktrade/platform-tools/blob/main/tests/platform_helper/utils/test_aws.py) to cover the detail.

PR #498, opened 2024-07-11
fix: DBPT-1128 Connection Error when connecting to Redis via Conduit

mramplin:
Addresses [DBPT-1128](https://uktrade.atlassian.net/browse/DBTP-1128).

Adds in the --execution-role switch to the copilot task run command, which utilises a custom IAM role (created via Terraform) to run the ECS tasks.

As this is a bug fix there are no changes to the existing tests; and no updates to the existing documentation

Regression tests have been completed [here](https://eu-west-2.console.aws.amazon.com/codesuite/codebuild/763451185160/projects/platform-tools-test/build/platform-tools-test%3Afe9f874a-7cc0-407b-80b5-d5c062f6277e/?region=eu-west-2)

---
### Title:
### Description:
### Tasks:

### DIFF
--- a/dbt_platform_helper/commands/conduit.py
+++ b/dbt_platform_helper/commands/conduit.py
@@ -17,6 +17,7 @@ from dbt_platform_helper.utils.versioning import (
     check_platform_helper_version_needs_update,
 )
 
+
 class ConduitError(Exception):
     pass
 
@@ -210,7 +211,7 @@ def create_addon_client_task(
 ):
     secret_name = f"/copilot/{app.name}/{env}/secrets/{normalise_secret_name(addon_name)}"
     session = app.environments[env].session
- 
+
     if addon_type == "postgres":
         if access == "read":
             secret_name += "_READ_ONLY_USER"
--- a/dbt_platform_helper/commands/conduit.py
+++ b/dbt_platform_helper/commands/conduit.py
@@ -225,6 +225,7 @@ def create_addon_client_task(
     subprocess.call(
         f"copilot task run --app {app.name} --env {env} "
         f"--task-group-name {task_name} "
+        f"--execution-role {app.name}-{addon_type}-{app.name}-{env}-ecsTask "
         f"--image {CONDUIT_DOCKER_IMAGE_LOCATION}:{addon_type} "
         f"--secrets CONNECTION_SECRET={get_connection_secret_arn(app, env, secret_name)} "
         "--platform-os linux "
--- a/dbt_platform_helper/commands/conduit.py
+++ b/dbt_platform_helper/commands/conduit.py
@@ -224,28 +224,25 @@ def create_addon_client_task(
         secret_name += "_ENDPOINT"
 
     try:
-        session.client("iam").get_role(RoleName=f"{app.name}-{addon_type}-{app.name}-{env}-ecsTask")
-
-        subprocess.call(
-            f"copilot task run --app {app.name} --env {env} "
-            f"--task-group-name {task_name} "
-            f"--execution-role {app.name}-{addon_type}-{app.name}-{env}-ecsTask "
-            f"--image {CONDUIT_DOCKER_IMAGE_LOCATION}:{addon_type} "
-            f"--secrets CONNECTION_SECRET={get_connection_secret_arn(app, env, secret_name)} "
-            "--platform-os linux "
-            "--platform-arch arm64",
-            shell=True,
+        session.client("iam").get_role(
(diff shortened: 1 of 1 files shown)


dbt_platform_helper/commands/conduit.py
```diff
@@ -222,15 +223,29 @@ def create_addon_client_task(
...
-        f"copilot task run --app {app.name} --env {env} "
-        f"--task-group-name {task_name} "
-        f"--image {CONDUIT_DOCKER_IMAGE_LOCATION}:{addon_type} "
-        f"--secrets CONNECTION_SECRET={get_connection_secret_arn(app, env, secret_name)} "
-        "--platform-os linux "
-        "--platform-arch arm64",
-        shell=True,
-    )
+    try:
+        session.client("iam").get_role(RoleName=f"{app.name}-{addon_type}-{app.name}-{env}-ecsTask")
```
samuele-mattiuzzo, line 227: if this is the line that's meant to raise the exception, then we could clean this up a bit to not repeat the `subprocess.call(...)` block?

we could do something like

```
try:
    session.client("iam").get_role(RoleName=f"{app.name}-{addon_type}-{app.name}-{env}-ecsTask")
    execution_role = "--execution-role {app.name}-{addon_type}-{app.name}-{env}-ecsTask "
except:
    execution_role = ""

suprocess.call(...
   f"{execution_role}"
)
```

samuele-mattiuzzo (commented):
a couple of comment, one inline, the other one is a question: has this been tested?

dbt_platform_helper/commands/conduit.py
```diff
@@ -222,15 +223,29 @@ def create_addon_client_task(
...
+    except:
+        subprocess.call(
+            f"copilot task run --app {app.name} --env {env} "
+            f"--task-group-name {task_name} "
+            f"--image {CONDUIT_DOCKER_IMAGE_LOCATION}:{addon_type} "
+            f"--secrets CONNECTION_SECRET={get_connection_secret_arn(app, env, secret_name)} "
+            "--platform-os linux "
+            "--platform-arch arm64",
+            shell=True,
+        )
```
WillGibson, line 248: Under what circumstance would we want to make the call without specifying the execution role?

mramplin: @WillGibson If/when the IAM role created by Terraform doesn't exist yet.

WillGibson: Isn't it already broken in that case? Or am I missing something?

samuele-mattiuzzo: @mramplin so that would make my comment a bit more useful? if the role hasn't been created then the IAM call would be the one raising the exception, correct?

also, probably worth making the `except` clause a bit more specific and have it catch the correct one

mramplin: @WillGibson it's already broken for Redis, but currently OpenSearch still works without that switch. I have also added the logic into the Terraform to enable the CMK with OpenSearch; so it will work with the execution role switch. Can we assume that the Terraform will be up to date? If we can then I can remove the check.

WillGibson: > Can we assume that the Terraform will be up to date?

No, don't remove it.

Thanks for explaining @mramplin.

I think @samuele-mattiuzzo has something with the duplication reduction idea though.

dbt_platform_helper/commands/conduit.py
antroy-madetech: Reminder: you'll need to change this if you action Will's comment in the other PR

dbt_platform_helper/commands/conduit.py
mramplin: Updates suggested by @samuele-mattiuzzo  have been made and regression [tests reran](https://eu-west-2.console.aws.amazon.com/codesuite/codebuild/763451185160/projects/platform-tools-test?region=eu-west-2&builds-meta=eyJmIjp7InRleHQiOiIifSwicyI6e30sIm4iOjIwLCJpIjowfQ#:~:text=platform%2Dtools%2Dtest%3Ab1e66874%2Df172%2D4ff9%2Dad4e%2D138bdcb7d858)

Thanks both for the feedback.

dbt_platform_helper/commands/conduit.py
mramplin: Thanks for the reminder. Both sides updated!

samuele-mattiuzzo (approved):
Just remember to update your branch with main @mramplin other than that lgtm 👍🏻

WillGibson (approved):

PR #501, opened 2024-07-12
fix: DBTP-1152 - slack notifications work when GitHub-Hookshot triggers build

ksugden:
Addresses [DBTP-1152](https://uktrade.atlassian.net/jira/software/c/projects/DBTP/boards/458?selectedIssue=DBTP-1152).

We use the commit hash to determine if the build commit is on the main branch.  This means that when a build is triggered from a detached HEAD state (as is the case with GitHub Hookshot triggers), we can still determine that it is a commit to main and that a notification should be sent to slack for build failures.

Manual tests were carried out from the command line locally and are not committed to the repo.  The change was also tested in Codebuild by temporarily updating the branch condition to match the PR branch and simulating an early exit 1 from the regression tests.  This failure successfully triggered a slack alert in the testing channel when the Codebuild was triggered with a commit sha override; and when it was triggered from the branch, providing assurance that the commit is being used to determine which branch is affected. 

---
### Title:
### Description:
### Tasks:

[DBTP-1152]: https://uktrade.atlassian.net/browse/DBTP-1152?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


buildspec.test.yml
```diff
@@ -26,7 +26,7 @@ phases:
   post_build:
     commands:
       - |
-        if [ "$(git rev-parse --abbrev-ref HEAD)" == "main" ] && [ "${CODEBUILD_BUILD_SUCCEEDING}" != "1" ]; then
+        if [ "${CODEBUILD_BUILD_SUCCEEDING}" != "1" ] && git branch --contains $CODEBUILD_RESOLVED_SOURCE_VERSION | grep -q "main"; then
```
WillGibson, line 29: I have a feeling this could evaluate as true if we ran the regression tests against something branched off `main` before adding any further commits. I am not sure why we would do that though, so maybe it's an edge case we do not need to worry about?

I would like to have the confidence that this has been tested in CodeBuild too. The pull request description suggests it hasn't. You should be able to change the branch to `DBTP-1152-slack-trigger` and use the `test-command-output` Slack channel to test and just add an `exit 1` to the top of the regression tests script to simulate an early failure.

ksugden: I did run the Codebuild job with my branch, but only for the non notification case, as the commit is not on main so the notification doesn't trigger.  Not sure if I can override the CODEBUILD_RESOLVED_SOURCE_VERSION env variable?

WillGibson: > Maybe we should only use the branch --contains logic in the specific case of the detached HEAD

I think it's probably such an edge case, it's probably worth mentioning the possibility on a comment in case it happens and someone comes looking, but not trying to remediate it in the code.

WillGibson: To test the failure scenario, I would...

* change the condition from `main` to your branch name in your branch
* set up a temporary pipeline that would build off your branch
* Test it from commit/push and by firing it off manually via the AWS Console

ksugden: As mentioned in the updated PR description, I have tested the alert scenario as suggested with temp commit [03809d](https://github.com/uktrade/platform-tools/pull/501/commits/03809d12353a2d8bf78922b2610fbc6e94ccadf7) and triggering the Codebuild job both from a branch override and from a commit sha (branch unaware).  I haven't had time to provision the test pipeline to test it from a pipeline triggered build.

WillGibson: Ah, cool, if you have done "commit sha (branch unaware)", I think that's covered it.

WillGibson (approved):

PR #504, opened 2024-07-16
feat: Add ability to override the terraform-platform-modules version in the platform-config.yml file.

antroy-madetech:
It can be overridden by the cli flag and defaults to a specific version otherwise

Addresses [DBTP-1133](https://uktrade.atlassian.net/browse/DBTP-1133)

[Documentation update](https://github.com/uktrade/platform-documentation/pull/338)

---
### Title:
### Description:
### Tasks:

[DBTP-1133]: https://uktrade.atlassian.net/browse/DBTP-1133?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dbt_platform_helper/commands/environment.py
```diff
@@ -399,18 +398,35 @@ def _generate_terraform_environment_manifests(
 ):
     env_template = setup_templates().get_template("environments/main.tf")
 
+    modules_version = _determine_terraform_platform_modules_version(
+        env_config, terraform_platform_modules_version
+    )
```
WillGibson, line 403: `modules_version` feels like it could be any kind of modules, I would prefer it to be named for what it is.

I can see how you got there though. Maybe this approach would be acceptable?

```suggestion
    terraform_platform_modules_version = _determine_terraform_platform_modules_version(
        env_config, terraform_platform_modules_version
    )
```

dbt_platform_helper/commands/environment.py
```diff
@@ -399,18 +398,35 @@ def _generate_terraform_environment_manifests(
...
+    cli_modules_version = terraform_platform_modules_version
+    env_conf_modules_version = env_conf.get("versions", {}).get("terraform-platform-modules")
+    version_preference_order = [
+        cli_modules_version,
+        env_conf_modules_version,
+        TERRAFORM_PLATFORM_MODULES_VERSION,
+    ]
+    modules_version = [version for version in version_preference_order if version][0]
+
+    return modules_version
```
WillGibson, line 427: Any reason not to just inline this?
```suggestion
    return [version for version in version_preference_order if version][0]
```

antroy-madetech: No, was probably a refactoring.

dbt_platform_helper/commands/environment.py
antroy-madetech: Renamed a load of these variables. Not a fan of reassigning variables as it can cause potential bugs, but hopefully this is OK.

WillGibson: I would agree normally, but in this case it felt better than any alternative I could come up with that didn't involve mile long variable names, and I'm quite comfortable with a long variable name.

WillGibson (approved):

PR #509, opened 2024-07-22
feat: DBTP-1154 add the prometheus prod service policy by default

lgarvey:
Addresses https://uktrade.atlassian.net/browse/DBTP-1154.

As we're rolling out prometheus/grafana metrics for every service, this PR causes the prometheus policy to be added to each service by fault.  

A platform documentation PR will arrive shortly.

---
### Title:
### Description:
### Tasks:


dbt_platform_helper/default-extensions.yml
```diff
@@ -17,3 +17,10 @@ monitoring:
   environments:
     "*":
       enable_ops_center: false
+
+prometheus-prod:
```
gabelton, line 21: Why is it called prod? Is there a non-prod policy coming in a later PR? Is there anything to stop this extension being deployed to a non-prod env?

lgarvey: Good question.  There's a prod and non prod grafana stack.  We're adding configuration for services to use prometheus prod by default, and if for some reason they want to push metrics to prometheus dev, they can add an explicit `prometheus-policy` extension with the dev assume role arn.

gabelton: I wonder if the demodjango toolspr env should connect up with prometheus dev, so that we can regression test the extension

lgarvey: So my plan was to wire up demodjango to prod prometheus, but we can point the toolspr env at dev if this makes more sense.

lgarvey: I've dropped the `-prod` bit

gabelton: If it's default for all services, then I think we should add a check to demodjango landing page that it's wired up correctly

lgarvey: Hmm, how would that work?

gabelton: Maybe it's not possible. I thought maybe it could hit a metrics endpoint or similar

lgarvey: Hmm, it's going to be tricky.  There is a metrics endpoint on the ADOT container, maybe demodjango could reverse scrape that and figure out if it is scraping metrics from the metadata endpoint.  

I'm going to create another ticket to look at this, so we can get this merged and moving forward.

gabelton (approved):

dbt_platform_helper/default-extensions.yml
gabelton: Sounds sensible. Just approved now

PR #516, opened 2024-07-29
feat: DBTP-1151 Add submodules script to pre_build.sh template

gabelton:
Part of https://uktrade.atlassian.net/browse/DBTP-1151

Please add any relevant context for you pull request here, or delete this if none needed.

---
### Title:
### Description:
- [n/a] If the work includes user interface changes, before and after screenshots included in description
### Tasks:
- [X] [Trigger the pull request regression tests for this branch](https://github.com/uktrade/platform-tools?tab=readme-ov-file#regression-tests) and confirm that they are passing


dbt_platform_helper/templates/.copilot/phases/pre_build.sh
```diff
@@ -3,4 +3,21 @@
 # Exit early if something goes wrong
 set -e
 
+if [ -f "./.gitmodules" ]; then
```
gabelton, line 6: Script based on current workaround used by lite-api https://github.com/uktrade/lite-api/blob/dev-migration/.copilot/phases/pre_build.sh

dbt_platform_helper/templates/pipelines/codebase/overrides/stack.ts
```diff
@@ -115,7 +116,7 @@ export class TransformedStack extends cdk.Stack {
                 location: `https://github.com/${this.codebaseConfiguration.repository}.git`,
                 gitCloneDepth: 0,
                 auth: {type: 'OAUTH'},
-                gitSubmodulesConfig: {fetchSubmodules: true},
+                gitSubmodulesConfig: {fetchSubmodules: false},
```
gabelton, line 119: When this is set to true, the build will fail at the `DOWNLOAD_SRC` phase before it reaches the pre-build script and we can edit the submodules file. Is there a scenario where we would need this to be true?

JohnStainsby: I don't think so, unless there was something in an install phase that used a public repo with submodules.

JohnStainsby (approved):

PR #523, opened 2024-08-02
fix: correct conduit log group name

gabelton:
I only needed to change filter name, not log group name, in [this ticket](https://github.com/uktrade/platform-tools/pull/520)


lgarvey (approved):
👍 Thanks!

PR #530, opened 2024-08-12
fix: DBTP-1207 - align extensions module name

ksugden:
Addresses [DBTP-1207](https://uktrade.atlassian.net/jira/software/c/projects/DBTP/boards/458?selectedIssue=DBTP-1207).

The terraform extensions module was initially generated with a suffix in its name, `extensions-tf`.  

When the terraform was subsequently regenerated by running the command,
`platform-helper environment generate-terraform --name <env>`

This caused an inconsistency in the generated code and the terraform state because the module name has been changed to `extensions` without a suffix in later terraform-platform-module versions.

We tested this change in the demo-django repo in our own environment.  We applied terraform with the `-tf` suffix on the extensions module name.  We regenerated the terraform using the new command which includes a moved block for the extensions module.  We applied the change and saw 'No changes' as expected.  We applied again (after the module name had already been updated) and the apply was also successful without changes.

---
### Title:
### Description:
### Tasks:

[DBTP-1207]: https://uktrade.atlassian.net/browse/DBTP-1207?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dbt_platform_helper/templates/environments/main.tf
+++ b/dbt_platform_helper/templates/environments/main.tf
@@ -37,6 +37,10 @@ module "extensions" {
   vpc_name    = "{{ config.vpc }}"
 }
 
+{% comment %} 
+Clean up because terraform modules were initially deployed with a -tf suffix.  This block moves those modules to naming without a suffix.
+Can be removed once all services have moved to the new naming.
+{% endcomment %}
 moved {
   from = module.extensions-tf
   to   = module.extensions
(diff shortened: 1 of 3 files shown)


gabelton:
Think I'd expect to see this unit test fail https://github.com/uktrade/platform-tools/blob/main/tests/platform_helper/test_command_environment.py#L522

Could you add a little assertion? 

Also just to double check this isn't a breaking change is it? The idea is that we'll gradually move over to the new naming system and this changes supports us in the meantime. Right?

PR #538, opened 2024-08-20
fix: Replace .platform-helper-version check in buildspec

antroy-madetech:
Addresses [DBTP-1153](https://uktrade.atlassian.net/browse/DBTP-1153)

Buildspec changes to support the move away from .platform-helper-version.

Also makes the `platform-helper version get-platform-helper-for-project` command fail if the version is not in the platform-config.yml and there isn't a .platform-helper-version file with a version to make it easier to use in pipelines.

---
### Title:
### Description:
### Tasks:

[DBTP-1153]: https://uktrade.atlassian.net/browse/DBTP-1153?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dbt_platform_helper/templates/pipelines/environments/buildspec.yml
```diff
@@ -14,29 +14,31 @@ phases:
           echo "Cannot find .copilot-version file"
           exit 1
         fi
+      - COPILOT_VERSION=$(cat .copilot-version)
+      # Install pyyaml and dbt-platform-helper
+      - pip install PyYAML dbt-platform-helper
```
gabelton, line 19: Isn't PyYAML already installed with dbt-platform-helper?

gabelton (approved):

dbt_platform_helper/templates/pipelines/environments/buildspec.yml
gabelton: Hadn't noticed this gets used in post-build. Happy to leave it in there

dbt_platform_helper/utils/versioning.py
```diff
@@ -245,7 +245,9 @@ def check_platform_helper_version_mismatch():
```
antroy-madetech, line 245: @WillGibson Do you know why this check is here? It seems to work fine when running locally with this check commented out. It means there's some functionality which can't be tested manually before releasing which isn't ideal.

gabelton (approved):

dbt_platform_helper/utils/versioning.py
WillGibson: It is to stop it messing up when you are using a locally built version of `platform-helper`, not one installed from PyPi.

E.g. in the regression tests.

antroy-madetech: I added a test around this check.

PR #542, opened 2024-08-27
fix: DBTP-1323 Broken `platform-helper codebase \*` commands

antroy-madetech:
A bug introduced in 10.7.2 meant that the `codebase` commands were failing because it could not find a `platform-config.yml` file. The command is used in the codebase repo, not the deploy repo, so it shouldn't be looking for, or failing because of this file.

Addresses [DBTP-1323](https://uktrade.atlassian.net/browse/DBTP-1323)

---
### Title:
### Description:
### Tasks:

[DBTP-1323]: https://uktrade.atlassian.net/browse/DBTP-1323?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dbt_platform_helper/utils/versioning.py
+++ b/dbt_platform_helper/utils/versioning.py
@@ -104,7 +104,7 @@ def get_github_released_version(repository: str, tags: bool = False) -> Tuple[in
     return parse_version(package_info["tag_name"])
 
 
-def get_platform_helper_versions(ignore_platform_config_versions=False) -> PlatformHelperVersions:
+def get_platform_helper_versions(local_and_latest_only=False) -> PlatformHelperVersions:
     try:
         locally_installed_version = parse_version(version("dbt-platform-helper"))
     except PackageNotFoundError:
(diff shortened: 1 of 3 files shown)


dbt_platform_helper/utils/versioning.py
```diff
@@ -104,7 +104,7 @@ def get_github_released_version(repository: str, tags: bool = False) -> Tuple[in
     return parse_version(package_info["tag_name"])
 
 
-def get_platform_helper_versions() -> PlatformHelperVersions:
+def get_platform_helper_versions(local_and_latest_only=False) -> PlatformHelperVersions:
```
WillGibson, line 107: I wonder whether this could be named to better express the intent? Maybe something like
```suggestion
def get_platform_helper_versions(include_desired_versions=False) -> PlatformHelperVersions:
```

antroy-madetech: I've changed the name about half a dozen times :-D
Local and latest only does express the intent, which is to get just the local and latest versions. include_desired_versions seems weird as when false it reads "I don't want my desired versions"

antroy-madetech: include_required_versions?

antroy-madetech: for_local_version_check?

antroy-madetech: include_project_versions?

WillGibson: That last one is good. Go with that.

PR #555, opened 2024-09-10
feat: DBTP 1162 support hosting static sites on s3

gabelton:
Addresses [DBTP-1162](https://uktrade.atlassian.net/browse/DBTP-1162)

---
### Title:
### Description:
### Tasks:

[DBTP-1162]: https://uktrade.atlassian.net/browse/DBTP-1162?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dbt_platform_helper/commands/copilot.py
```diff
@@ -313,6 +316,8 @@ def make_addons():
         log_destination_arns = get_log_destination_arn()
 
         if addon_type in ["s3", "s3-policy"]:
+            if config[addon_name].get("serve_static"):
```
gabelton, line 319: We don't need svc or env addons, for a static bucket because we don't need the svc addon policy. I'm not adding logic for digital workspace because we're migrating them off copilot and they don't require this functionality, so I think it's wasted effort

dbt_platform_helper/custom_resources/s3_object.py
```diff
@@ -72,10 +72,12 @@ def handler(event, context):
                 Key=properties["S3ObjectKey"],
             )
         else:
+            content_type = properties.get("S3ObjectContentType", "binary/octet-stream")
```
gabelton, line 75: "binary/octet-stream" is the AWS default when adding an object to bucket

dbt_platform_helper/utils/validation.py
```diff
@@ -328,12 +329,7 @@ def iam_role_arn_regex(key):
 S3_DEFINITION.update(
     {
         "type": "s3",
-        Optional("objects"): [
-            {
-                "key": str,
-                Optional("body"): str,
-            }
-        ],
+        Optional("objects"): [{"key": str, Optional("body"): str, Optional("content_type"): str}],
```
gabelton, line 332: Unless content type is specified, the objects default to "binary/octet-stream". This means that browsers will download an html file rather displaying in browser

WillGibson: Worth mentioning this in the documentation part.

WillGibson (commented):
Is this included in our regression tests via demodjango?

gabelton:
> Is this included in our regression tests via demodjango?

It will be because I've added a check to grab some static content, so it will get run as part of the smoke tests.

WillGibson (approved):

PR #560, opened 2024-09-12
fix: remove version update check from version group commands

gabelton:
[This line](https://github.com/uktrade/terraform-platform-modules/blob/d67710ec1bc72ed7b756b9fa72770a7638e5361a/environment-pipelines/buildspec-install-build-tools.yml#L20) in env pipeline fails because, instead of a version number,`REQUIRED_VERSION` is assigned the output of the version update check i.e.

```
import re, sys; print('tag' if re.match(r'\d+\.\d+\.\d+', 'You are running platform-helper v10.9.1, upgrade to v10.10.0 by running run `pip install --upgrade dbt-platform-helper`.
```

### DIFF
--- a/dbt_platform_helper/commands/version.py
+++ b/dbt_platform_helper/commands/version.py
@@ -2,9 +2,6 @@ import click
 
 from dbt_platform_helper.utils.click import ClickDocOptGroup
 from dbt_platform_helper.utils.platform_config import get_environment_pipeline_names
-from dbt_platform_helper.utils.versioning import (
-    check_platform_helper_version_needs_update,
-)
 from dbt_platform_helper.utils.versioning import get_required_platform_helper_version
 
 
@@ -12,7 +9,6 @@ from dbt_platform_helper.utils.versioning import get_required_platform_helper_ve
 def version():
     """Contains subcommands for getting version information about the current
     project."""
-    check_platform_helper_version_needs_update()
 
 
 @version.command(help="Print the version of platform-tools required by the current project")


dbt_platform_helper/commands/version.py
```diff
@@ -2,17 +2,13 @@
...
-    check_platform_helper_version_needs_update,
-)
 from dbt_platform_helper.utils.versioning import get_required_platform_helper_version
 
 
 @click.group(chain=True, cls=ClickDocOptGroup)
 def version():
     """Contains subcommands for getting version information about the current
     project."""
-    check_platform_helper_version_needs_update()
```
WillGibson, line 15: I kind of want to know why this check was here before we remove it.

Another way of addressing this could be to change...

https://github.com/uktrade/terraform-platform-modules/blob/d67710ec1bc72ed7b756b9fa72770a7638e5361a/environment-pipelines/buildspec-install-build-tools.yml#L19

...to...

```
REQUIRED_VERSION=$(PLATFORM_TOOLS_SKIP_VERSION_CHECK=true platform-helper version get-platform-helper-for-project --pipeline "${PIPELINE_NAME}")
```

gabelton: Thinking about it with fresher brain this morning. In this context I think we do want to make sure that we always have the latest released version of the package, so that we have the latest `get_platform_helper_for_project` logic. I don't think we want to skip the check by removing it or adding the skip line then.

Trying to work out how this scenario hasn't happened before. I merged 10.10.0 release PR yesterday afternoon, which has kicked off the regression tests. The PYPI release job finished at 3.51, while the regression tests failed at 3.52. Could it be that in the time it's taken the env pipeline install command to finish building `platform-helper` the release has been deployed and so it's out of date? Or there's some kind of caching of the earlier version going on. I think the former seems more likely given the closeness in time of the two codebuild jobs finishing.

gabelton:
I've rerun the regression tests now and they've passed. So I think this is an edge case and the solution definitely isn't to skip the package version check. Closing this and will consider creating a bug ticket around how we sequence running regression tests and creating a release.

dbt_platform_helper/commands/version.py
antroy-madetech: Sounds like the former agreed. The times are very close. I presume the issue is that the warning is getting picked up in the output of the command and so isn't a clean version number. 

If you decide to go the route of skipping that check, I think a flag would be better than an env var - otherwise it gets very hard to work out what things should be configured as options and which should be env vars. If that var is already there it would be better to deprecate it in favour of a flag if possible.

Honestly though, the command is used to determine which version should be installed, so I think the additional check makes little sense. I think I put it in there to be consistent with the other commands.

gabelton: >  I presume the issue is that the warning is getting picked up in the output of the command and so isn't a clean version number.
> 
Exactly.

I think ultimately the solution would be to make sure that regression tests run against the release PR before it gets merged to main and then we wouldn't have to be running the regression tests at the same time as the pypi release job

PR #575, opened 2024-09-19
fix: DBTP-1375 - platform helper version get-platform-helper-for-project no longer validates config

ksugden:
Addresses [DBTP-1375](https://uktrade.atlassian.net/browse/DBTP-1375)

When the `platform-helper version get-platform-helper-for-project` command was called, config validation was being done twice.  First, in the command decorators and again while executing the command.  These validations fail if the configuration found in `platform-config.yml` is for a different version of platform-helper than that currently installed, resulting in errors.

This PR fixes the errors by making the following changes:

- add functions in `platform_config.py` to read and load the `platform-config.yml` file, ignoring yaml and configuration errors
- new unit tests for the new load and read functions
- `get_environment_pipeline_names` uses the new function so no longer validates config
- `get_platform_helper_for_project` uses the new function so no longer validates config
- `get_platform_helper_for_project` and `version` functions are moved into a new class, `VersionCommand`.  This is so that a new class instance can be created in tests, which re-evaluates the click decorators based on the current state of the filesystem.
- new unit tests covering the version command given invalid config, and config with invalid yaml in `platform-config.yml`

---
### Title:
### Description:
### Tasks:

[DBTP-1375]: https://uktrade.atlassian.net/browse/DBTP-1375?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dbt_platform_helper/utils/platform_config.py
```diff
@@ -1,17 +1,32 @@
...
+
+
+def load_config_file():
+    file_contents = read_config_file()
+    if not file_contents:
+        return None
+    try:
+        return yaml.safe_load(file_contents)
+    except yaml.parser.ParserError:
+        return None
```
antroy-madetech, line 20: I think if you return an empty dict rather than None, there will be less need for error handling.

dbt_platform_helper/utils/platform_config.py
```diff
@@ -1,17 +1,32 @@
...
-    if not Path(PLATFORM_CONFIG_FILE).exists():
-        return {}
+    config = load_config_file()
 
-    config = load_and_validate_platform_config(disable_aws_validation=True, disable_file_check=True)
-    return config.get("environment_pipelines", {}).keys()
+    try:
+        return config.get("environment_pipelines", {}).keys()
+    except AttributeError:
+        return {}
```
antroy-madetech, line 29: e.g. the error handling will be unnecessary here if load_config_file returns an empty dict.

dbt_platform_helper/utils/platform_config.py
```diff
@@ -1,17 +1,32 @@
...
 from dbt_platform_helper.constants import PLATFORM_CONFIG_FILE
-from dbt_platform_helper.utils.validation import load_and_validate_platform_config
+
+
+def read_config_file():
+    if Path(PLATFORM_CONFIG_FILE).exists():
+        return Path(PLATFORM_CONFIG_FILE).read_text()
+
+
+def load_config_file():
```
antroy-madetech, line 13: Probably worth naming this to reflect that it ignores errors by design. To stop people adding validation in the future or using it where they should be using the load_and_validate_platform_config function.

dbt_platform_helper/utils/versioning.py
```diff
@@ -104,22 +104,40 @@ def get_github_released_version(repository: str, tags: bool = False) -> Tuple[in
     return parse_version(package_info["tag_name"])
 
 
+def get_latest_release():
```
antroy-madetech, line 107: If this was extracted and has no tests for itself, it should probably be prefixed with "_" to indicate it's private.

dbt_platform_helper/commands/version.py
```diff
@@ -2,37 +2,39 @@
...
-def version():
-    """Contains subcommands for getting version information about the current
-    project."""
-    check_platform_helper_version_needs_update()
+class VersionCommand:
+    def __init__(self):
+        self.command_group = self.version
+        self.command = self.get_platform_helper_for_project
 
+    _ENVIRONMENT_PIPELINE_NAMES = get_environment_pipeline_names()
```
WillGibson, line 13: Is this a constant, but it's initialised when exactly?

Maybe ` self._environment_pipeline_names = get_environment_pipeline_names()` in the init method would be more intuitive?

tests/platform_helper/utils/test_versioning.py
```diff
@@ -278,6 +278,53 @@ def test_get_platform_helper_versions(
     assert not mock_aws.called  # Ensure that an AWS session is not required for this function
 
 
+@patch("requests.get")
+@patch("dbt_platform_helper.utils.versioning.version")
+@patch("dbt_platform_helper.utils.platform_config.read_config_file")
+def test_get_platform_helper_versions_with_invalid_yaml_in_platform_config(
+    mock_config_file_contents, mock_version, mock_get, fakefs
```
WillGibson, line 285: Can we name `mock_version` and `mock_get` to make it clearer what they are?

WillGibson (commented):
Partial pass

dbt_platform_helper/utils/versioning.py
ksugden: I have actually made this function more robust and added tests in another branch.  I was worried the scope of this PR was getting a bit large, but I could pull those changes in if you don't mind?

ksugden: Maybe it should be part of an independent piece of work because I'm not sure what we really want to return in the requests error cases at the moment! But feel free to take a look and add your thoughts.  I'm happy to move get_latest_release to a private function for now.

dbt_platform_helper/commands/version.py
antroy-madetech: I think that would be too late for the decorators to pick it up @WillGibson . Worth testing though.

dbt_platform_helper/utils/versioning.py
antroy-madetech: Make it private for now, and address the error and how to handle it in the other ticket.

dbt_platform_helper/commands/version.py
ksugden: It's instantiated when the decorators are applied to the class functions, which is when the class is instantiated.  @antroy-madetech is right.  It's not possible to use it in the decorator if it is a class attribute as the attributes are set after the decorators are applied (I think!! confirmed by test).  If you prefer, I could ditch the const declaration and just call it directly inside the decorator?  I've done that, so see if you prefer, and I can revert if not.

antroy-madetech (approved):

WillGibson:
@ksugden / @antroy-madetech I haven't had time/headspace to do a decent review on this. I trust you to to have it covered between you.

PR #578, opened 2024-09-20
fix: ensure delete_listener_rule deletes multiple rules with same name

gabelton:
There can be multiple listener rules with the same name. When we run `online`, we want to delete all of these. Currently only one of each name will be deleted

### DIFF
--- a/dbt_platform_helper/commands/environment.py
+++ b/dbt_platform_helper/commands/environment.py
@@ -483,11 +483,11 @@ def delete_listener_rule(tag_descriptions: list, tag_name: str, lb_client: boto3
         tags = {t["Key"]: t["Value"] for t in description["Tags"]}
         if tags.get("name") == tag_name:
             current_rule_arn = description["ResourceArn"]
+            if current_rule_arn:
+                lb_client.delete_rule(RuleArn=current_rule_arn)
 
     if not current_rule_arn:
         return current_rule_arn
-
-    lb_client.delete_rule(RuleArn=current_rule_arn)
 
     return current_rule_arn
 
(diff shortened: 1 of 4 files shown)


dbt_platform_helper/commands/environment.py
```diff
@@ -483,12 +483,12 @@ def delete_listener_rule(tag_descriptions: list, tag_name: str, lb_client: boto3
         tags = {t["Key"]: t["Value"] for t in description["Tags"]}
         if tags.get("name") == tag_name:
             current_rule_arn = description["ResourceArn"]
+            if current_rule_arn:
+                lb_client.delete_rule(RuleArn=current_rule_arn)
 
     if not current_rule_arn:
```
ksugden, line 489: This bit looks a bit weird now - 
```if not x: 
       return x
return x```
Couldn't it just be `return x`?

File: regression_tests/stages/run_maintenance_page_tests.sh [hunk 2]
```diff
@@ -14,7 +14,7 @@ cd "${CODEBUILD_SRC_DIR}/demodjango"
 ./smoke_tests.sh ${TARGET_ENVIRONMENT} smoke
 
 echo -e "\nRunning offline command"
-OUTPUT=$(echo "y" | AWS_PROFILE=platform-sandbox PLATFORM_TOOLS_SKIP_VERSION_CHECK=true platform-helper environment offline --app demodjango --env "${TARGET_ENVIRONMENT}" --vpc platform-sandbox-dev)
+OUTPUT=$(echo "y" | AWS_PROFILE=platform-sandbox PLATFORM_TOOLS_SKIP_VERSION_CHECK=true platform-helper environment offline --app demodjango --env "${TARGET_ENVIRONMENT}" --svc "*" --vpc platform-sandbox-dev)
```
gabelton, line 17: Otherwise the listener rules don't allow access to the API service when under maintenance

DeveloperConnor (approved):

PR #607, opened 2024-10-24
feat: DBTP-1215 Improve error message when AWS profile not set

chiaramapellimt:
Addresses [DBTP-1215](https://uktrade.atlassian.net/browse/DBTP-1215)

We have quite a few errors which can be thrown when there's an issue with the aws_profile, so instead of adding yet another one to the pile I extracted them into a parametarised test and the errors are now clear in the implementation code, and most importantly they are all tested individually

---
### Title:
### Description:
### Tasks:

[DBTP-1215]: https://uktrade.atlassian.net/browse/DBTP-1215?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dbt_platform_helper/utils/aws.py
+++ b/dbt_platform_helper/utils/aws.py
@@ -21,17 +21,20 @@ AWS_SESSION_CACHE = {}
 
 
 def get_aws_session_or_abort(aws_profile: str = None) -> boto3.session.Session:
-    aws_profile = aws_profile if aws_profile else os.getenv("AWS_PROFILE")
+    aws_profile = aws_profile or os.getenv("AWS_PROFILE")
     if aws_profile in AWS_SESSION_CACHE:
         return AWS_SESSION_CACHE[aws_profile]
 
-    # Check that the aws profile exists and is set.
-    click.secho(f"""Checking AWS connection for profile "{aws_profile}"...""", fg="cyan")
+    click.secho(f'Checking AWS connection for profile "{aws_profile}"...', fg="cyan")
 
     try:
         session = boto3.session.Session(profile_name=aws_profile)
+        sts = session.client("sts")
+        account_id, user_id = get_account_details(sts)
+        click.secho("Credentials are valid.", fg="green")
+
     except botocore.exceptions.ProfileNotFound:
-        click.secho(f"""AWS profile "{aws_profile}" is not configured.""", fg="red")
+        click.secho(f'AWS profile "{aws_profile}" is not configured.', fg="red")
         exit(1)
     except botocore.exceptions.ClientError as e:
         if e.response["Error"]["Code"] == "ExpiredToken":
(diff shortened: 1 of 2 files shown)


dbt_platform_helper/utils/aws.py
```diff
@@ -48,9 +49,10 @@ def get_aws_session_or_abort(aws_profile: str = None) -> boto3.session.Session:
         botocore.exceptions.UnauthorizedSSOTokenError,
         botocore.exceptions.TokenRetrievalError,
         botocore.exceptions.SSOTokenLoadError,
+        botocore.exceptions.NoCredentialsError,
     ):
         click.secho(
-            "The SSO session associated with this profile has expired or is otherwise invalid."
+            "The SSO session associated with this profile has expired, is not set or is otherwise invalid."
```
WillGibson, line 55: I think the issue here is that the AWS Profile is either not specified, or doesn't exist, not that the SSO session is expired or invalid.

I wonder if we could catch this separately an output a different error message?

chiaramapellimt: Indeed. We trying adding a block towards the beginning of the method, so that it would be a showstopper. I am talking about putting this here, on this [Line](https://github.com/uktrade/platform-tools/blob/449fb0c050b5ec879743f1ad747e8ffc97108171/dbt_platform_helper/utils/aws.py#L28).
The problem is that across the unit tests, we abuse patching, which is a minefield for this sort of change.

The idea was to add a "if not aws_profile" --> showstopper / exit(1).

But the way we build our unit tests actually prevents us from doing this with ease. I spent a lot of time navigating patches and mocks, I tried adding the above functionality in a @mock fixture, to no avail. Connor and I spent ages looking at this. 

I think it's something we need to address ASAP Will. But this looked like the better choice all things considering. Testing it on the console after having uset all the direnv gave us the expected error message, too.

tests/platform_helper/utils/test_aws.py
```diff
@@ -99,7 +99,7 @@ def test_get_aws_session_or_abort_with_invalid_credentials(
 ):
     aws_profile = "existing_profile"
     expected_error_message = (
-        "The SSO session associated with this profile has expired or is otherwise invalid."
+        "The SSO session associated with this profile has expired, is not set or is otherwise invalid."
         + "To refresh this SSO session run `aws sso login` with the corresponding profile"
     )
     mock_get_account_details.side_effect = botocore.exceptions.SSOTokenLoadError(
```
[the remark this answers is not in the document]
chiaramapellimt: I am happy to differentiate the four error messages, just want to ensure we're on the same page regarding the functionality first (meaning adding the other exception instead of showstopping the no "_aws_profile_" at the start).

dbt_platform_helper/utils/aws.py
```diff
@@ -20,62 +21,76 @@
...
+        )
+    except botocore.exceptions.UnauthorizedSSOTokenError:
+        _handle_error(
+            "The SSO Token used for this session is unauthorised."
+            "To refresh this SSO session run `aws sso login` with the corresponding profile"
+        )
+    except botocore.exceptions.TokenRetrievalError:
+        _handle_error(
+            "Unable to retrieve the Token for this session."
+            "To refresh this SSO session run `aws sso login` with the corresponding profile"
```
WillGibson, line 56: Might be worth extracting this string to a variable as it's repeated three times.

WillGibson (commented):
Looking good. One very small refactor suggestion.

dbt_platform_helper/utils/aws.py
```diff
@@ -18,6 +18,9 @@
 SSM_BASE_PATH = "/copilot/{app}/{env}/secrets/"
 SSM_PATH = "/copilot/{app}/{env}/secrets/{name}"
 AWS_SESSION_CACHE = {}
+REFRESH_TOKEN_MESSAGE = (
+    "To refresh this SSO session run `aws sso login` with the corresponding profile"
+)
```
WillGibson, line 23: I would be inclined to declare this in `get_aws_session_or_abort` so it's scoped to that method.

The tests can remain as strings, not using this, otherwise you could splurge a typo into here and the test would still pass.

chiaramapellimt: I rescoped the var, was this what you meant RE strings in tests?

WillGibson: The test could have its own variable to avoid duplication, but it should not be the one in the production code. Up to you whether you do that last little optional refactor. I am approving.

WillGibson (approved):

PR #625, opened 2024-11-08
fix: DBTP-1548 - Maintenance page listener rules can be based on both CIDR range or IP in EGRESS_IP ssm parameter

ksugden:
Addresses [DBTP-1548](https://uktrade.atlassian.net/browse/DBTP-1548)

We should allow CIDR ranges as well as IP addresses in the EGRESS_IPS ssm parameter.  This means we need to change how we create source IP rules for our maintenance page as we do not necessarily need to add a prefix length to the IP addresses retrieved from SSM.  This change adds a default cidr prefix to IPs without the prefix but leaves IPs with a CIDR range unchanged.

This is a small change, which doesn't introduce any validations.  This will be addressed in a follow up ticket - DBTP-1553, where we will better handle errors encountered in the maintenance_page commands.

---
### Title:
### Description:
### Tasks:

[DBTP-1548]: https://uktrade.atlassian.net/browse/DBTP-1548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dbt_platform_helper/domain/maintenance_page.py
```diff
@@ -387,6 +387,12 @@ def create_header_rule(
     )
 
 
+def add_default_cidr_prefix(ip_network: str, default_cidr_prefix_length: int):
+    if "/" in ip_network:
+        return ip_network
+    return ip_network + "/" + str(default_cidr_prefix_length)
```
WillGibson, line 393: The tests suggest that there is not a default value.

```suggestion
def ensure_cidr_suffix(ip_network: str, cidr_suffix: int):
    if "/" in ip_network:
        return ip_network
     
    return ip_network + "/" + str(cidr_suffix)
```

ksugden: I've removed the prefix parameter (we should never need to default it to anything other than 32) and renamed, hopefully making intention clearer.

WillGibson (approved):

PR #652, opened 2024-11-21
feat: DBTP-1568 - Add s3 support for external role access

ksugden:
Addresses [DBTP-1568](https://uktrade.atlassian.net/browse/DBTP-<ticket>) adding validation and schema for s3 external role access configuration block.  Feature is documented here:
https://github.com/uktrade/platform-documentation/pull/422
---
### Title:
### Description:
### Tasks:

[DBTP-1568]: https://uktrade.atlassian.net/browse/DBTP-1568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JohnStainsby (approved):

dbt_platform_helper/utils/validation.py
```diff
@@ -266,6 +266,25 @@ def iam_role_arn_regex(key):
     )
 
 
+def dbt_email_address_regex(key):
+    return Regex(
+        r"^[\w\-\.]+@(businessandtrade.gov.uk|digital.trade.gov.uk)$",
```
antroy-madetech, line 271: I don't think you need to escape the . and hyphen in the character group, but the hyphen will probably need to come last to stop the regex parser from thinking it's a character range.
```suggestion
        r"^[\w.-]+@(businessandtrade.gov.uk|digital.trade.gov.uk)$",
```

Are we sure that these are the only characters used in DBT email addresses?

dbt_platform_helper/utils/validation.py
```diff
@@ -266,6 +266,25 @@ def iam_role_arn_regex(key):
...
+
+EXTERNAL_ROLE_ACCESS = {
+    "role_arn": iam_role_arn_regex("role_arn"),
+    "read": bool,
+    "write": bool,
+    "cyber_sign_off_by": dbt_email_address_regex("cyber_sign_off_by"),
+}
+
+EXTERNAL_ROLE_ACCESS_NAME = Regex(
+    r"^([a-z][a-zA-Z0-9_-]*|\*)$",
```
antroy-madetech, line 284: This allows for the external role access name to be just `*`?

dbt_platform_helper/utils/validation.py
```diff
@@ -266,6 +266,25 @@ def iam_role_arn_regex(key):
...
+    error="External role access block name {} is invalid: names must only contain lowercase alphanumeric characters separated by hypen or underscore",
```
antroy-madetech, line 285: In which case we should allow for that in the error message.

antroy-madetech (commented):
Initial comments - will probably fix some of these as I start on the next stuff.

dbt_platform_helper/utils/validation.py
ejayesh: yes, we only use these two.  Eveyones old address have been moved over to businessandtrade

dbt_platform_helper/utils/validation.py
ksugden: that was a copy-paste error from ENV regex.  I've removed the * option

PR #660, opened 2024-11-25
fix: DBTP-1572 - Fix _validate_exension_supported_versions incorrectly raising an error when no version is supplied

DeveloperConnor:
Addresses [DBTP-<ticket>](https://uktrade.atlassian.net/browse/DBTP-<ticket>)

_validate_extension_supported_versions() would throw a warning when no engine version is provided, even if a version is provided under '*' (which would apply to all environments)

Unit tests updated with new scenarios.
---
### Title:
### Description:
### Tasks:

E2E Pipeline - https://eu-west-2.console.aws.amazon.com/codesuite/codepipeline/pipelines/pull-request-end-to-end-tests/executions/3ff57595-550f-4540-939b-bc6b0ad6b0a4/visualization?region=eu-west-2

### DIFF
--- a/dbt_platform_helper/utils/validation.py
+++ b/dbt_platform_helper/utils/validation.py
@@ -266,6 +266,25 @@ def iam_role_arn_regex(key):
     )
 
 
+def dbt_email_address_regex(key):
+    return Regex(
+        r"^[\w.-]+@(businessandtrade.gov.uk|digital.trade.gov.uk)$",
+        error=f"{key} must contain a valid DBT email address",
+    )
+
+
+EXTERNAL_ROLE_ACCESS = {
+    "role_arn": iam_role_arn_regex("role_arn"),
+    "read": bool,
+    "write": bool,
+    "cyber_sign_off_by": dbt_email_address_regex("cyber_sign_off_by"),
+}
+
+EXTERNAL_ROLE_ACCESS_NAME = Regex(
+    r"^([a-z][a-zA-Z0-9_-]*)$",
+    error="External role access block name {} is invalid: names must only contain lowercase alphanumeric characters separated by hypen or underscore",
+)
+
 DATA_IMPORT = {
     Optional("source_kms_key_arn"): kms_key_arn_regex("source_kms_key_arn"),
     "source_bucket_arn": s3_bucket_arn_regex("source_bucket_arn"),
@@ -288,7 +307,8 @@ S3_BASE = {
             Optional("versioning"): bool,
             Optional("lifecycle_rules"): [LIFECYCLE_RULE],
             Optional("data_migration"): DATA_MIGRATION,
-        }
+            Optional("external_role_access"): {EXTERNAL_ROLE_ACCESS_NAME: EXTERNAL_ROLE_ACCESS},
+        },
     },
 }
 
--- a/tests/platform_helper/utils/test_validation.py
+++ b/tests/platform_helper/utils/test_validation.py
@@ -107,6 +107,8 @@ def test_validate_addons_success(addons_file):
                 "my-s3-bucket-data-migration-source-bucket-invalid-arn": r"source_bucket_arn must contain a valid ARN for an S3 bucket",
                 "my-s3-bucket-data-migration-source-kms-key-invalid-arn": r"source_kms_key_arn must contain a valid ARN for a KMS key",
                 "my-s3-bucket-data-migration-worker-role-invalid-arn": r"worker_role_arn must contain a valid ARN for an IAM role",
+                "my-s3-external-access-bucket-invalid-arn": r"role_arn must contain a valid ARN for an IAM role",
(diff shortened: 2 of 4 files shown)


dbt_platform_helper/utils/validation.py
```diff
@@ -569,13 +569,16 @@ def _validate_extension_supported_versions(
...
                 fg="red",
             )
             continue
         for environment, env_config in environments.items():
+
+            # An extension version doesn't need to be specified for all environments, provided one is specified under "*".
+            # So check if the version is set before checking if it's supported
             extension_version = env_config.get(version_key)
-            if extension_version not in supported_extension_versions:
+            if extension_version and extension_version not in supported_extension_versions:
```
WillGibson, line 581: I think this might let "no version" slip through.

Can we add test to verify the correction validation behaviour occurs when:

* there is no version for the default or specific environment
* there is no version for the default, but it is specified for the specific environment

Then perhaps the code here needs to work out what the version is for each environment, falling back to the default (`"*"`) as required, before validating it.

DeveloperConnor: I thought about adding code to check that a engine is specified for each environment, or there is an engine defined at the * level. 

That sort of validation should be delegated to the platform config schema no? This method is looking only that if the version is specified that it falls within supported versions.

Will add a test for above - just to clarify:

- there is no version for the default or specific environment
Would cause no warnings to be raised as no version specified
- there is no version for the default, but it is specified for the specific environment
Would cause no warning to be raised unless the version set for a specific environment is outside supported versions.

tests/platform_helper/utils/test_validation.py
```diff
@@ -947,7 +947,10 @@ def test_validate_extensions_supported_versions_successful_with_supported_versio
         "application": "test-app",
         "environments": {"dev": {}, "test": {}, "prod": {}},
         "extensions": {
-            "connors-redis": {"type": "redis", "environments": {"*": {"engine": "7.1"}}}
+            "connors-redis": {
+                "type": "redis",
+                "environments": {"*": {"engine": "7.1", "plan": "tiny"}, "prod": {"plan": "tiny"}},
+            }
```
WillGibson, line 953: I think it's worth paramaterising this test so we can run the previous and the new example.

dbt_platform_helper/utils/validation.py
WillGibson: Cool.

tests/platform_helper/utils/test_validation.py
DeveloperConnor: Agreed - done.

dbt_platform_helper/utils/validation.py
```diff
@@ -589,13 +589,16 @@ def _validate_extension_supported_versions(
 
         if not isinstance(environments, dict):
             click.secho(
-                "Error: Opensearch extension definition is invalid type, expected dictionary",
+                f"Error: {extension_type} extension definition is invalid type, expected dictionary",
```
WillGibson, line 592: Good catch.

tests/platform_helper/utils/test_validation.py
```diff
@@ -940,58 +940,66 @@ def test_validate_database_copy_multi_postgres_failures(capfd):
...
+        ),
+        (
+            # Invalid engine defined in prod environment
+            {
+                "extensions": {
+                    "connors-redis": {
+                        "type": "redis",
+                        "environments": {
+                            "*": {"plan": "tiny"},
+                            "prod": {"engine": "6.1", "plan": "tiny"},
```
WillGibson, line 981: 6.1 looks like it could be OK, this could make it a bit clearer...
```suggestion
                            "prod": {"engine": "invalid", "plan": "tiny"},
```

tests/platform_helper/utils/test_validation.py
```diff
@@ -940,58 +940,66 @@ def test_validate_database_copy_multi_postgres_failures(capfd):
...
+                        },
+                    }
+                },
+            },
+            "redis version for environment prod is not in the list of supported redis versions: ['7.1']. Provided Version: 6.1",
```
WillGibson, line 986:
```suggestion
            "redis version for environment prod is not in the list of supported redis versions: ['7.1']. Provided Version: invalid",
```

tests/platform_helper/utils/test_validation.py
```diff
@@ -940,58 +940,66 @@ def test_validate_database_copy_multi_postgres_failures(capfd):
...
+        ),
+    ],
+)
 @patch("dbt_platform_helper.utils.validation.get_supported_redis_versions", return_value=["7.1"])
-def test_validate_extensions_supported_versions_successful_with_supported_version(
-    mock_supported_versions, capsys
+def test_validate_extensions_different_scenarios(
```
WillGibson, line 991: The different scenarios bit is covered by the parameterisation.
```suggestion
def test_validate_extension_supported_versions(
```

WillGibson (commented):
Think this will be my last pass :-)

PR #680, opened 2024-12-05
refactor: DBTP-1596 Separate out the platform-config.yml schema validation (part 1)

WillGibson:
Partially addresses [DBTP-1596](https://uktrade.atlassian.net/browse/DBTP-1596).

This is a first pass at separating the `platform-config.yml` schema validation from other validation concerns.

In readiness for the next effort, which will likely involve converting it to a class, this attempts to:

* organise it more coherently
* more clearly name things
* identify "private/public" elements
* reduce duplication
* remove unused elements

---
### Title:
### Description:
### Tasks:

Passing run: https://eu-west-2.console.aws.amazon.com/codesuite/codepipeline/pipelines/pull-request-end-to-end-tests/executions/90c68d28-753e-4f1a-98e8-b9db00b8351a/variables?region=eu-west-2

[DBTP-1596]: https://uktrade.atlassian.net/browse/DBTP-1596?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dbt_platform_helper/constants.py
```diff
@@ -1,6 +1,10 @@
+# Todo: Move to Config provider
 PLATFORM_CONFIG_FILE = "platform-config.yml"
+# Todo: Can we get rid of this yet?
 PLATFORM_HELPER_VERSION_FILE = ".platform-helper-version"
```
WillGibson, line 4: @antroy-madetech ?

dbt_platform_helper/utils/platform_config_schema.py
```diff
@@ -0,0 +1,603 @@
...
+import re
+
+from schema import Optional
+from schema import Or
+from schema import Regex
+from schema import Schema
+from schema import SchemaError
+
+
+def _string_matching_regex(regex_pattern: str):
```
ksugden, line 11: from the error message, this function is a copilot specific validation so should probably be named as such

dbt_platform_helper/constants.py
antroy-madetech: Yes, I think so. Any teams still using it will be forced to upgrade, so it'll be a breaking change.

WillGibson: I'm not clear from your answer whether you think we should keep support for this file?

dbt_platform_helper/utils/platform_config_schema.py
```diff
@@ -0,0 +1,603 @@
...
+
+
+def _integer_between(lower, upper):
+    def _is_between(value):
+        if isinstance(value, int) and lower <= value <= upper:
+            return True
+        # Todo: Raise suitable PlatformException?
+        raise SchemaError(f"should be an integer between {lower} and {upper}")
+
+    return _is_between
```
ksugden, line 30: I wonder if these validation functions would be easier to understand at first pass if they always returned something called `validation_function` or something like that?  I found the idea of `_is_between(value)` a bit confusing.

dbt_platform_helper/utils/platform_config_schema.py
WillGibson: It's actually only used to validate the branch name in the codebase pipeline, the bit of the error message was misleading. I have removed it.

dbt_platform_helper/utils/platform_config_schema.py
WillGibson: Reworked in person.

ksugden (approved):

PR #682, opened 2024-12-06
fix: DBTP-1290 - fix red herring "Not a Git repository" error

ksugden:
Addresses [DBTP-1290](https://uktrade.atlassian.net/browse/DBTP-1290)

When no pipelines were configured, we were not receiving the expected warning - `No pipelines defined: nothing to do`. 
This was because the check for pipelines included a check for legacy pipelines which was based on the presence of the `environments` key.  This is truthy even if there are no pipelines configured in `platform-config.yml`, meaning the warning echo was not called.

The legacy pipelines check is no longer needed as it is related to copilot environment pipelines which no-one uses any more.  This change removes that check.

I've added a single, rather arbitrary test which was failing before the change and passes now.  We already have tests covering config that is missing pipelines, and tests covering the behaviour when pipeline config is defined, so I have not added further tests with this change.

Also cleaned up a legacy copilot function.

---
### Title:
### Description:
### Tasks:

[DBTP-1290]: https://uktrade.atlassian.net/browse/DBTP-1290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dbt_platform_helper/commands/pipeline.py
```diff
@@ -178,29 +173,6 @@ def _generate_codebase_pipeline(
...
-
-    template_data = {
-        "app_name": app_name,
-        "git_repo": git_repo,
-        "codestar_connection_arn": codestar_connection_arn,
-        "pipeline_environments": configuration,
-    }
-
-    _create_file_from_template(
-        base_path, "environments/buildspec.yml", pipelines_dir, template_data, templates
```
JohnStainsby, line 194: I think all of these copilot env pipeline template files can also be removed
https://github.com/uktrade/platform-tools/tree/DBTP-1290/dbt_platform_helper/templates/pipelines/environments

JohnStainsby (approved):
I think

WillGibson (approved):

PR #723, opened 2025-01-14
feat: DBTP-1505 Generate terraform codebase pipeline configuration.

antroy-madetech:
Addresses [DBTP-1505](https://uktrade.atlassian.net/browse/DBTP-1505)

Templates the codebase pipeline configuration. Note: the terraform manifest is written in JSON format, as this makes it hugely easier to generate and test. It shouldn't need to be edited manually, but honestly isn't hugely worse that HCL.

---
### Title:
### Description:
### Tasks:

[DBTP-1505]: https://uktrade.atlassian.net/browse/DBTP-1505?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dbt_platform_helper/templates/environment-pipelines/main.tf
```diff
@@ -15,7 +15,7 @@ provider "aws" {
 }
 
 terraform {
-  required_version = "~> 1.8"
+  required_version = "{{ terraform_version }}"
```
antroy-madetech, line 18: So we can change these in a single place (constants.py)

dbt_platform_helper/templates/environment-pipelines/main.tf
```diff
@@ -27,7 +27,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = "~> 5"
+      version = "{{ aws_provider_version }}"
```
antroy-madetech, line 30: So we can change these in a single place (constants.py)

dbt_platform_helper/templates/environments/main.tf
```diff
@@ -7,12 +7,12 @@ locals {
...
     application    = "{{ application }}"
     services       = local.config["extensions"]
-    dns_account_id = local.env_config["{{ environment }}"]["accounts"]["dns"]["id"]
+    env_config     = local.env_config
   }
 }
 
 terraform {
-  required_version = "~> 1.8"
+  required_version = "{{ terraform_version }}"
```
antroy-madetech, line 15: So we can change these in a single place (constants.py)

dbt_platform_helper/templates/environments/main.tf
```diff
@@ -24,7 +24,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = "~> 5"
+      version = "{{ aws_provider_version }}"
```
antroy-madetech, line 27: So we can change these in a single place (constants.py)

dbt_platform_helper/domain/pipelines.py
```diff
@@ -32,22 +32,29 @@ def __init__(
...
+        self.terraform_manifest_provider = terraform_manifest_provider
+        self.ecr_provider = ecr_provider
 
-    def generate(self, terraform_platform_modules_version, deploy_branch):
-        pipeline_config = self.config_provider.load_and_validate_platform_config()
+    def generate(self, cli_terraform_platform_modules_version, deploy_branch):
+        try:
+            self._generate_pipeline_config(cli_terraform_platform_modules_version, deploy_branch)
+        except Exception as exc:
+            self.abort(str(exc))
```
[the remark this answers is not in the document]
a-gleeson: generally we have been catchting exceptions at the command level and calling abort there. This may not be the final approach as a general exception story will be played but for consitency would it make sense to push the exception catching up to that layer?

dbt_platform_helper/domain/pipelines.py
```diff
@@ -32,22 +34,29 @@ def __init__(
         self.abort = abort
         self.get_git_remote = get_git_remote
         self.get_codestar_arn = get_codestar_arn
+        self.terraform_manifest_provider = terraform_manifest_provider
+        self.ecr_provider = ecr_provider
 
-    def generate(self, terraform_platform_modules_version, deploy_branch):
-        pipeline_config = self.config_provider.load_and_validate_platform_config()
+    def generate(self, cli_terraform_platform_modules_version, deploy_branch):
```
a-gleeson, line 40: Should we add type hints for the function?

dbt_platform_helper/domain/pipelines.py
```diff
@@ -63,147 +72,67 @@ def generate(self, terraform_platform_modules_version, deploy_branch):
...
-                    templates,
-                )
+            codebase_pipelines = platform_config[CODEBASE_PIPELINES_KEY]
+            provisioned_ecrs = set(self.ecr_provider.get_ecr_repo_names())
+            required_ecrs = {
+                codebase: f"{platform_config['application']}/{codebase}"
+                for codebase in codebase_pipelines.keys()
+                if f"{platform_config['application']}/{codebase}" in provisioned_ecrs
+            }
+            required_imports = {
```
a-gleeson, line 104: So it might be the time of the day but I am stuck reading what exactly this is doing here. 
Is this correct?
- you build a dictionary of `required_ecrs` if the ecr repo is in `provisioned_ecrs`. You build with the key: value of the codebase and the repo.
- you then build a dictionary of `required_imports` if the ecr repo is in `provisioned_ecrs`.  You build with the key: value of the `codebase` and the `repo`.
will the values of `required_ecrs` and `required_imports` ever be different?

dbt_platform_helper/providers/ecr.py
```diff
@@ -0,0 +1,20 @@
...
+    def __init__(self, session: Session = None):
+        self.session = session
+        self.client = None
+
+    def _get_client(self):
+        if not self.session:
+            self.session = get_aws_session_or_abort()
+        return self.session.client("ecr")
+
+    def get_ecr_repo_names(self):
```
a-gleeson, line 16: type hint for return value?

dbt_platform_helper/providers/ecr.py
```diff
@@ -0,0 +1,20 @@
...
+        return [
+            repo["repositoryName"]
+            for repo in self._get_client().describe_repositories().get("repositories", {})
```
a-gleeson, line 19: we may want to use the boto3 paginator for describe calls. Recently we have had a few bugs when results are >20

dbt_platform_helper/providers/platform_config_schema.py
```diff
@@ -127,14 +127,15 @@ def __alb_schema() -> dict:
         }
 
     @staticmethod
-    def __codebase_pipelines_schema() -> list[dict]:
-        return [
-            {
-                "name": str,
+    def __codebase_pipelines_schema() -> dict:
```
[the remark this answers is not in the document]
a-gleeson: Do you think we should somehow show the updated schema for the `yaml`. Even just for the PR at the moment

dbt_platform_helper/domain/pipelines.py
antroy-madetech: Yes, they'll be different if there is a new environment. i.e. if there is a required ecr that hasn't already been created. The required imports are the ECRs that are required but already exist (so need to be imported).

I'll check to see if there's a test that describes this case and add one if not. And perhaps rename some variables to make them more descriptive.

dbt_platform_helper/providers/ecr.py
antroy-madetech: I'll have a look.

dbt_platform_helper/constants.py
```diff
@@ -4,6 +4,8 @@
 PLATFORM_HELPER_VERSION_FILE = ".platform-helper-version"
 # Todo: Move to ???
 DEFAULT_TERRAFORM_PLATFORM_MODULES_VERSION = "5"
+SUPPORTED_TERRAFORM_VERSION = "~> 1.8"
```
WillGibson, line 7: https://uktrade.atlassian.net/wiki/spaces/DBTP/pages/4340056065/ADR009+Draft+-+Support+2+minor+releases+of+Terraform + https://github.com/hashicorp/terraform/releases
```suggestion
SUPPORTED_TERRAFORM_VERSION = "~> 1.9"
```

dbt_platform_helper/domain/pipelines.py
WillGibson: I think this suggestion applies in other places too.

dbt_platform_helper/domain/pipelines.py
```diff
@@ -63,147 +72,67 @@ def generate(self, terraform_platform_modules_version, deploy_branch):
...
-                    base_path,
-                    copilot_pipelines_dir,
-                    templates,
-                )
+            codebase_pipelines = platform_config[CODEBASE_PIPELINES_KEY]
+            provisioned_ecrs = set(self.ecr_provider.get_ecr_repo_names())
+            required_ecrs = {
+                codebase: f"{platform_config['application']}/{codebase}"
+                for codebase in codebase_pipelines.keys()
+                if f"{platform_config['application']}/{codebase}" in provisioned_ecrs
```
WillGibson, line 102: Does this line mean that `required_ecrs` will only ever contain `provisioned_ecrs`?

dbt_platform_helper/providers/platform_config_schema.py
```diff
@@ -94,12 +92,12 @@ def __alb_schema() -> dict:
             Optional("environments"): {
                 PlatformConfigSchema.__valid_environment_name(): Or(
                     {
-                        Optional("additional_address_list"): list,
-                        Optional("allowed_methods"): list,
-                        Optional("cached_methods"): list,
+                        Optional("additional_address_list"): [str],
```
WillGibson, line 95: I'm used to seeing this expressed as `list[str]`, e.g. https://docs.python.org/3/library/typing.html#type-aliases

What is prompting the `[str]` approach?

dbt_platform_helper/providers/platform_config_schema.py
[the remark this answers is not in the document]
WillGibson: I'm about to start looking at whether we can generate https://platform.readme.trade.gov.uk/reference/platform-config-yml/ from this schema somehow, so that it automatically stays up to date.

WillGibson (commented):
First pass as far up to and including `dbt_platform_helper/providers`.

tests/platform_helper/providers/test_terraform_manifest.py
```diff
@@ -0,0 +1,170 @@
...
+    assert json_content["import"]["for_each"] == '${["test_codebase"]}'
+    assert json_content["import"]["id"] == "${each.value}"
+    assert (
+        json_content["import"]["to"]
+        == "module.codebase-pipelines[each.key].aws_ecr_repository.this"
+    )
+
+
+@freeze_time("2025-01-16 13:00:00")
+def test_generate_codebase_pipeline_config_creates_required_imports_for_two_codebases(
```
a-gleeson, line 120: could this test be paramertised with the test above?

dbt_platform_helper/domain/pipelines.py
antroy-madetech: Looks like I refactored and left that `in provisioned_ecrs` filter in there twice.

dbt_platform_helper/providers/platform_config_schema.py
antroy-madetech: list[str] is wrong. It allows a list of anything. It's type hint syntax not schema syntax.

It's not been noticed before as it is overly permissive.

tests/platform_helper/domain/test_pipelines.py
```diff
@@ -120,6 +126,66 @@ def test_generate_pipeline_command_generate_terraform_files_for_environment_pipe
     )
 
 
+@pytest.mark.parametrize(
+    "cli_version, exp_version", [("6", "6"), (None, DEFAULT_TERRAFORM_PLATFORM_MODULES_VERSION)]
+)
+def test_generate_pipeline_generates_codebase_pipeline(
```
WillGibson, line 132:
```suggestion
def test_generate_calls_generate_codebase_pipeline_config(
```

tests/platform_helper/providers/test_terraform_manifest.py
```diff
@@ -0,0 +1,170 @@
...
+    assert file_provider.mkfile.call_count == 1
+    base_path, file_path, contents, overwrite = file_provider.mkfile.call_args.args
+    assert base_path == str(Path(".").absolute())
+    assert file_path == "terraform/codebase-pipelines/main.tf.json"
+    assert overwrite
+
+    json_content = json.loads(contents)
+
+    exp_version = version("dbt-platform-helper")
+    # Initial Comments
```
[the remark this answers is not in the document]
antroy-madetech: No. It was just to explain that the following tests were for the standard "comments" in the JSON. I'll remove or re-word.

tests/platform_helper/domain/test_pipelines.py
```diff
@@ -120,6 +126,66 @@ def test_generate_pipeline_command_generate_terraform_files_for_environment_pipe
...
+    pipelines = Pipelines(**mocks.params())
+    pipelines.generate(cli_version, None)
+
+    mock_t_m_p = mocks.mock_terraform_manifest_provider
+    mock_t_m_p.generate_codebase_pipeline_config.assert_called_once_with(
+        codebase_pipeline_config, exp_version, {}
+    )
+
+
+def test_generate_pipeline_generates_codebase_pipeline_with_imports(
```
WillGibson, line 149:
```suggestion
test_generate_calls_generate_codebase_pipeline_config_with_imports(
```

dbt_platform_helper/providers/platform_config_schema.py
```diff
@@ -127,14 +125,15 @@ def __alb_schema() -> dict:
         }
 
     @staticmethod
-    def __codebase_pipelines_schema() -> list[dict]:
-        return [
-            {
-                "name": str,
```
WillGibson, line 133: How is the pipeline named now?

tests/platform_helper/providers/test_config_validator.py
```diff
@@ -286,12 +287,164 @@ def test_validate_database_copy_fails_if_cross_account_with_incorrect_account_id
...
         ConfigValidator().validate_database_copy_section(config)
 
-    console_message = capfd.readouterr().err
+    console_message = capsys.readouterr().err
 
     msg = f"Incorrect value for 'from_account' for environment 'prod'"
     assert msg in console_message
 
 
+def test_validate_platform_config_fails_if_database_copy_config_is_invalid(
```
WillGibson, line 296: Maybe...
```suggestion
def test_validate_platform_config_fails_if_database_copy_to_and_from_are_the_same(
```

WillGibson: Or if we are not specifically testing this misconfiguration, perhaps asserting on the message is not needed here?

tests/platform_helper/providers/test_config_validator.py
```diff
@@ -286,12 +287,164 @@ def test_validate_database_copy_fails_if_cross_account_with_incorrect_account_id
...
+
+    message = capsys.readouterr().err
+
+    assert (
+        f"database_copy 'to' and 'from' cannot be the same environment in extension 'our-postgres'."
+        in message
+    )
+
+
+def test_validate_platform_config_catches_environment_pipeline_errors(platform_env_config, capsys):
```
WillGibson, line 324: Same question as the test above applies here.

tests/platform_helper/providers/test_ecr.py
```diff
@@ -0,0 +1,65 @@
...
+                "encryptionConfiguration": {"encryptionType": "AES256"},
+                "imageScanningConfiguration": {"scanOnPush": False},
+                "imageTagMutability": "MUTABLE",
+                "registryId": "12345",
+                "repositoryArn": "arn:aws:ecr:eu-west-2:12345:repository/some-other-repo",
+                "repositoryName": "some-other-repo",
+                "repositoryUri": "12345.dkr.ecr.eu-west-2.amazonaws.com/some-other-repo",
+            },
+        ],
+    }
```
WillGibson, line 56: Do we need all the items in the above response for the purposes of the test?

If not, we could just add a comment explaining that it only includes the [arts relevant for the test and lighten the load a bit.

tests/platform_helper/providers/test_terraform_manifest.py
```diff
@@ -0,0 +1,170 @@
...
+        == '${lookup(each.value, "additional_ecr_repository", null)}'
+    )
+    assert module["pipelines"] == "${each.value.pipelines}"
+    assert module["services"] == "${each.value.services}"
+    assert module["requires_image_build"] == '${lookup(each.value, "requires_image_build", true)}'
+    assert (
+        module["slack_channel"]
+        == '${lookup(each.value, "slack_channel", "/codebuild/slack_channel_id")}'
+    )
+    assert module["env_config"] == "${local.environments}"
```
WillGibson, line 85: Love the many, specific asserts. Will make it very easy to see where things go wrong, if they do.

tests/platform_helper/providers/test_terraform_manifest.py
```diff
@@ -0,0 +1,170 @@
...
+    assert overwrite
+
+    json_content = json.loads(contents)
+
+    exp_version = version("dbt-platform-helper")
+    # Initial Comments
+    assert "WARNING: This is an autogenerated file, not for manual editing." in json_content["//"]
+    assert f"Generated by platform-helper {exp_version} / 2025-01-16 13:00:00" in json_content["//"]
+
+    assert "import" in json_content
```
WillGibson, line 110: Should we assert that `import` is not present in the previous test?

tests/platform_helper/providers/test_terraform_manifest.py
```diff
@@ -0,0 +1,170 @@
...
+    assert base_path == str(Path(".").absolute())
+    assert file_path == "terraform/codebase-pipelines/main.tf.json"
+    assert overwrite
+
+    json_content = json.loads(contents)
+
+    exp_version = version("dbt-platform-helper")
+    # Initial Comments
+    assert "WARNING: This is an autogenerated file, not for manual editing." in json_content["//"]
+    assert f"Generated by platform-helper {exp_version} / 2025-01-16 13:00:00" in json_content["//"]
```
WillGibson, line 108: Already covered in `test_generate_codebase_pipeline_config_creates_file`...
```suggestion
```

tests/platform_helper/providers/test_terraform_manifest.py
```diff
@@ -0,0 +1,170 @@
...
+    )
+
+    contents = file_provider.mkfile.call_args.args[2]
+
+    json_content = json.loads(contents)
+
+    exp_version = version("dbt-platform-helper")
+    # Initial Comments
+    assert "WARNING: This is an autogenerated file, not for manual editing." in json_content["//"]
+    assert f"Generated by platform-helper {exp_version} / 2025-01-16 13:00:00" in json_content["//"]
```
WillGibson, line 137: Already covered in `test_generate_codebase_pipeline_config_creates_file`...
```suggestion
```

tests/platform_helper/providers/test_terraform_manifest.py
```diff
@@ -0,0 +1,170 @@
...
+    assert base_path == str(Path(".").absolute())
+    assert file_path == "terraform/codebase-pipelines/main.tf.json"
+    assert overwrite
+
+    json_content = json.loads(contents)
+
+    exp_version = version("dbt-platform-helper")
+    # Initial Comments
+    assert "WARNING: This is an autogenerated file, not for manual editing." in json_content["//"]
+    assert f"Generated by platform-helper {exp_version} / 2025-01-16 13:00:00" in json_content["//"]
```
WillGibson, line 168: Already covered in `test_generate_codebase_pipeline_config_creates_file`...
```suggestion
```

WillGibson (commented):
First pass on the rest...

tests/platform_helper/conftest.py
```diff
@@ -579,15 +579,17 @@ def valid_platform_config():
       prod:
         requires_approval: true
 
-codebase_pipelines:
+codebase-pipelines:
```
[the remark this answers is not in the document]
antroy-madetech: I'll change it. It's just mistyped in this fixture, which looks like it's only used in the ConfigProvider tests and no longer validates against the schema. I'll speak to Kate, but this fixture probably isn't used for much now. If it was validated against the schema it would fail.

dbt_platform_helper/providers/platform_config_schema.py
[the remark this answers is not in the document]
WillGibson: It feels like it would be massive scope creep for me to use Pydantic right now.

I'm hoping we can include some text with the basic stuff like, whether a property is required, what type it should be, so that we end up with something like the following, but automatically generated...

image

dbt_platform_helper/domain/pipelines.py
antroy-madetech: @WillGibson I think I have them all now.

dbt_platform_helper/providers/platform_config_schema.py
antroy-madetech: It's the key in the map. Same as environment pipelines.

tests/platform_helper/providers/test_terraform_manifest.py
antroy-madetech: Much better than the templated manifests where we do a shotgun text search on some things.

tests/platform_helper/providers/test_terraform_manifest.py
antroy-madetech: There's a specific test for it below.

dbt_platform_helper/providers/ecr.py
WillGibson: Example here: https://github.com/uktrade/platform-tools/blob/d9bbe4182e401a6d331d04ae929ab840dde82be0/dbt_platform_helper/utils/aws.py#L426

tests/platform_helper/providers/test_terraform_manifest.py
```diff
@@ -76,91 +80,69 @@ def test_generate_codebase_pipeline_config_creates_file(codebase_pipeline_config
...
+    assert json_content["import"]["id"] == "${each.value}"
     assert (
         json_content["import"]["to"]
-        == "${module.codebase-pipelines[each.key].aws_ecr_repository.this}"
+        == "module.codebase-pipelines[each.key].aws_ecr_repository.this"
     )
 
 
 @freeze_time("2025-01-16 13:00:00")
 def test_generate_codebase_pipeline_config_omits_include_block_if_no_codebases_provided(
```
WillGibson, line 130:
```suggestion
def test_generate_codebase_pipeline_config_omits_import_block_if_no_codebases_provided(
```

tests/platform_helper/domain/test_pipelines.py
```diff
@@ -120,6 +126,66 @@ def test_generate_pipeline_command_generate_terraform_files_for_environment_pipe
     )
 
 
+@pytest.mark.parametrize(
+    "cli_version, exp_version", [("6", "6"), (None, DEFAULT_TERRAFORM_PLATFORM_MODULES_VERSION)]
```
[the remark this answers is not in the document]
WillGibson: It might allow this, but don't lose sleep over it...
```suggestion
    "cli_version, exp_version",
    [
        ("6", "6"),
        (None, DEFAULT_TERRAFORM_PLATFORM_MODULES_VERSION),
    ]
```

dbt_platform_helper/providers/platform_config_schema.py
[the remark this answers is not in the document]
a-gleeson: On this task it is scope creep for sure. Although part of me feels like creating a central model for the platform config would be higher priority and then from there we could implement automatically generating docs from the pydantic model.
Is there a standard syntax we should be using for a schema?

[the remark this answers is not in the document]
WillGibson: > Is there a standard syntax we should be using for a schema?

I'm not sure what you mean? We're using YAML and it has to match this schema.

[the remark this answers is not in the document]
tony-griffin: @WillGibson would you gain access to the schema in the same manner we do the platform-tools commands.md page, or were you thinking of another method?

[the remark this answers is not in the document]
WillGibson: `Schema.json_schema()` nearly works for us.

See https://github.com/keleshev/schema/issues/329 for some context.

tests/platform_helper/providers/test_config.py
```diff
@@ -220,11 +78,13 @@ def test_validate_platform_config_succeeds_if_pipeline_account_matches_environme
...
-    # Should not error if config is sound.
-    config_provider = ConfigProvider(ConfigValidator())
+    mock_file_provider = Mock(spec=FileProvider)
+    mock_file_provider.load.return_value = platform_env_config
+    config_provider = ConfigProvider(ConfigValidator(), mock_file_provider)
     config_provider.config = platform_env_config
 
-    config_provider.validate_platform_config()
+    # Should not error if config is sound.
+    config_provider._validate_platform_config()
```
ksugden, line 87:
```suggestion
    config_provider.load_and_validate_platform_config()
```
since we're using the file provider mock, we need to include the load part in this test, otherwise the mocked config will not be loaded into the config attribute yet

tests/platform_helper/providers/test_config.py
```diff
@@ -15,6 +16,15 @@
 from tests.platform_helper.conftest import FIXTURES_DIR
 
 
+def test_comprehensive_platform_config_validates_successfully(valid_platform_config):
+    mock_file_provider = Mock(spec=FileProvider)
+    mock_file_provider.load.return_value = valid_platform_config
+    config_provider = ConfigProvider(ConfigValidator(), mock_file_provider)
+
+    config_provider.load_and_validate_platform_config()
+    # No assertions as this will raise an error if there is one.
```
ksugden, line 25: wish there was something we could do to be assured that the validations ran successfully against the expected config, but aside from implementing a ConfigValidatorSpy, or adding a `verbose` flag to the validator, I haven't had an idea how....

tests/platform_helper/providers/test_config.py
ksugden: We have considered loading config up front in the __init__ of the config provider which might provide a more intuitive structure to the class, but it was a disruptive change as we're suddenly getting errors the moment the ConfigProvider is initialised everywhere so hundreds of tests break, hence not done it yet.  But please comment  if you think that would be a good idea and we could pick it up again.

antroy-madetech: Ah - thought I'd done that, but must have missed it...

tests/platform_helper/providers/test_config.py
antroy-madetech: I agree. We could add a success property to the validator or something. 

Perhaps this is something that can be looked at in https://uktrade.atlassian.net/browse/DBTP-1703

The validator, rather than sys.exit'ing on error could return an object that holds a list of errors and a list of warnings and a success flag. And perhaps a `handle_errors_and_warnings` method that you'd call to act on any errors.

So in the actual code it would be used something like:

```
validator.load_and_validate_platform_config().handle_errors_and_warnings()
```

In tests:

```
result = validator.load_and_validate_platform_config()
assert result.success
```

ksugden (approved):

dbt_platform_helper/providers/platform_config_schema.py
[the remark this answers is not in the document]
a-gleeson: > > Is there a standard syntax we should be using for a schema?
> 
> I'm not sure what you mean? We're using YAML and it has to match this schema.

I had meant is there a standard way we should be respresnting the schema for users to view and understand? Like the screenshot you displayed for the doc page.

dbt_platform_helper/domain/pipelines.py
a-gleeson: The updated code now makes a lot more sense

a-gleeson (approved):

WillGibson (approved):

PR #725, opened 2025-01-15
refactor: DBTP-1678 maintenance page refactor

a-gleeson:
Addresses [DBTP-1678](https://uktrade.atlassian.net/browse/DBTP-1678)

---
## What

The maintenance page still used patches so the domain level was refactored to inject dependencies.
The application gracefully exits when a user says no to confirm prompts
Slightly imporved coverage.
Tests moved from legacy enivornment command.

---
### Title:
### Description:
### Tasks:

[DBTP-1678]: https://uktrade.atlassian.net/browse/DBTP-1678?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dbt_platform_helper/domain/database_copy.py
```diff
@@ -33,9 +33,7 @@ def __init__(
         db_connection_string: Callable[
             [Session, str, str, str, Callable], str
         ] = get_connection_string,
-        maintenance_page_provider: Callable[
-            [str, str, list[str], str, str], None
-        ] = MaintenancePage(),
+        maintenance_page: Callable[[str, str, list[str], str, str], None] = MaintenancePage,
```
james-francis-MT, line 36: What is the advantage of keeping a reference of the maintenance page class rather than creating an object and using that?

dbt_platform_helper/domain/database_copy.py
```diff
@@ -46,7 +44,7 @@ def __init__(
         self.auto_approve = auto_approve
         self.vpc_provider = vpc_provider
         self.db_connection_string = db_connection_string
-        self.maintenance_page_provider = maintenance_page_provider
+        self.maintenance_page = maintenance_page
```
james-francis-MT, line 47: Advantage of keeping a reference to the class rather than creating an object?

dbt_platform_helper/domain/maintenance_page.py
```diff
@@ -18,142 +18,293 @@
...
+        remove_maintenance_page: Callable[[boto3.Session, str], None] = remove_maintenance_page,
     ):
+        self.application = application
         self.user_prompt_callback = user_prompt_callback
         self.echo = echo
+        self.find_https_listener = find_https_listener
+        self.get_maintenance_page = get_maintenance_page
+        self.get_env_ips = get_env_ips
+        self.add_maintenance_page = add_maintenance_page
+        self.remove_maintenance_page = remove_maintenance_page
```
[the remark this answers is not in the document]
ksugden: It's the first stage of the refactor so that dependencies are injected into the usecase and can be mocked in test.  The next stage is to group these functions into coherent 'providers' where appropriate.

dbt_platform_helper/domain/maintenance_page.py
```diff
@@ -18,142 +18,293 @@
...
     pass
 
 
 class LoadBalancedWebServiceNotFoundException(MaintenancePageException):
-    pass
+    def __init__(self, application_name):
+        super().__init__(f"No services deployed yet to {application_name} ")
+
+
+def get_maintenance_page(session: boto3.Session, listener_arn: str) -> Union[str, None]:
```
ksugden, line 32: wondering why we're concerned with 'type' here.  We only seem to use this function for its truthy/fasley values

dbt_platform_helper/domain/maintenance_page.py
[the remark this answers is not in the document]
james-francis-MT: Cool, I'll keep an eye out for follow up PRs that implement this

dbt_platform_helper/domain/maintenance_page.py
ksugden: ah, I see type used in the user prompt.  Still, a bit awkward having a function that sounds like it returns the current mainenance page, but it actually returns the maintenance page type, if there is one....

tests/platform_helper/test_command_environment.py
```diff
@@ -43,14 +50,17 @@ def test_online_failure(self, mock_click, mock_maintenance_page):
 
         assert result.exit_code == 1
         mock_click.assert_called_with("""i've failed""", fg="red")
-        mock_maintenance_page_instance.deactivate.assert_called_with("test-app", "test-env")
+        mock_maintenance_page.assert_called_with(mock_application)
+        mock_maintenance_page_instance.deactivate.assert_called_with("test-env")
```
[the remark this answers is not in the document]
a-gleeson: maybe a bit verbos in my testing, due to updating how application is passed into the domain (via the __init__ for application and the env via the `deactivate` method) I wanted to assert that the maintenace page instantiation was called with the expected application.
This could all be quite redudant if we had integration tests which tested how all the components are tied together. Currently without them the best way to avoid that is with ensuring you assert how messages are called and manual testing.

dbt_platform_helper/domain/maintenance_page.py
a-gleeson: that's a good point, maybe I should update the name of the function?

dbt_platform_helper/domain/database_copy.py
james-francis-MT: Sorry, I mean you might as well do `self.maintenance_page = maintenance_page(self.app)` and create the object when initialising

a-gleeson: updated and will test the cross account functionality in the e2e tests

dbt_platform_helper/domain/maintenance_page.py
a-gleeson: updated to `get_maintenance_page_type`

dbt_platform_helper/commands/environment.py
```diff
@@ -36,8 +37,11 @@ def environment():
 @click.option("--vpc", type=str)
 def offline(app, env, svc, template, vpc):
     """Take load-balanced web services offline with a maintenance page."""
+
+    application = load_application(app)
```
ksugden, line 41: I guess there's a chance of ApplicationNotFoundException here.  Maybe we should move the try above load_application to catch that (it's also a PlatfromException)

dbt_platform_helper/commands/environment.py
```diff
@@ -36,8 +37,11 @@ def environment():
...
     try:
-        MaintenancePage().activate(app, env, svc, template, vpc)
+        # TODO move application loading here
```
ksugden, line 43: can we do this TODO now?

dbt_platform_helper/domain/maintenance_page.py
```diff
@@ -18,142 +18,293 @@
...
 
 
 class MaintenancePage:
     def __init__(
         self,
+        application: Application,
         user_prompt_callback: Callable[[str], bool] = click.confirm,
         echo: Callable[[str], str] = click.secho,
+        find_https_listener: Callable[[boto3.Session, str, str], str] = find_https_listener,
+        get_maintenance_page: Callable[
```
ksugden, line 165: just as a TODO - I think the get_maintenance_page, add_maintenance_page and remove_maintenance_page should all be functions that live in the MaintenancePage class, but that all have a dependency on a LoadBalancerProvider which provides the necessary to activate and deactivate maintenance pages.

tests/platform_helper/domain/test_maintenance_page.py
```diff
@@ -537,3 +538,454 @@ def test_get_rules_tag_descriptions(self, mock_boto_client):
...
+        session = Mock()
+        sessions = {"000000000": session}
+        dummy_application = Application(app_name)
+        dummy_application.environments = {env: Environment(env, "000000000", sessions)}
+        dummy_application.services = {"web": Service("web", "Load Balanced Web Service")}
+        self.application = dummy_application
+
+        self.user_prompt_callback = kwargs.get("user_prompt_callback", Mock())
+        self.echo = kwargs.get("echo", Mock())
+        self.find_https_listener = kwargs.get(
```
ksugden, line 554: Do we have unit tests for these injected functions somewhere?

dbt_platform_helper/commands/environment.py
a-gleeson: good catch

dbt_platform_helper/commands/environment.py
a-gleeson: updated, this was left over but still relevant moving it into the `try` `catch`

tests/platform_helper/domain/test_maintenance_page.py
a-gleeson: yup tested and echo functions have been asserted within the tests

dbt_platform_helper/domain/maintenance_page.py
a-gleeson: I will add a todo for this

ksugden (approved):

PR #729, opened 2025-01-21
refactor: DBTP-1698 -  Implementing IO provider in codebase and conduit commands

ksugden:
Addresses [DBTP-1698](https://uktrade.atlassian.net/browse/DBTP-1698)

The IO provider replaces individual "click" commands currently injected into the domains.

In terms of error handling, any exceptions raised by the IO provider will bubble up to the command level.
Currently we're only catching the exceptions related to "confirm" (e.g. non-yes or no input) but the idea is to extend this to all the other functions.

---
### Title:
### Description:
### Tasks:

[DBTP-1698]: https://uktrade.atlassian.net/browse/DBTP-1698?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dbt_platform_helper/providers/io.py
```diff
@@ -0,0 +1,32 @@
+import click
+
+from dbt_platform_helper.platform_exception import PlatformException
+
+
+class ClickIOProvider:
```
james-francis-MT, line 6: Should these be static methods or is it best to initialise a click provider?

tests/platform_helper/providers/test_io.py
```diff
@@ -0,0 +1,82 @@
...
+
+import pytest
+
+from dbt_platform_helper.providers.io import ClickIOProvider
+from dbt_platform_helper.providers.io import ClickIOProviderException
+
+
+class TestClickIOProvider:
+    def test_warn_stdout(self, capsys):
+        io = ClickIOProvider()
```
james-francis-MT, line 12: Object is initialised in the tests but use the class in the code. Would be good to make this consistent

james-francis-MT (commented):
What is the definition of a provider? Seems like such a generic term here that it might as well be dropped

james-francis-MT:
What is the definition of a provider? Seems like such a generic term here that it might as well be dropped

dbt_platform_helper/providers/io.py
WillGibson: I think we generally initialise them, so that may be more consistent with the general approach.

james-francis-MT:
> What is the definition of a provider? Seems like such a generic term here that it might as well be dropped

Had a look at the providers folder and just seems like a bunch of classes for various things?

dbt_platform_helper/providers/io.py
ksugden: Probably better in terms of mocking to initialise them.  We've committed that now.

ksugden:
Yeah, the Provider term is a bit redundant (hence dropped it from the class attribute names!).  We used it initially for these sorts of classes as they are 'providing' stuff to the domain classes to do their job, rather than being 'entities' or dataclass type objects.  We will probably come back and review the directory structure and class names as we further decouple things.

james-francis-MT (approved):

PR #738, opened 2025-01-27
feat: DBTP-1635 Add additional S3 source bucket validation 

tony-griffin:
Addresses [DBTP-1635](https://uktrade.atlassian.net/browse/DBTP-1635)

- GREAT service has the requirement of pulling data from multiple S3 source buckets.
- Validation added to support this configuration in a service's `<application>-deploy/platform-config.yml` file

Docs PR: https://github.com/uktrade/platform-documentation/pull/457
---
### Title:
### Description:
### Tasks:

[DBTP-1635]: https://uktrade.atlassian.net/browse/DBTP-1635?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dbt_platform_helper/domain/config_validator.py
+++ b/dbt_platform_helper/domain/config_validator.py
@@ -23,6 +23,7 @@ class ConfigValidator:
             self.validate_codebase_pipelines,
             self.validate_environment_pipelines_triggers,
             self.validate_database_copy_section,
+            self.validate_database_migration_input_sources,
         ]
         self.io = io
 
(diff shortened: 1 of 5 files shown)


dbt_platform_helper/providers/platform_config_schema.py
```diff
@@ -443,14 +443,17 @@ def _valid_s3_bucket_arn(key):
...
         _valid_s3_data_migration = {
-            "import": {
-                Optional("source_kms_key_arn"): PlatformConfigSchema.__valid_kms_key_arn(
-                    "source_kms_key_arn"
-                ),
-                "source_bucket_arn": _valid_s3_bucket_arn("source_bucket_arn"),
-                "worker_role_arn": PlatformConfigSchema.__valid_iam_role_arn("worker_role_arn"),
-            },
+            Optional("import"): _single_import_config,
+            Optional("import_sources"): [_single_import_config],
```
WillGibson, line 456: Isn't this invalid if it has both?

dbt_platform_helper/domain/config_validator.py
```diff
@@ -241,3 +242,34 @@ def validate_database_copy_section(self, config):
...
+        for extension_name, extension in s3_extensions.items():
+            for env, env_config in extension.get("environments", {}).items():
+                if "data_migration" not in env_config:
+                    continue
+                data_migration = env_config.get("data_migration", {})
+                if "import" in data_migration and "import_sources" in data_migration:
+                    errors.append(
+                        f"Error in '{extension_name}.environments.{env}.data_migration': only the 'import_sources' property is required - 'import' is deprecated."
+                    )
+                if "import" not in data_migration and "import_sources" not in data_migration:
```
ejayesh, line 269: Shouldn't we just be checking if "import_source" is not in data_migration?  "import" shouldn't be in there anyway.  Unless this should be an OR if we still want to allow "import" until we remove it at a later date.

ejayesh: Ah this is to check neither is in there

ejayesh (approved):

dbt_platform_helper/providers/platform_config_schema.py
antroy-madetech: See the code block that Jayesh commented on above. It covers the case where either both have been supplied or neither have. See tests below.

WillGibson (approved):

PR #746, opened 2025-01-30
fix: DBTP-1387 Fix maintenance page offline command when specifiying a specific service

a-gleeson:
Addresses [DBTP-1387](https://uktrade.atlassian.net/browse/DBTP-1387)

---

## What:
- fix to add host header for each service that is being taken offline
- reduces calls to get host header conditions from 3 to 1 for each service
- improves naming of the `get_host_conditions` to `get_host_header_conditions`
### Title:
### Description:
### Tasks:
- ~[ ] [Run the end to end tests for this branch]([https://github.com/uktrade/platform-tools?tab=readme-ov-file#regression-tests](https://github.com/uktrade/platform-end-to-end-tests?tab=readme-ov-file#running-the-tests)) and confirm that they are passing~
- e2e test will be added in https://uktrade.atlassian.net/browse/DBTP-1419
- Manually tested thourghly to confirm working well

[DBTP-1387]: https://uktrade.atlassian.net/browse/DBTP-1387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dbt_platform_helper/domain/maintenance_page.py
+++ b/dbt_platform_helper/domain/maintenance_page.py
@@ -154,7 +154,16 @@ def add_maintenance_page(
                     "Field": "path-pattern",
                     "PathPatternConfig": {"Values": ["/*"]},
                 },
-                *host_header_conditions,
+                {
+                    "Field": "host-header",
+                    "HostHeaderConfig": {
+                        "Values": [
+                            value
+                            for condition in host_header_conditions
+                            for value in condition["HostHeaderConfig"]["Values"]
+                        ]
+                    },
+                },
             ],
             Actions=[
                 {
--- a/dbt_platform_helper/domain/maintenance_page.py
+++ b/dbt_platform_helper/domain/maintenance_page.py
@@ -157,12 +157,14 @@ def add_maintenance_page(
                 {
                     "Field": "host-header",
                     "HostHeaderConfig": {
-                        "Values": list(
-                            {
-                                value
-                                for condition in host_header_conditions
-                                for value in condition["HostHeaderConfig"]["Values"]
-                            }
+                        "Values": sorted(
+                            list(
+                                {
+                                    value
+                                    for condition in host_header_conditions
+                                    for value in condition["HostHeaderConfig"]["Values"]
+                                }
+                            )
                         )
                     },
                 },
--- a/dbt_platform_helper/domain/maintenance_page.py
+++ b/dbt_platform_helper/domain/maintenance_page.py
@@ -90,7 +90,7 @@ def add_maintenance_page(
     app: str,
     env: str,
     services: List[Service],
-    allowed_ips: tuple,
+    allowed_ips: List[str],
     template: str = "default",
 ):
     lb_client = session.client("elbv2")
--- a/dbt_platform_helper/domain/maintenance_page.py
+++ b/dbt_platform_helper/domain/maintenance_page.py
@@ -107,7 +107,9 @@ def add_maintenance_page(
             if not target_group_arn:
                 continue
 
-            # TODO get host header conditions here and pass in, update the list of host header conditions
+            conditions = get_host_conditions(lb_client, listener_arn, target_group_arn)
+            host_header_conditions.extend(conditions)
+
             for ip in allowed_ips:
                 create_header_rule(
                     lb_client,
@@ -117,6 +119,7 @@ def add_maintenance_page(
                     [ip],
                     "AllowedIps",
                     next(rule_priority),
+                    conditions,
                 )
                 create_source_ip_rule(
                     lb_client,
@@ -125,6 +128,7 @@ def add_maintenance_page(
                     [ip],
                     "AllowedSourceIps",
                     next(rule_priority),
+                    conditions,
                 )
 
             create_header_rule(
@@ -135,6 +139,7 @@ def add_maintenance_page(
                 [bypass_value],
                 "BypassIpFilter",
                 next(rule_priority),
+                conditions,
             )
 
         click.secho(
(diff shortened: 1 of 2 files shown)


tests/platform_helper/domain/test_maintenance_page.py
```diff
@@ -868,6 +943,115 @@ def test_listener_roll_back_on_exception_multiple_services(
...
+    )
+    @mock_aws
+    @patch(
+        "dbt_platform_helper.domain.maintenance_page.random.choices", return_value=["a", "b", "c"]
+    )
+    @patch("dbt_platform_helper.domain.maintenance_page.get_maintenance_page_template")
+    def test_host_header_conditions_in_maintenance_page_rule(
+        self,
+        get_maintenance_page_template,
+        choices,
```
james-francis-MT, line 972: Unused mock?

tests/platform_helper/domain/test_maintenance_page.py
```diff
@@ -868,6 +943,115 @@ def test_listener_roll_back_on_exception_multiple_services(
...
+            "HostHeaderConfig": {"Values": ["/test-path"]},
+        }
+        assert rules[0]["Priority"] == "500"
+        assert rules[1]["Conditions"][0] == {
+            "Field": "host-header",
+            "HostHeaderConfig": {"Values": ["/test-path", "/test-path-2"]},
+        }
+        assert rules[1]["Priority"] == "501"
+
+        # check rule number 6 as it is the maintenance page Rule, validate that it has a host header for web but not web2
```
james-francis-MT, line 1043: Comment perhaps doesn't make much sense in parameterised test

james-francis-MT (approved):
Couple minor comments but otherwise looks good

tests/platform_helper/domain/test_maintenance_page.py
a-gleeson: looks unused as it has no assert checks but this mocks the line 98 `bypass_value = "".join(random.choices(string.ascii_lowercase + string.digits, k=12))` to make the random generation consitent in the test.

tests/platform_helper/domain/test_maintenance_page.py
```diff
@@ -868,6 +943,115 @@ def test_listener_roll_back_on_exception_multiple_services(
...
+        rules = elbv2_client.describe_rules(ListenerArn=listener_arn)["Rules"]
+        assert len(rules) == 3
+
+        add_maintenance_page(
+            boto3.Session(),
+            listener_arn,
+            "test-application",
+            "development",
+            [mock_application.services[service] for service in services],
+            ["1.2.3.4"],
```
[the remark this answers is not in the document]
a-gleeson: the method signature indicates that, alot have tests have passed a list. I'll update and re-test :)

dbt_platform_helper/domain/maintenance_page.py
```diff
@@ -106,6 +107,9 @@ def add_maintenance_page(
             if not target_group_arn:
                 continue
 
+            conditions = get_host_header_conditions(lb_client, listener_arn, target_group_arn)
+            host_header_conditions.extend(conditions)
```
ksugden, line 111: without full context, just trying to understand what is happening here:  each service has its own target_group_arn, but host_header_conditions is collecting all header conditions for all services.  Each time through the svc loop, create_header_rule is called with a longer list of host header conditions?  Is this a change in functionality?

ksugden: ok, got it now - line 472 XD

tests/platform_helper/domain/test_maintenance_page.py
james-francis-MT: Ahh right, didn't realise you needed it to be consistent

james-francis-MT: I can't see where the mocked values are used in the test, and it passes fine without it

tests/platform_helper/domain/test_maintenance_page.py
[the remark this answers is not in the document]
james-francis-MT: Yeah, it just loops through so doesn't matter to python if it a list or a tuple but types should match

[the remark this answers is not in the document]
james-francis-MT: If you are going to use type hints, it might be worth running a linter to make sure they are correct - maybe something like: https://mypy-lang.org/

ksugden (approved):

tests/platform_helper/domain/test_maintenance_page.py
a-gleeson: so it patches that function without needing to be accessed within the test at all. There are a few places where we use patches that aren't accessed. I think in this case we don't need to assert it.

tests/platform_helper/domain/test_maintenance_page.py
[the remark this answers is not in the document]
a-gleeson: thanks for that, we have only recently started to use them so we should implement something like that for sure!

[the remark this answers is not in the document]
a-gleeson: I'll have a proper look at this next week but the IP list comes from [get_env_ips](https://github.com/uktrade/platform-tools/blob/b60ad7b8f591cbb97620eb884de5461c9d3a6375/dbt_platform_helper/domain/maintenance_page.py#L73) which returns a List.

dbt_platform_helper/domain/maintenance_page.py
a-gleeson: after a chat this has been cleared up and we are moving line 111 until after the functions called within the loop. This is because it makes it clearer that `host_header_conditions` (renamed to maintenance_page_host_header_conditions) is used for the maintenace page rule outwith the loop while `conditions` (renamed to service_conditions) is used within the loop.

PR #750, opened 2025-02-03
test: DBTP-1675 Increase coverage in terraform manifest generator test

ddwilliams74:
Addresses [DBTP-1675](https://uktrade.atlassian.net/browse/DBTP-1675)

---
### Title:
### Description:

~~- [ ] If the work includes user interface changes, before and after screenshots included in description~~
~~- [ ] Includes any applicable changes to the documentation in this code base~~
~~- [ ] Includes link(s) to any applicable changes to the documentation in the [DBT Platform Documentation]~~(https://platform.readme.trade.gov.uk/) (can be to a pull request)
### Tasks:

[DBTP-1675]: https://uktrade.atlassian.net/browse/DBTP-1675?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


WillGibson:
Can the pull request title more clearly summarise what the commit changes?

tests/platform_helper/domain/test_terraform_environment.py
```diff
@@ -73,3 +78,32 @@ def test_terraform_environment_generate_writes_the_expected_manifest_to_file(
...
+        terraform_manifest_generator = PlatformTerraformManifestGenerator(mock_file_provider)
+        generated_manifest = terraform_manifest_generator.generate_manifest(
+            "test-env",
+            "test-app",
+            self.VALID_ENV_CONFIG,
+        )
+
+        # Get expected manifest as string using jinja2 render on example manifest file
+        env = Environment(loader=FileSystemLoader({FIXTURES_DIR}))
+        expected_manifest = env.get_template("terraform_environment_manifest_valid.yml").render()
```
WillGibson, line 107: As demonstrated [here](https://github.com/uktrade/platform-tools/blob/main/tests/platform_helper/providers/test_terraform_manifest.py#L21), we are trying to move away from asserting whole files match, and move to asserting on individual values in the resulting YAML.

We hope that this leads to tests that are easier to see where failures occurred and require less of these voluminous and often brittle expected test files.

ddwilliams74: Thanks for the feedback @WillGibson . I'll go through this with the team tomorrow in the squad 2 sprint.

ksugden: @WillGibson We have fairly extensive assertions on the terraform environment manifest [here ](https://github.com/uktrade/platform-tools/blob/main/dbt_platform_helper/domain/test_platform_terraform_manifest_generator.py) (which as an aside needs moving into the tests directory). The idea of this test was to add a 'belts and braces' check on the entire file contents to catch any additional text being injected, or things occurring in the wrong order, as these would not be caught by individual assertions.  We were mirroring tests in the copilot environment which is based on a fixture file.  This test existed before the refactor and we decided not to remove it so as not to reduce our code coverage.  We thought it'd be a good idea to also have such a test on the terraform environment.  I'm ok with not including this test however, if you think it's more than required and will just lead to awkward failures in the future.

ddwilliams74: I've been thinking about this a bit more. If we want to move from asserting whole files to individual values but still think there's value in validating the manifest structure and position of elements, we could assert against a list rather than a string. If we split on the string returned from generate_template to create a list we can then assert each line positionally as well as contents wise, all via the same test. e.g., assert expected_manifest[0] == "# First line comment"

WillGibson: Oh, this might be worth a play with: https://pypi.org/project/python-hcl2/

ksugden: Thanks, we'll look into the python-hcl thing.  Using the template engine was a workaround because of line breaks messing up the comparison faff when just using load :-/

PR #752, opened 2025-02-04
chore: Use mock file provider over fakefs

ksugden:
Off ticket.

We noticed some of the config tests were generating and deleting files from the fs.  These have been updated to instead use mock file provider which returns the expected platform config dictionary.

Other tests in this test file which were using fakefs have also been updated to use the mock file provider.

Some tests were deleted because they were covered in the file provider tests.

The file being generated during tests, `platform-config.yml` has been deleted.

Tests have been grouped into classes.  This is to aid upcoming refactor where some of these tests should be moved to the validator tests as they are concerned with validations rather than config provider.

The diff looks huge because of moving tests into classes :( Might be easier to follow individual commits.

---
### Title:
### Description:
### Tasks:


tests/platform_helper/providers/test_config.py
```diff
@@ -12,548 +12,518 @@
...
-
-    config_provider = ConfigProvider(ConfigValidator())
-    config_provider.config = config
-    mock_io = Mock()
-    config_provider.io = mock_io
+        config_provider.load_and_validate_platform_config()
+        # No assertions as this will raise an error if there is one.
 
-    config_provider._validate_platform_config()
+    def test_load_and_validate_exits_if_load_fails_with_duplicate_keys_error(self, capsys):
```
a-gleeson, line 28: as class based can we simplify name?

tests/platform_helper/providers/test_config.py
```diff
@@ -12,548 +12,518 @@
...
-    [
-        "pipeline/platform-config.yml",
-        "pipeline/platform-config-with-public-repo.yml",
-        "pipeline/platform-config-for-terraform.yml",
-    ],
-)
-def test_load_and_validate_config_valid_file(yaml_file):
-    """Test that, given the path to a valid yaml file, load_and_validate_config
-    returns the loaded yaml unmodified."""
+    def test_load_and_validate_exits_with_invalid_yaml(self, capsys):
```
a-gleeson, line 38: simplify?

tests/platform_helper/providers/test_config.py
```diff
@@ -12,548 +12,518 @@
...
-    }
+    @pytest.mark.parametrize(
+        "yaml_file",
+        [
+            "pipeline/platform-config.yml",
+            "pipeline/platform-config-with-public-repo.yml",
+            "pipeline/platform-config-for-terraform.yml",
+        ],
+    )
+    def test_load_and_validate_config_valid_file(self, yaml_file):
```
a-gleeson, line 96: simplfy?

tests/platform_helper/providers/test_config.py
```diff
@@ -12,548 +12,518 @@
...
+        path = FIXTURES_DIR / yaml_file
+        validated = config_provider.load_and_validate_platform_config(path=path)
+
+        with open(path, "r") as fd:
+            conf = yaml.safe_load(fd)
+
+        assert validated == conf
+
+
+class TestDataMigrationValidation:
```
a-gleeson, line 111: possible to simplify names within this class too?

chiaramapellimt (approved):

a-gleeson (approved):
LGTM!

tests/platform_helper/providers/test_config.py
chiaramapellimt: I think it will be part of the "naming" refactor we'll do

tests/platform_helper/providers/test_config.py
ksugden: Hi, yeah good shout.  I will add this to the follow up ticket, looking at the config validator as a lot of tests will move out of this file, so we can also do the renaming then

PR #759, opened 2025-02-06
fix: Dbtp 1602 resource renaming bug

antroy-madetech:
Addresses [DBTP-1602](https://uktrade.atlassian.net/browse/DBTP-1602)

This adds in a "moved" block for each of the S3 resources that had a "count" parameter added in tpm 5.4.0

This allows teams on older versions of tpm to upgrade without deleting their KMS keys.

---
### Title:
### Description:
### Tasks:


dbt_platform_helper/providers/terraform_manifest.py
```diff
@@ -159,15 +159,41 @@ def _add_environment_locals(terraform: dict, app: str):
...
+            ]
+            moves = [f'module.extensions.module.s3["{name}"].{resource}' for resource in resources]
+            for move in moves:
+                terraform["moved"].append(
+                    {
+                        "//": s3_comment,
+                        "from": move,
+                        "to": f"{move}[0]",
+                    }
+                )
```
WillGibson, line 195: I wonder whether it's worth having two methods now, with names asking it clear what each is adding.

antroy-madetech: I had thought of doing that, but it means checking in each to see if the "moved" block is there yet and appending to it if not. Happy to do that if you think it's worth it, but I suspect we'll be able to remove all these at the same time.

WillGibson: OK.

PR #761, opened 2025-02-07
refactor: DBTP-1766 - version provider

ksugden:
Addresses [DBTP-<ticket>](https://uktrade.atlassian.net/browse/DBTP-<ticket>)

---
### Title:
### Description:
### Tasks:


chiaramapellimt (approved):

dbt_platform_helper/providers/version.py
```diff
@@ -0,0 +1,35 @@
...
+import requests
+
+from dbt_platform_helper.providers.semantic_version import SemanticVersion
+
+
+class VersionProvider(ABC):
+    pass
+
+
+# TODO add timeouts and exception handling for requests
```
WillGibson, line 12: Is this todo not address yet on purpose or by omission?

dbt_platform_helper/providers/version.py
```diff
@@ -0,0 +1,35 @@
+from abc import ABC
+
+import requests
+
+from dbt_platform_helper.providers.semantic_version import SemanticVersion
+
+
+class VersionProvider(ABC):
+    pass
```
WillGibson, line 9: I started writing this suggestion, then realised that the two concrete implementations methods have different arguments. So I'm wondering what the benefit of the abstract class is?
```suggestion
class VersionProvider(ABC):
    @abstractmethod
    def get_latest_version(repo_name: str, tags: bool = False) -> SemanticVersion:
        pass
```

WillGibson: I wonder whether the interface can be the same if `GithubVersionProvider. get_latest_version()` returns whichever version is highest from the releases or tags?

WillGibson (commented):
Some questions.

dbt_platform_helper/providers/version.py
ksugden: on purpose as we were not currently handling this so need to do some investigation to decide what we should return in the case of not being able to resolve one of the versions.  Will be addressed in a follow up PR as part of the same ticket.

WillGibson (approved):

PR #764, opened 2025-02-07
fix: Delete old manifest before creating json one

antroy-madetech:
Addresses [DBTP-1602](https://uktrade.atlassian.net/browse/DBTP-1602)

Realised during manual testing that because we are moving to JSON formatted env manifests, if a team has an old `main.tf` file either in git or just locally, they will end up with 2 manifests trying to provision the same stuff in their env dir, so terraform apply will fall over. This PR cleans up that file first. 

---
### Title:
### Description:
### Tasks:

[DBTP-1602]: https://uktrade.atlassian.net/browse/DBTP-1602?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dbt_platform_helper/providers/files.py
+++ b/dbt_platform_helper/providers/files.py
@@ -24,3 +24,8 @@ class FileProvider:
         file.write_text(contents)
 
         return f"File {file_path} {action}"
+
+    @staticmethod
+    def delete_file_if_present(base_path: str, file_path: str | Path):
+        file_path = Path(base_path) / file_path
+        file_path.unlink(missing_ok=True)
--- a/dbt_platform_helper/providers/files.py
+++ b/dbt_platform_helper/providers/files.py
@@ -26,7 +26,7 @@ class FileProvider:
         return f"File {file_path} {action}"
 
     @staticmethod
-    def delete_file(base_path: str, file: str | Path):
+    def delete_file(base_path: str, file: str):
         file_path = Path(base_path) / file
         if file_path.exists():
             file_path.unlink()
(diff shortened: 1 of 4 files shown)


dbt_platform_helper/providers/files.py
```diff
@@ -24,3 +24,8 @@ def mkfile(base_path: str, file_path: str, contents: str, overwrite=False) -> st
         file.write_text(contents)
 
         return f"File {file_path} {action}"
+
+    @staticmethod
+    def delete_file_if_present(base_path: str, file_path: str | Path):
```
WillGibson, line 29: Is there ever a use case where we would want it to error if the file doesn't exist?

If we did, I think we might check for it's existence separately.

Would this be enough for the name?
```suggestion
    def delete_file(base_path: str, file_path: str | Path):
```

dbt_platform_helper/providers/terraform_manifest.py
```diff
@@ -194,11 +196,14 @@ def _add_moved(terraform, platform_config):
...
-            str(Path(".").absolute()),
-            tf_json,
+            str(Path(env_dir).absolute()),
+            "main.tf.json",
             json.dumps(terraform, indent=2),
             True,
         )
         self.io.info(message)
+
+    def _cleanup_old_manifests(self, env_dir):
```
WillGibson, line 208: Maybe...
```suggestion
    def _ensure_no_hcl_manifest_file(self, env_dir):
```

WillGibson (approved):

PR #774, opened 2025-02-17
refactor: DBTP-1796 Load balancer provider

a-gleeson:
Addresses [DBTP-1796](https://uktrade.atlassian.net/browse/DBTP-1796)

---
### Title:
### Description:
### Tasks:

[DBTP-1796]: https://uktrade.atlassian.net/browse/DBTP-1796?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dbt_platform_helper/domain/codebase.py
```diff
@@ -242,13 +244,6 @@ def __init__(self, codebase: str):
         super().__init__(f"""Your deployment for {codebase} was not triggered.""")
 
 
-class ApplicationEnvironmentNotFoundException(ApplicationException):
```
a-gleeson, line 245: This exception is used within maintenace page too so we moved it too a common place

dbt_platform_helper/domain/maintenance_page.py
```diff
@@ -55,21 +54,7 @@ def __str__(self):
...
-    maintenance_page_type = None
-    for description in tag_descriptions:
-        tags = {t["Key"]: t["Value"] for t in description["Tags"]}
-        if tags.get("name") == "MaintenancePage":
-            maintenance_page_type = tags.get("type")
-
-    return maintenance_page_type
-
-
+# TODO should this be in its own provider, inside the VPC one, what logic is this sepcific too?
```
a-gleeson, line 57: Out of scope of load balancer provider but probably one which can be easily refactored if there is a case for it.

dbt_platform_helper/domain/maintenance_page.py
```diff
@@ -84,164 +69,19 @@ def get_env_ips(vpc: str, application_environment: Environment) -> List[str]:
...
-        add_maintenance_page: Callable[
-            [boto3.Session, str, str, str, List[Service], tuple, str], None
-        ] = add_maintenance_page,
-        remove_maintenance_page: Callable[[boto3.Session, str], None] = remove_maintenance_page,
     ):
         self.application = application
-        self.get_https_listener_for_application = get_https_listener_for_application
         self.io = io
-        self.get_maintenance_page_type = get_maintenance_page_type
+        self.load_balancer_provider = load_balancer_provider  # TODO requires session from environment in application object which is only known during method execution
```
a-gleeson, line 82: we have an inconsistent approach to how we initialise providers, this is partly due to how we manage the Application object and should be revisted.

ksugden (approved):

dbt_platform_helper/domain/copilot_environment.py
```diff
@@ -38,6 +37,7 @@ def __init__(
         )
         self.io = io
         self.session = session
+        self.load_balancer = load_balancer_provider(session)
```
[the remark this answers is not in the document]
a-gleeson: So this would then create a new session. It would default to `None` and when setting the client in `LoadBalancerProvider` we do:
```
    def _get_client(self, client: str):
        if not self.session:
            self.session = get_aws_session_or_abort()
        return self.session.client(client)
```

PR #790, opened 2025-02-26
feat: DBTP-1845 Improve Maintenance Page logging

a-gleeson:
Addresses [DBTP-1845](https://uktrade.atlassian.net/browse/DBTP-1845)

---
### Title:
### Description:
### Tasks:

### DIFF
--- a/dbt_platform_helper/domain/maintenance_page.py
+++ b/dbt_platform_helper/domain/maintenance_page.py
@@ -38,20 +38,14 @@ class FailedToActivateMaintenancePageException(MaintenancePageException):
         application_name: str,
         env: str,
         original_exception: Exception,
-        rolled_back_rules: dict[str, bool] = {},
     ):
         super().__init__(
             f"Maintenance page failed to activate for the {application_name} application in environment {env}."
         )
         self.orginal_exception = original_exception
-        self.rolled_back_rules = rolled_back_rules
 
     def __str__(self):
-        return (
-            f"{super().__str__()}\n"
-            f"Rolled-back rules: {self.rolled_back_rules }\n"
-            f"Original exception: {self.orginal_exception}"
-        )
+        return f"{super().__str__()}\n" f"Original exception: {self.orginal_exception}"
 
 
 # TODO should this be in its own provider, inside the VPC one, what logic is this sepcific too?
@@ -187,6 +181,12 @@ class MaintenancePage:
                     listener_arn, target_group_arn
                 )
 
+                self.io.debug(
+                    "\n#----------------------------------------------------------#\n# Creating listener rules for service "
+                    + svc.name.ljust(21, " ")
+                    + "#\n#----------------------------------------------------------#\n\n",
+                )
+
                 for ip in allowed_ips:
                     self.load_balancer.create_header_rule(
                         listener_arn,
@@ -196,6 +196,7 @@ class MaintenancePage:
                         "AllowedIps",
                         next(rule_priority),
                         service_conditions,
+                        [{"Key": "service", "Value": svc.name}],
                     )
                     self.load_balancer.create_source_ip_rule(
                         listener_arn,
@@ -204,6 +205,7 @@ class MaintenancePage:
                         "AllowedSourceIps",
                         next(rule_priority),
                         service_conditions,
+                        [{"Key": "service", "Value": svc.name}],
                     )
 
                 self.load_balancer.create_header_rule(
@@ -214,12 +216,13 @@ class MaintenancePage:
                     "BypassIpFilter",
                     next(rule_priority),
                     service_conditions,
+                    [{"Key": "service", "Value": svc.name}],
                 )
 
                 # add to accumilating list of conditions for maintenace page rule
                 maintenance_page_host_header_conditions.extend(service_conditions)
 
-            self.io.info(
+            self.io.debug(
                 f"\nUse a browser plugin to add `Bypass-Key` header with value {bypass_value} to your requests. For more detail, visit https://platform.readme.trade.gov.uk/next-steps/put-a-service-under-maintenance/",
             )
 
@@ -262,9 +265,9 @@ class MaintenancePage:
                 ],
             )
         except Exception as e:
-            deleted_rules = self.__clean_up_maintenance_page_rules(listener_arn)
+            self.__clean_up_maintenance_page_rules(listener_arn)
             raise FailedToActivateMaintenancePageException(
-                app, env, f"{e}:\n {traceback.format_exc()}", deleted_rules
+                app, env, f"{e}:\n {traceback.format_exc()}"
             )
 
     def __clean_up_maintenance_page_rules(
@@ -275,13 +278,35 @@ class MaintenancePage:
             listener_arn
         )
 
-        deletes = {}
+        # keep track of rules deleted
+        deletes = {"MaintenancePage": 0}
         for name in ["MaintenancePage", "AllowedIps", "BypassIpFilter", "AllowedSourceIps"]:
-            deleted = self.load_balancer.delete_listener_rule_by_tags(tag_descriptions, name)
-            deletes[name] = bool(deleted)
-            if fail_when_not_deleted and name == "MaintenancePage" and not deleted:
+            deleted_list = self.load_balancer.delete_listener_rule_by_tags(tag_descriptions, name)
+
+            # track the rules deleted grouped by service
+            for deleted_rule in deleted_list:
+                tags = {t["Key"]: t["Value"] for t in deleted_rule["Tags"]}
+                if tags.get("service"):
+                    if tags["service"] not in deletes:
+                        deletes[tags["service"]] = {
+                            "AllowedIps": 0,
+                            "BypassIpFilter": 0,
+                            "AllowedSourceIps": 0,
+                        }
+                    deletes[tags["service"]][name] += 1
+                elif tags.get("name") == "MaintenancePage":
+                    deletes["MaintenancePage"] += 1
+
+            if (
+                fail_when_not_deleted
+                and name == "MaintenancePage"
+                and deletes["MaintenancePage"] == 0
+            ):
                 raise ListenerRuleNotFoundException()
 
+        self.io.warn(
+            f"Rules deleted by type and grouped by service: {deletes}",
+        )
         return deletes
 
     def __remove_maintenance_page(self, listener_arn: str) -> dict[str, bool]:
--- a/dbt_platform_helper/providers/io.py
+++ b/dbt_platform_helper/providers/io.py
@@ -7,10 +7,16 @@ class ClickIOProvider:
     def warn(self, message: str):
         click.secho(message, fg="magenta")
 
+    def debug(self, message: str):
+        click.secho(message, fg="green")
+
     def error(self, message: str):
         click.secho(f"Error: {message}", fg="red")
 
-    def info(self, message: str):
+    def info(
+        self,
+        message: str,
+    ):
         click.secho(message)
 
     def input(self, message: str) -> str:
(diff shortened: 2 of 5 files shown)


dbt_platform_helper/domain/maintenance_page.py
```diff
@@ -275,13 +278,35 @@ def __clean_up_maintenance_page_rules(
...
         )
 
-        deletes = {}
+        # keep track of rules deleted
+        deletes = {"MaintenancePage": 0}
         for name in ["MaintenancePage", "AllowedIps", "BypassIpFilter", "AllowedSourceIps"]:
-            deleted = self.load_balancer.delete_listener_rule_by_tags(tag_descriptions, name)
-            deletes[name] = bool(deleted)
-            if fail_when_not_deleted and name == "MaintenancePage" and not deleted:
+            deleted_list = self.load_balancer.delete_listener_rule_by_tags(tag_descriptions, name)
```
ksugden, line 284: not for just now, but maybe it'd make sense for `deletes` to be the data structure returned from `delete_listener_rule_by_tags` if that's the form we need it in for the logging?

ksugden: it's just quite a lot to take in in the context of the usecase (the sorting and tracking aspect)

tests/platform_helper/domain/test_maintenance_page.py
```diff
@@ -330,11 +330,12 @@ def test_listener_roll_back_on_exception(
 
         excepted = (
             "Maintenance page failed to activate for the test-application application in environment development.\n"
-            f"Rolled-back rules: {expected_roll_back_message}\n"
             "Original exception: An error occurred (ValidationError) when calling the CreateRule operation: Simulated failure"
         )
         assert str(e.value).startswith(excepted)
 
+        # TODO add check for warn output or mock Io and assert called with
```
ksugden, line 337: just to check this was intentionally left as a TODO?

dbt_platform_helper/domain/maintenance_page.py
```diff
@@ -38,20 +38,14 @@ def __init__(
         application_name: str,
         env: str,
         original_exception: Exception,
-        rolled_back_rules: dict[str, bool] = {},
```
antroy-madetech, line 41: I assume this defaulted parameter was never used?

dbt_platform_helper/domain/maintenance_page.py
```diff
@@ -187,6 +181,12 @@ def add_maintenance_page(
                     listener_arn, target_group_arn
                 )
 
+                self.io.debug(
+                    "\n#----------------------------------------------------------#\n# Creating listener rules for service "
+                    + svc.name.ljust(21, " ")
+                    + "#\n#----------------------------------------------------------#\n\n",
+                )
+
```
antroy-madetech, line 189:
```suggestion
messge = f"Creating listener rules for service {svc.name}"
headline = "-" * 50
self.io.debug(f"""
#{line}#
# {svc:48} #
#{line}#

""")
```

antroy-madetech: Or similar.

dbt_platform_helper/domain/maintenance_page.py
```diff
@@ -275,13 +278,35 @@ def __clean_up_maintenance_page_rules(
...
+                fail_when_not_deleted
+                and name == "MaintenancePage"
+                and deletes["MaintenancePage"] == 0
+            ):
                 raise ListenerRuleNotFoundException()
 
+        self.io.warn(
+            f"Rules deleted by type and grouped by service: {deletes}",
+        )
         return deletes
```
antroy-madetech, line 310: Does this still need returning? Doesn't look like it's used any more

dbt_platform_helper/domain/maintenance_page.py
```diff
@@ -275,13 +278,35 @@ def __clean_up_maintenance_page_rules(
...
         for name in ["MaintenancePage", "AllowedIps", "BypassIpFilter", "AllowedSourceIps"]:
-            deleted = self.load_balancer.delete_listener_rule_by_tags(tag_descriptions, name)
-            deletes[name] = bool(deleted)
-            if fail_when_not_deleted and name == "MaintenancePage" and not deleted:
+            deleted_list = self.load_balancer.delete_listener_rule_by_tags(tag_descriptions, name)
+
+            # track the rules deleted grouped by service
+            for deleted_rule in deleted_list:
+                tags = {t["Key"]: t["Value"] for t in deleted_rule["Tags"]}
+                if tags.get("service"):
```
antroy-madetech, line 289: More pythonic to do this I think:
```suggestion
                if "service" in tags:
```

tests/platform_helper/domain/test_maintenance_page.py
```diff
@@ -330,11 +330,12 @@ def test_listener_roll_back_on_exception(
...
+
```
antroy-madetech, line 338: Intentionally left in? Or is there a ticket to address this? If the latter it would be good to have the ticket number in here

dbt_platform_helper/domain/maintenance_page.py
a-gleeson: Maybe I had thought that but then that ties deleting rules to requiring the service tag's to group them. The code around here isn't my favourite as it is. We pass in a list of tag_descriptions and a name each time. tag_descriptions contain all the tags so we should just be able to pass in a list of name values and extract them all. 
It is a question of how much the funciton should do though.

tests/platform_helper/domain/test_maintenance_page.py
a-gleeson: ah yes, I'll add an assertion here, thanks!

dbt_platform_helper/domain/maintenance_page.py
a-gleeson: yea it wasn't, everytime this exception would be raised we would log the rolled back rule types but that is now a part of the standard delete logging

dbt_platform_helper/domain/maintenance_page.py
a-gleeson: No it doesn't, I can remove it for now as it isn't used.

tests/platform_helper/domain/test_maintenance_page.py
a-gleeson: Kate caught this too, I used to test it in the message now I want to cehck that the warning log is correct. I will mock IO to test assert_called_with

dbt_platform_helper/domain/maintenance_page.py
```diff
@@ -214,12 +216,13 @@ def add_maintenance_page(
...
                     next(rule_priority),
                     service_conditions,
+                    [{"Key": "service", "Value": svc.name}],
                 )
 
                 # add to accumilating list of conditions for maintenace page rule
                 maintenance_page_host_header_conditions.extend(service_conditions)
 
-            self.io.info(
+            self.io.debug(
```
WillGibson, line 225: Why has this changed to debug?

dbt_platform_helper/domain/maintenance_page.py
```diff
@@ -262,27 +265,48 @@ def add_maintenance_page(
...
+            self.__clean_up_maintenance_page_rules(listener_arn)
             raise FailedToActivateMaintenancePageException(
-                app, env, f"{e}:\n {traceback.format_exc()}", deleted_rules
+                app, env, f"{e}:\n {traceback.format_exc()}"
             )
 
     def __clean_up_maintenance_page_rules(
         self, listener_arn: str, fail_when_not_deleted: bool = False
-    ) -> dict[str, bool]:
+    ):
```
WillGibson, line 275: Maybe this now?
```suggestion
    ) -> None:
```

dbt_platform_helper/domain/maintenance_page.py
```diff
@@ -262,27 +265,48 @@ def add_maintenance_page(
...
 
         tag_descriptions = self.load_balancer.get_rules_tag_descriptions_by_listener_arn(
             listener_arn
         )
 
-        deletes = {}
+        # keep track of rules deleted
+        deletes = {"MaintenancePage": 0}
```
WillGibson, line 282: Could it be named such that the comment is not needed?
```suggestion
        deleted_rules = {"MaintenancePage": 0}
```

dbt_platform_helper/providers/io.py
```diff
@@ -7,10 +7,16 @@ class ClickIOProvider:
...
+        click.secho(message, fg="green")
+
     def error(self, message: str):
         click.secho(f"Error: {message}", fg="red")
 
-    def info(self, message: str):
+    def info(
+        self,
+        message: str,
+    ):
```
WillGibson, line 19: Seems odd that this one is multi-lined and the others are not...
```suggestion
    def info(self, message: str):
```

a-gleeson: I'll see if the linter picks this up

dbt_platform_helper/domain/maintenance_page.py
a-gleeson: I think this one should be info, good catch. I think other logs should be debug as we may want to suppress them for normal logging but still have a way to enable debugging logs with a flag

dbt_platform_helper/domain/maintenance_page.py
a-gleeson: I guess if a function with no return statement corresponds to a function which returns None.

dbt_platform_helper/domain/maintenance_page.py
antroy-madetech: Did you not have any success with this?

antroy-madetech: Oh - I see. You have changed it. I think this will result in the formatting being off. I think the "Creating..." line will be indented by 10 spaces or something. Prob best to either use explicit line breaks in a double-quoted string, or triple quotes without line breaks but dedented to avoid the extra whitespace.

Is it the "{svc.name:21}" that black doesn't like?

WillGibson: @antroy-madetech I don't understand your suggestion above. Is it fully formed?

dbt_platform_helper/providers/io.py
```diff
@@ -7,6 +7,9 @@ class ClickIOProvider:
     def warn(self, message: str):
         click.secho(message, fg="magenta")
 
+    def debug(self, message: str):
```
WillGibson, line 10: Out of interest, can we set the log level with configuration and/or a command option?

dbt_platform_helper/domain/maintenance_page.py
a-gleeson: So using the multi line strings works `"""` but originally I had set it up like with multiple lines and used normal double quote strings. Eg/
```
  self.io.debug(
                   "\n#----------------------------------------------------------#\n
                    f"# Creating listener rules for service {svc.name.ljust(21, " ")}#\n"
                    "#----------------------------------------------------------#\n\n",
                )
```
Black had a syntax error with the bracket `}` and would error out.

dbt_platform_helper/providers/io.py
a-gleeson: Currently no, I thought adding in this level could be the first step. 
My thinking is we could add a check initially the check if `DEBUG=True` and if it isn't this function wouldn't call the click function.

antroy-madetech (approved):

dbt_platform_helper/domain/maintenance_page.py
a-gleeson: updated too
```
self.io.debug(
                    f"""
#----------------------------------------------------------#
# Creating listener rules for service {svc.name.ljust(21, " ")}#
#----------------------------------------------------------#

""",
                )
```

PR #791, opened 2025-02-27
refactor: DBTP-1798 - consolidate required version and platform helper versioning

ksugden:
Addresses [DBTP-1798](https://uktrade.atlassian.net/browse/DBTP-1798)

Refactors include: 

* consolidation of `PlatformHelperVersioning` provider and `RequiredVersion` domain into a single domain object. 
    * Tests updated and also consolidated across provider and domain, as well as pertinent ones within the `utils/versioning.py`, to accommodate this refactor 
* Usages of `patch` and `fakefs` etc. throughout refactored test files in favour of provider mocks
* Addressed versioning related TODOs within these files

---
### Title:
### Description:
### Tasks:

[DBTP-1798]: https://uktrade.atlassian.net/browse/DBTP-1798?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dbt_platform_helper/domain/versioning.py
```diff
@@ -65,3 +98,65 @@ def check_platform_helper_version_mismatch(self):
                 f"v{required_version} specified for the project."
             )
             self.io.warn(message)
+
+    def _get_version_status(
```
rav-pradhan, line 102: We originally had tests for this (implicitly) private method in `test_versioning`, but removed them as we believe all of its logic was being tested through the public methods. I think there may be an argument as to whether this method has enough complexity to warrant retaining those tests, though, owing to the multiple shapes the `PlatformHelperVersionStatus` object returned could take (perhaps a wider discussion around whether this should be refactored into its own class)

dbt_platform_helper/domain/versioning.py
```diff
@@ -65,3 +78,85 @@ def check_platform_helper_version_mismatch(self):
...
+        self,
+        include_project_versions: bool = True,
+    ) -> PlatformHelperVersionStatus:
+        try:
+            locally_installed_version = self.local_version_provider.get_installed_tool_version(
+                "dbt-platform-helper"
+            )
+        except LocalVersionProviderException:
```
rav-pradhan, line 109: Codecov raised that l109-110 wasn't tested. However, I'm not sure if this would ever be executed? Presumably if the locally installed version can't be found then the execution shouldn't progress further? Or if it should continue, then some of the warning messages we show to the user needs to be tweaked as we're directly referencing `version_status.local` (e.g. instead of showing "You are running platform-helper {local_version} against {required_version", we display another error message)

tests/platform_helper/domain/test_versioning.py
```diff
@@ -1,203 +1,287 @@
...
+        result = PlatformHelperVersioning(
+            version_file_version_provider=mock_version_file_version_provider,
+            config_provider=mock_config_provider,
+            pypi_provider=mock_pypi_provider,
+            local_version_provider=mock_local_version_provider,
+            skip_versioning_checks=Mock(return_value=False),
+        ).get_required_version("bogus_pipeline_not_in_config")
 
-    result = required_version.get_required_platform_helper_version(version_status=version_status)
+        print(result)
```
ksugden, line 282:
```suggestion
```

dbt_platform_helper/domain/versioning.py
rav-pradhan: Discussed with Kate - further refactors may come with regards to this, but not within the scope of this ticket

dbt_platform_helper/domain/versioning.py
rav-pradhan: Discussed with Kate, removed this except block as it's unnecessary

tests/platform_helper/domain/test_maintenance_page.py
```diff
@@ -9,7 +9,6 @@
 from botocore.exceptions import ClientError
 from moto import mock_aws
 
-from dbt_platform_helper.domain.maintenance_page import MaintenancePage
 from dbt_platform_helper.domain.maintenance_page import *
```
chiaramapellimt, line 12: this might need looking at, we don't use * elsewhere, might be better expliciting what we're importing

chiaramapellimt (approved):

tests/platform_helper/domain/test_maintenance_page.py
ksugden: bizarre - don't know where that change came from but if I revert it 31 tests fail

ksugden: Leaving as * for this PR so as not to muddy the water more

chiaramapellimt: that's ok.

dbt_platform_helper/domain/versioning.py
```diff
@@ -1,62 +1,74 @@
...
 
-        if not out:
-            raise PlatformHelperVersionNotFoundException
 
-        return out
+class PlatformHelperVersioning:
+    def __init__(
+        self,
+        io: ClickIOProvider = ClickIOProvider(),
+        version_file_version_provider: DeprecatedVersionFileVersionProvider = DeprecatedVersionFileVersionProvider(
```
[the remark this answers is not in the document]
WillGibson: But as `deprecated_version_file_version_provider` it would be "what it says on the tin".

PR #792, opened 2025-02-27
feat: Merge in terraform-platform-modules (DBTP-1842)

antroy-madetech:
Addresses [DBTP-1842](https://uktrade.atlassian.net/browse/DBTP-1842)

Bring terraform-platform-modules into platform-tools.

### Templating

- Pointed all terraform manifest templates at git::<https://github.com/uktrade/platform-tools/terraform-platform-modules.git//<module_name>>?depth=1&ref=<platform_modules_version>

- Version to come from platform-config.yml

### Commands

`pipeline generate` 
& 
`environment generate-terraform` 
commands have an optional `--terraform-platform-modules-version` flag. This has been removed.

All references to `DEFAULT_TERRAFORM_PLATFORM_MODULES_VERSION` stripped out, along with anything that uses the  t-p-m version.

The platform-config.yml optional parameters:
```
   default_versions: terraform-platform-modules
``` 
and 

```
environments: 
    <env>: 
        versions: terraform-platform-modules
```  
are both removed.

---
### Title:
### Description:
### Tasks:

[DBTP-1842]: https://uktrade.atlassian.net/browse/DBTP-1842?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dbt_platform_helper/commands/environment.py
```diff
@@ -88,14 +87,15 @@ def generate(name):
     "--name", "-n", required=True, help="The name of the environment to generate a manifest for."
 )
 @click.option(
-    "--terraform-platform-modules-version",
-    help=f"Override the default version of terraform-platform-modules. (Default version is '{DEFAULT_TERRAFORM_PLATFORM_MODULES_VERSION}').",
+    "--platform-helper-version",
+    help=f"Override the default version of platform-helper. (Default version is the installed version for `dbt-platform-helper`.",
 )
-def generate_terraform(name, terraform_platform_modules_version):
+def generate_terraform(name, platform_helper_version):
```
ksugden, line 93: Does it make sense to have a `platform_helper_version` override at all?  It's a bit confusing to be using one version of `platform-helper` to run the `generate` command, but specify another version in the flag which controls some part of the internal behaviour (which tpm to use)
```
source = f"git::https://github.com/uktrade/platform-tools.git//terraform/codebase-pipelines?depth=1&ref={platform_helper_version}"
```
Could we do without the flag and just use a different platform helper version to run the command?  Is that the plan eventually?

WillGibson: Good shout. I think we should be using the version specified in `default_versions.platform-helper` to run the command.

dbt_platform_helper/providers/version_status.py
```diff
@@ -33,6 +43,7 @@ class PlatformHelperVersionStatus(VersionStatus):
     latest: Optional[SemanticVersion] = None
     deprecated_version_file: Optional[SemanticVersion] = None
     platform_config_default: Optional[SemanticVersion] = None
+    cli_override: Optional[SemanticVersion] = None
```
ksugden, line 46: do we need to have two `override` attributes?  Also, we shouldn't be mentioning `cli` in a class that is decoupled from the cli implementation.  Perhaps we could look at just having an `overrides` attribute which could hold information about specific pipeline overrides as well as extension overrides?

dbt_platform_helper/providers/version_status.py
```diff
@@ -78,3 +89,25 @@ def validate(self) -> dict:
             "warnings": warnings,
             "errors": errors,
         }
+
+    def get_required_platform_helper_version(self, io: ClickIOProvider):
```
ksugden, line 93: getting the required version should be the responsibility of the `PlatformHelperVersioning` class - Is it possible to extend the current implementation for resolving the required version?

tests/platform_helper/domain/test_maintenance_page.py
```diff
@@ -79,7 +79,10 @@ def _create_listener(self, elbv2_client):
             Name="test-load-balancer", Subnets=[subnet_id]
         )["LoadBalancers"][0]["LoadBalancerArn"]
         return elbv2_client.create_listener(
-            LoadBalancerArn=load_balancer_arn, DefaultActions=[{"Type": "forward"}]
+            LoadBalancerArn=load_balancer_arn,
+            DefaultActions=[{"Type": "forward"}],
+            Port=443,
```
ksugden, line 84: what's this about?

tests/platform_helper/providers/test_config.py
```diff
@@ -278,27 +277,6 @@ def test_validation_fails_if_invalid_default_version_keys_present(
...
 
-    @pytest.mark.parametrize(
-        "invalid_key",
-        (
-            "",
-            "invalid-key",
-            "platform-helper",  # platform-helper is not valid in the environment overrides.
-        ),
-    )
-    def test_validation_fails_if_invalid_environment_version_override_keys_present(
```
ksugden, line 289: I need to come back to this to understand why this test is no longer relevant.  Are we no longer allowing environment version overrides at all?

tests/platform_helper/providers/test_config.py
```diff
@@ -341,101 +319,12 @@ def test_apply_defaults(self):
...
-            ),
-            (
-                {"terraform-platform-modules": "4.0.0"},
-                {"terraform-platform-modules": "5.0.0"},
-                {"terraform-platform-modules": "6.0.0"},
-                {"terraform-platform-modules": "6.0.0"},
-            ),
-        ],
-    )
-    def test_apply_defaults_for_versions(
```
ksugden, line 420: (Also I need to come back to this one to understand why the platform-helper version overrides are no longer tested here)

tests/platform_helper/providers/test_config_validator.py
```diff
@@ -185,38 +169,6 @@ def test_validate_database_copy_multi_postgres_failures():
     )
 
 
-def test_validate_database_copy_fails_if_cross_account_with_incorrect_account_ids():
```
ksugden, line 188: Is this maybe part of another PR?

tests/platform_helper/providers/test_semantic_version.py
```diff
@@ -75,3 +75,25 @@ def test_from_string(self, suite):
         input_version, expected_version = suite
         result = SemanticVersion.from_string(input_version)
         assert result == expected_version
+
+    @pytest.mark.parametrize(
+        "version",
+        ["v1.2.3", "1.2.3", "V1.2.3", "v1-2-3", "v1-2.3"],
+    )
+    def test_is_semantic_version(self, version):
+        result = SemanticVersion.is_semantic_version(version)
```
ksugden, line 84: `SemanticVersion` being anything other than a `semantic_version` is a strange concept!

dbt_platform_helper/commands/environment.py
```diff
@@ -88,14 +87,15 @@ def generate(name):
     "--name", "-n", required=True, help="The name of the environment to generate a manifest for."
 )
 @click.option(
-    "--terraform-platform-modules-version",
```
ksugden, line 91: breaking change if we're removing a flag?

tests/platform_helper/domain/test_maintenance_page.py
tony-griffin: PORT and PROTOCOL are now required attributes. This raised an error when we pulled in the latest version of boto3.

tony-griffin: PORT and PROTOCOL are now required attributes. This raised an error when we pulled in the latest version of boto3.

tests/platform_helper/providers/test_config_validator.py
tony-griffin: @john removed some attributes from the DB copy config

tests/platform_helper/providers/test_load_balancers.py
```diff
@@ -46,6 +46,7 @@ def _create_listener(session, load_balancer_arn=None):
             LoadBalancerArn=load_balancer_arn,
             DefaultActions=[{"Type": "forward"}],
             Protocol="HTTPS",
+            Port=443,
```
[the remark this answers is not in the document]
tony-griffin: PORT and PROTOCOL are now required attributes. This raised an error when we pulled in the latest version of boto3.

dbt_platform_helper/commands/environment.py
tony-griffin: There was talk on dbt-platform about juts removing this flag as it will be a breaking change anyway:

https://ditdigitalteam.slack.com/archives/C046VC4HTKN/p1743430217304329

tests/platform_helper/providers/test_config_validator.py
```diff
@@ -13,30 +13,14 @@
         None,
         [{"from": "dev", "to": "test"}],
         [{"from": "test", "to": "dev"}],
-        [
-            {
-                "from": "prod",
-                "to": "test",
-                "from_account": "9999999999",
```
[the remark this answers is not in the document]
tony-griffin: Yes we have removed this configuration from TPM. @JohnStainsby  removed some attributes from the DB copy config

tests/platform_helper/providers/test_config.py
tony-griffin: No environment overrides were removed. We are only setting version overrides in `default_versions` & `environment_pipelines`

tests/platform_helper/providers/test_config.py
tony-griffin: platform-helper version overrides were removed from 'environments' as well as TPM overrides

tests/platform_helper/providers/test_semantic_version.py
WillGibson: I think `semantic_version` is a static method on `SemanticVersion` used to check a string is a semantic version.

It would be a weird method to use on an instance of `SemanticVersion` though.

dbt_platform_helper/providers/version_status.py
antroy-madetech: I think we need to have a cross-squad chat about how we override the version we want going forward. Have a think about the use cases and the possibility of having overrides for the purposes of E2E testing done using env-vars to avoid polluting the UI.

JohnStainsby (approved):

tests/platform_helper/providers/test_semantic_version.py
ksugden: oops, of course.

PR #793, opened 2025-02-28
refactor: DBTP-1829 Parameters Provider

DeveloperConnor:
Addresses [DBTP-1829](https://uktrade.atlassian.net/browse/DBTP-1829)

Updated codebase domain to use the new provider and streamlined a lot of the tests that were setting up unneeded mocks. 

Opted not to update the below files to use the new provider - even though they are making ssm calls:
- application.py - Is a util, would need some more thinking on how we want to bring a provider into this.
- utils/aws.py - Again a util, see above and I think this will likely fall away in future
- commands/secrets.py - Lots of moving parts here, this work should be done as part of the secrets refactor.
---
### Title:
### Description:
~~- [ ] If the work includes user interface changes, before and after screenshots included in description~~
~~- [ ] Includes any applicable changes to the documentation in this code base~~
~~- [ ] Includes link(s) to any applicable changes to the documentation in the [DBT Platform Documentation](https://platform.readme.trade.gov.uk/) (can be to a pull request)~~
### Tasks:
https://eu-west-2.console.aws.amazon.com/codesuite/codepipeline/pipelines/pull-request-end-to-end-tests/executions/999a23ac-a954-4dd4-8645-88cab6e41a44/visualization?region=eu-west-2)

[DBTP-1829]: https://uktrade.atlassian.net/browse/DBTP-1829?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/platform_helper/domain/test_codebase.py
```diff
@@ -253,12 +246,6 @@ def test_codebase_deploy_successfully_triggers_a_pipeline_based_deploy(mock_appl
 def test_codebase_deploy_exception_with_a_nonexistent_codebase(exception_type):
     mocks = CodebaseMocks(check_image_exists=Mock(side_effect=exception_type("application")))
 
-    client = mock_aws_client(mocks.get_aws_session_or_abort)
-
-    client.get_parameter.return_value = {
```
ksugden, line 258: I assume this mock can be removed because it never gets to the point where it tries to retrieve the parameter?

tests/platform_helper/domain/test_codebase.py
```diff
@@ -188,11 +190,6 @@ def test_codebase_build_does_not_trigger_deployment_without_confirmation():
...
     mocks.io.confirm.return_value = False
 
-    client = mock_aws_client(mocks.get_aws_session_or_abort)
-    client.get_parameter.return_value = {
-        "Parameter": {"Value": json.dumps({"name": "application"})},
-    }
-
     with pytest.raises(ApplicationDeploymentNotTriggered):
         codebase = Codebase(**mocks.params())
         codebase.build("test-application", "application", "ab1c234")
```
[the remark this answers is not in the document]
ksugden: just a thought - probably not for this PR since it is regarding the codebase domain

tests/platform_helper/providers/test_parameter_store.py
```diff
@@ -0,0 +1,73 @@
...
+import pytest
+from botocore.stub import Stubber
+
+from dbt_platform_helper.providers.parameter_store import (
+    ParameterNotFoundForPathException,
+)
+from dbt_platform_helper.providers.parameter_store import ParameterStore
+
+
+def test_get_ssm_parameter_by_name_sucsess():
```
ksugden, line 11:
```suggestion
def test_get_ssm_parameter_by_name_success():
```

ksugden:
you might find that when you merge main into this PR, the `@mock_aws` in the `copilot` tests becomes superfluous as you are mocking the `get_aws_session_or_abort` function

dbt_platform_helper/domain/codebase.py
```diff
@@ -46,6 +48,7 @@ def __init__(
         check_if_commit_exists: Callable[[str], str] = check_if_commit_exists,
         run_subprocess: Callable[[str], str] = subprocess.run,
     ):
+        self.parameter_store = parameter_provider
```
ksugden, line 51: could we have consistency on whether this is called `parameter_store` or `parameter_provider` - in copilot it's `self.parameter_provider` - maybe both should be `_store'?

ksugden:
Again, just a thought on this, not something to address with this PR - I feel like another abstraction could be used between the domain and the secrets provider to hide the fact we're digging into secrets - e.g. the copilot domain just cares about the log group arns and the codebase domain just cares about the codebase repository names... the parameter store is just really a repository for the information are we actually interested in.

tests/platform_helper/domain/test_codebase.py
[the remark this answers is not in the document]
DeveloperConnor: Out of scope for the PR but Ill add it so the code is left in a better place :)

dbt_platform_helper/domain/codebase.py
DeveloperConnor: Gone for parameter_provider for clarity.

tests/platform_helper/providers/test_parameter_store.py
```diff
@@ -0,0 +1,73 @@
...
+import pytest
+from botocore.stub import Stubber
+
+from dbt_platform_helper.providers.parameter_store import (
+    ParameterNotFoundForPathException,
+)
+from dbt_platform_helper.providers.parameter_store import ParameterStore
+
+
+def test_get_ssm_parameter_by_name_success():
```
ksugden, line 11: Should we have an unhappy path test for the _by_name function?

ksugden (approved):
Apart from potentially missing test that I commented on, looks good, so I leave up to you if you want to add a test for that

tests/platform_helper/providers/test_parameter_store.py
DeveloperConnor: Yes, ill add one - good spot

DeveloperConnor: There isn't really an unhappy path in this scenario besides boto3 failing which we don't want to be testing anyway since there is nothing in the code to capture that exception.

PR #794, opened 2025-03-04
refactor: DBTP-1846 Create KMS provider

ddwilliams74:
Addresses [DBTP-1846 ](https://uktrade.atlassian.net/browse/DBTP-1846 )

Introduction of new kms provider and tests with refactor of exiting copilot command and tests. To enable the command module to work, the copilot domain has also been modified to inject the new kms provider, but it's not being used currently. The copilot domain refactor will be tackle this.

---
### Title:
### Description:
~- [ ] If the work includes user interface changes, before and after screenshots included in description~
~- [ ] Includes any applicable changes to the documentation in this code base~
~- [ ] Includes link(s) to any applicable changes to the documentation in the [DBT Platform Documentation](https://platform.readme.trade.gov.uk/) (can be to a pull request)~
### Tasks:


dbt_platform_helper/providers/kms.py
```diff
@@ -0,0 +1,24 @@
...
+        Retrieves metadata about a KMS key using its alias.
+
+        Args:
+            alias_name (str): The alias name of the KMS key.
+
+        Returns:
+            dict: A dictionary containing metadata about the specified KMS key.
+        """
+        # The kms client can take an alias name as the KeyId
+        response = self.kms_client.describe_key(KeyId=alias_name)
```
DeveloperConnor, line 22:
```suggestion
        return self.kms_client.describe_key(KeyId=alias_name)
```

dbt_platform_helper/providers/kms.py
```diff
@@ -0,0 +1,24 @@
...
+
+        return response
```
DeveloperConnor, line 24:
```suggestion
```

ddwilliams74: Good point. I think that was left from previously interrogating the response and raising an exception (which we're now not doing). Makes sense to simply return.

DeveloperConnor (approved):

PR #798, opened 2025-03-05
refactor: Move PlatformHelperVersioning mocks to a class

ksugden:
- Follows the pattern adopted in other domain level tests where we create a class of default mocks that can be overwritten in the individual tests.
- renames local -> installed and provider LocalVersionProvider -> InstalledVersionProvider for clarity
- replaces injected function `skip_versioning_checks` with a boolean flag

---
### Title:
### Description:
### Tasks:


tests/platform_helper/domain/test_copilot.py
```diff
@@ -403,6 +403,7 @@ def test_make_addons_removes_old_addons_files(
             path = Path("copilot", f)
             assert not path.exists(), f"{path} should not exist"
 
+
```
ksugden, line 406:
```suggestion
```

dbt_platform_helper/domain/versioning.py
```diff
@@ -40,16 +39,16 @@ def __init__(
         ),
         config_provider: ConfigProvider = ConfigProvider(),
         pypi_provider: PyPiVersionProvider = PyPiVersionProvider,
-        local_version_provider: LocalVersionProvider = LocalVersionProvider(),
-        skip_versioning_checks: Callable[[], bool] = None,
+        installed_version_provider: InstalledVersionProvider = InstalledVersionProvider(),
+        skip_versioning_checks: bool = None,
```
a-gleeson, line 43: Should the default be the boolean `False`?

a-gleeson (approved):
A couple comments but LGTM!

dbt_platform_helper/domain/versioning.py
ksugden: the default needs to be calculated according to whether the command is running from an installed package, or whether it's running locally directly from the python script, plus the value of an env variable, so I think it needs to be None and set underneath in the __init__ by calling the function

dbt_platform_helper/providers/version.py
```diff
@@ -28,7 +28,7 @@ class VersionProvider(ABC):
     pass
 
 
-class LocalVersionProvider:
+class InstalledVersionProvider:
```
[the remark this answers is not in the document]
ksugden: no.... yeah, maybe we should drop the Provider suffix.  Maybe as a refactoring marathon wrap up task?

PR #799, opened 2025-03-06
refactor: Config command refactor (DBTP-1538)

a-gleeson:
Addresses [DBTP-1538](https://uktrade.atlassian.net/browse/DBTP-1538)

---
### Title:
### Description:
### Tasks:

[DBTP-1538]: https://uktrade.atlassian.net/browse/DBTP-1538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dbt_platform_helper/domain/config.py
```diff
@@ -0,0 +1,39 @@
...
+            version_file_version_provider=YamlFileProvider  # Overrides DeprecatedVersionFileVersionProvider wrapper
+        ),
+    ):
+        self.io = io
+        self.platform_helper_versioning_domain = platform_helper_versioning_domain
+
+    def validate(self):
+        if Path("copilot").exists():
+            self.io.debug("\nDetected a deployment repository\n")
+            platform_helper_version_status = (
```
ksugden, line 29: maybe getting the status, validating it and printing the errors could be made into a public method on `PlatformHelperVersioning` (if it isn't already!)

ksugden: since there's no reason to expose 'status' here - all we're interested in is 'is my versioning config valid'

ksugden: `PlatformHelperVersioning().print_validation_messages()` or some such

a-gleeson: mmm, that could be nice! A good one to discuss further

dbt_platform_helper/providers/version.py
```diff
@@ -76,3 +79,36 @@ def get_required_version(self) -> SemanticVersion:
         except FileProviderException:
             version_from_file = None
         return version_from_file
+
+
+class AWSVersionProvider(VersionProvider):
```
ksugden, line 84: I think there's a bit of inconsistency in the naming and placing of these classes, with the existing `PlatformHelperVersionStatus` and `PlatformHelperVersioning` concepts.  The `Versioning` classes were supposed to be responsible for resolving the version status of a tool (and doing validations).  Whereas the `VersionProvider` classes are responsible for getting the version from a source such as Github, the installed version, or PyPi.  I think to fit with this pattern, this class would be `class AWSVersioning` with a `(_)get_version_status` method returning the `VersionStatus`.  Happy to chat about it.  Could think about following up this PR with some alignment.

dbt_platform_helper/providers/version.py
```diff
@@ -76,3 +79,36 @@ def get_required_version(self) -> SemanticVersion:
...
+    @staticmethod
+    def get_versions() -> VersionStatus:
```
ksugden, line 86: to align, we would change this function name and the associated variables to `get_version_status`

tests/platform_helper/providers/test_version.py
```diff
@@ -69,3 +71,32 @@ def test_get_latest_version(self, request_get):
...
+class TestCopilotVersionProvider:
+    @patch("subprocess.run")
+    @patch(
+        "dbt_platform_helper.providers.version.GithubVersionProvider.get_latest_version",
+        return_value=SemanticVersion(2, 0, 0),
+    )
+    def test_copilot_versions(self, mock_get_github_released_version, mock_run):
+        mock_run.return_value.stdout = b"1.0.0"
+
+        versions = CopilotVersionProvider.get_versions()
```
ksugden, line 99: This would be `CopilotVersioning(github_verison=GithubVersionProvider).get_verison_status()` (injected github provider, without patch)

tests/platform_helper/test_command_config.py
```diff
@@ -1,235 +1,65 @@
...
-@patch(
-    "dbt_platform_helper.utils.tool_versioning.get_aws_versions",
-    return_value=VersionStatus(SemanticVersion(0, 1, 0), SemanticVersion(1, 0, 0)),
-)
-@patch(
-    "dbt_platform_helper.utils.tool_versioning.get_copilot_versions",
-    return_value=VersionStatus(SemanticVersion(0, 1, 0), SemanticVersion(1, 0, 0)),
+    assert result.exit_code == 0
+    mock_config_domain_instance.validate.assert_called()
+    mock_config_domain.assert_called()
```
ksugden, line 20: is this assertion redundant given the above assertion? (e.g. we have to have an instance if it's been called with `validate()`). I do remember us talking about checking the constructor was called correctly though.... maybe we need to check the arguments of the constructor to make the assertion valuable?

tests/platform_helper/test_command_config.py
```diff
@@ -1,235 +1,65 @@
...
+@patch("dbt_platform_helper.commands.config.get_aws_session_or_abort")
+@patch("dbt_platform_helper.commands.config.Config")
+def test_command_aws_success(mock_config_domain, mock_session, cli_args, called_with):
+    mock_config_domain_instance = mock_config_domain.return_value
+
+    runner = CliRunner()
+    result = runner.invoke(aws, cli_args)
+    mock_config_domain_instance.generate_aws.assert_called_with(called_with)
+
+    mock_config_domain.assert_called()
```
ksugden, line 52: as above

tests/platform_helper/test_command_config.py
```diff
@@ -1,235 +1,65 @@
...
+@patch("dbt_platform_helper.commands.config.get_aws_session_or_abort")
+@patch("dbt_platform_helper.commands.config.Config")
+def test_command_aws_raises_platform_exception(mock_config_domain, mock_session):
+    mock_config_domain_instance = mock_config_domain.return_value
+    mock_config_domain_instance.generate_aws.side_effect = PlatformException("i've failed")
+
+    runner = CliRunner()
+    result = runner.invoke(aws)
+
+    assert result.exit_code == 1
```
ksugden, line 65: we sometimes patch click.secho and assert on the error message in these tests.  On the other hand, sometimes we don't...

tests/platform_helper/utils/test_tool_versioning.py
```diff
@@ -1,109 +1,9 @@
...
-    assert result.installed == SemanticVersion(1, 1, 1)
-    assert result.latest == SemanticVersion(2, 2, 2)
-    assert result.deprecated_version_file == None
-    assert result.platform_config_default == SemanticVersion(3, 3, 3)
-    assert result.pipeline_overrides["main"] == str(SemanticVersion(4, 4, 4))
-
-
-# TODO Relocate when we refactor config command in DBTP-1538
-class TestConfigCommand:
-    @pytest.mark.parametrize(
```
ksugden, line 61: couldn't see where these tests have moved to....

dbt_platform_helper/domain/config.py
```diff
@@ -0,0 +1,352 @@
...
+"""
+
+
+class NoDeploymentRepoConfigException(PlatformException):
+    def __init__(self):
+        super().__init__("Could not find a deployment repository, no checks to run.")
+
+
+# TODO move to generic location so it can be reused
+class NoPlatformConfigException(PlatformException):
```
ksugden, line 65: I'm kind of confused why we care if there's a config file, since we don't seem to use the platform-config in the rest of the domain??  Is it just a proxy for being in the right directory?

ksugden, line 65: I find this exception kind of confusing as it's not clear why we care if there's a platform-config file at this point?  Is it a proxy for checking we're in the right directory?  Since there's no config provider injected into the domain, it's not clear what needs to config file. I think it's happening with the default arguments of the platform helper versioning domain... I am wondering if that means it we should inject the config provider, do the check from there and inject it into the platform helper versioning class.  I would think it's expected that the Config domain has an explicit dependency on the config provider. hmmm.....

dbt_platform_helper/providers/version.py
a-gleeson: That is a good point I don't think this should necessarily inherit from `VersionProvider`, although I currently think it lives in this provider. It calls the github provider to get the latest version but mainly it is to get the locally installed version. I feel like this does something which can live in this provider although in a different format.

tests/platform_helper/test_command_config.py
a-gleeson: we will have a play about with this

dbt_platform_helper/domain/config.py
a-gleeson: So this isn't exactly a change in functionality as previously `ConfigProvider.config_file_check()` carried out a check half way through the function (after the tools table but before the make addons table). So if the platform-config wasn't present it would abort. I thought the function for the check was redundant but also that it should be called earlier.

tests/platform_helper/utils/test_tool_versioning.py
a-gleeson: `get_versions` got moved but to simplify I removed `validate_template_version` and it is tested within the config domain. Do you think I should explicitly test these scenarios?

rav-pradhan: It seems like this test doesn't add any value as other tests already cover the `SemanticVersion` [Codecov Semantic Version](https://app.codecov.io/gh/uktrade/platform-tools/blob/DBTP-1538--config-command-refactor/dbt_platform_helper%2Fproviders%2Fsemantic_version.py)

The config domain test, [`test_validate_outdated`](https://github.com/uktrade/platform-tools/blob/3105d17076683f59b54595919f31567da9c74061/tests/platform_helper/domain/test_config.py#L310), implicitly covers the raising of these exceptions

dbt_platform_helper/providers/version.py
rav-pradhan: *ACTION*: to make a sub-task on this ticket to discuss how to best define the `Version` and `Versioning` concepts we currently have, and any follow-on implementation

dbt_platform_helper/domain/config.py
ksugden: captured in ticket [DBTP-1895](https://uktrade.atlassian.net/browse/DBTP-1895)

tests/platform_helper/test_command_config.py
ksugden: I've left a note in the `consolidate refactor` to decide whether we should always be doing click.secho mock asserts at command level

ksugden (approved):

PR #800, opened 2025-03-06
refactor: DBTP-1849 - outputting human readable error when SSO token expired when running "environment generate-terraform"

chiaramapellimt:
Addresses [DBTP-1849](https://uktrade.atlassian.net/browse/DBTP-1849)

[DBTP-1849]: https://uktrade.atlassian.net/browse/DBTP-1849?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUi
Screenshot 2025-03-06 at 14 10 16

### DIFF
--- a/dbt_platform_helper/commands/environment.py
+++ b/dbt_platform_helper/commands/environment.py
@@ -94,6 +94,7 @@ def generate(name):
 def generate_terraform(name, terraform_platform_modules_version):
     click_io = ClickIOProvider()
     try:
+        # TODO = pass the session to ConfigValidator
         get_aws_session_or_abort()
         config_provider = ConfigProvider(ConfigValidator())
         TerraformEnvironment(config_provider).generate(name, terraform_platform_modules_version)
--- a/dbt_platform_helper/commands/environment.py
+++ b/dbt_platform_helper/commands/environment.py
@@ -92,9 +92,10 @@ def generate(name):
     help=f"Override the default version of terraform-platform-modules. (Default version is '{DEFAULT_TERRAFORM_PLATFORM_MODULES_VERSION}').",
 )
 def generate_terraform(name, terraform_platform_modules_version):
-
+    click_io = ClickIOProvider()
     try:
+        get_aws_session_or_abort()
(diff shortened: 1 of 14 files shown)


tests/platform_helper/test_command_environment.py
```diff
@@ -246,3 +250,22 @@ def test_generate_terraform_catches_platform_exception_and_exits(
         assert result.exit_code == 1
         mock_click.assert_called_with("""Error: i've failed""", err=True, fg="red")
         mock_terraform_environment_instance.generate.assert_called_with("test", "123")
+
+    @patch("dbt_platform_helper.commands.environment.TerraformEnvironment")
+    @patch("click.secho")
+    def test_generate_terraform_sso_token_expired(self, mock_click, terraform_environment_mock):
```
ksugden, line 256: we'd probably want to patch `get_aws_session_or_abort` with a sideeffect here so it's not really trying to open a session?

dbt_platform_helper/commands/environment.py
```diff
@@ -92,9 +92,10 @@ def generate(name):
     help=f"Override the default version of terraform-platform-modules. (Default version is '{DEFAULT_TERRAFORM_PLATFORM_MODULES_VERSION}').",
 )
 def generate_terraform(name, terraform_platform_modules_version):
-
+    click_io = ClickIOProvider()
     try:
+        get_aws_session_or_abort()
```
ksugden, line 97: I guess the next step would be for the session to be passed into the `ConfigValidator` (if that's the place where session is needed) and used to create the boto3 clients

tests/platform_helper/test_command_environment.py
chiaramapellimt: if I patch it then it bypasses it no? as in it thinks it's all good so it carries on, when actually I want to simulate the abort ASAP (so that when we call generate it doesn't carry on)

chiaramapellimt: I've edited now, probably best this way.

a-gleeson (approved):

PR #807, opened 2025-03-14
feat: Allow manual releases based on image reference (DBTP-1165)

irenegrandeh:
Addresses [DBTP-1165](https://uktrade.atlassian.net/browse/DBTP-1165)
Related PR: https://github.com/uktrade/platform-documentation/pull/536

---
### Title:
- [x] Scope included as per [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [x] Ticket reference included (unless it's a quick out of ticket thing)
### Description:
- [x] Link to ticket included (unless it's a quick out of ticket thing)
- [x] Includes tests (or an explanation for why it doesn't)
- [ ] If the work includes user interface changes, before and after screenshots included in description
- [x] Includes any applicable changes to the documentation in this code base
- [x] Includes link(s) to any applicable changes to the documentation in the [DBT Platform Documentation](https://platform.readme.trade.gov.uk/) (can be to a pull request)
### Tasks:
- [ ] [Run the end to end tests for this branch]([https://github.com/uktrade/platform-tools?tab=readme-ov-file#regression-tests](https://github.com/uktrade/platform-end-to-end-tests?tab=readme-ov-file#running-the-tests)) and confirm that they are passing

[DBTP-1165]: https://uktrade.atlassian.net/browse/DBTP-1165?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #809, opened 2025-03-14
feat: Allow `codebase build` to be run outside the application repository (off-ticket)

CamLamb:
Removes they check for the commit existing in the local git repository, thereby allowing the command to be run from anywhere, e.g. from the deploy repository codebase.

---
### Title:
- `[ ] Ticket reference included (unless it's a quick out of ticket thing)`
### Description:
- `[ ] Link to ticket included (unless it's a quick out of ticket thing)`
- `[ ] Includes link(s) to any applicable changes to the documentation in the [DBT Platform Documentation](https://platform.readme.trade.gov.uk/) (can be to a pull request)`
### Tasks:

[End to end tests run](https://763451185160-2qjkzelg.eu-west-2.console.aws.amazon.com/codesuite/codepipeline/pipelines/pull-request-end-to-end-tests/executions/9319f484-0b10-48aa-a9ec-a43a5976a173/visualization?region=eu-west-2&tab=variables).


CamLamb:
> I think there may be overlap with #807.

Ahh there might be!
In that case I don't mind this PR being closed 😄 

Although I do prefer the approach in this PR for the interface between the user and the commands being `--commit`, `--tag` and `--branch` instead of them having to know the image reference.
They can be told about the image reference that has been generated and for the purpose of the deploy maybe provide the ability to provide commit/tag/branch OR the image reference.

WillGibson:
@CamLamb The `--ref` approach we are taking in @irenegrandeh's pull request handles it all with less code/options to maintain. Branch, tag and commit hashes are all Git references, so keeping to single parameter makes sense.

#807 does not include the build command though, so that might still be worth adding, but the interface for your examples would just be...

```shell
platform-helper codebase build ... --ref ab1c23d
platform-helper codebase build … --ref 1.2.3
platform-helper codebase build … --ref TIK-123-feature-name
```

In terms of cognitive load working out what you pass in...

```shell
platform-helper codebase deploy … --ref tag-1.2.3
```

...is no more thinking than...

```shell
platform-helper codebase deploy … --tag 1.2.3
```

I think you, @irenegrandeh, and whoever else is relevant should have a chat.

I will also write some contributing guidance for this codebase, it's brilliant to have willing contributors from outside team, but changes need to be discussed and agreed before implementation to avoid additional overhead or people working on something that does not align to plans for DBT Platform.

antroy-madetech (approved):

PR #823, opened 2025-03-24
feat: Add new field for Datadog ticket (DBTP-1879)

chiaramapellimt:
Addresses https://uktrade.atlassian.net/browse/DBTP-1879.


dbt_platform_helper/providers/platform_config_schema.py
```diff
@@ -434,6 +436,22 @@ def valid_s3_bucket_name(name: str):
...
+    @staticmethod
+    def __datadog_schema() -> dict:
+        return {
+            "type": "datadog",
+            Optional("environments"): {
+                PlatformConfigSchema.__valid_environment_name(): {
+                    Optional("team_name"): str,
+                    Optional("contact_name"): str,
+                    Optional("contact_email"): str,
+                    Optional("repository"): str,
```
WillGibson, line 448: Might be worth being clear that this is referring to the codebase repository and that it's a URL, not just `uktrade/<application>`?
```suggestion
                    Optional("codebase_repository_url"): str,
```

dbt_platform_helper/providers/platform_config_schema.py
```diff
@@ -434,6 +436,22 @@ def valid_s3_bucket_name(name: str):
...
+                    Optional("docs"): str,
```
WillGibson, line 449: Maybe...
```suggestion
                    Optional("codebase_documentation_url"): str,
```

dbt_platform_helper/providers/platform_config_schema.py
```diff
@@ -434,6 +436,22 @@ def valid_s3_bucket_name(name: str):
...
+                    Optional("services_to_monitor"): list,
```
WillGibson, line 450: If the services are running off different codebases, would there be multiple Datadog extensions?

tests/platform_helper/utils/fixtures/addons_files/datadog_addons_bad_data.yml
```diff
@@ -0,0 +1,40 @@
+my-datadog-bad-key:
+  type: datadog
+  bad_key:
+
```
WillGibson, line 4:
```suggestion
```

dbt_platform_helper/providers/platform_config_schema.py
ejayesh: Actually we can remove `Optional("repository"): str,` from here as we pull that in from the ci-image builder.
https://github.com/uktrade/ci-image-builder/blob/a63976b13865f2dbcade669590531710f1565adf/image_builder/pack.py#L152

WillGibson: Every repository should have a README, even if it links out to documentation elsewhere.

Would is be fair to suggest that we don't need the docs option too?

dbt_platform_helper/providers/platform_config_schema.py
WillGibson: Every repository should have a README, even if it links out to documentation elsewhere.

Would is be fair to suggest that we don't need the docs option here?

ejayesh: This field is more for generic documentation about the application.  Maybe pointing to a confluence page or a run book.

dbt_platform_helper/providers/platform_config_schema.py
```diff
@@ -434,6 +436,21 @@ def valid_s3_bucket_name(name: str):
...
+    @staticmethod
+    def __datadog_schema() -> dict:
+        return {
+            "type": "datadog",
+            Optional("environments"): {
+                PlatformConfigSchema.__valid_environment_name(): {
+                    Optional("team_name"): str,
+                    Optional("contact_name"): str,
+                    Optional("contact_email"): str,
+                    Optional("documentation_url"): str,
```
WillGibson, line 448: https://github.com/uktrade/terraform-platform-modules/pull/373/files#r2016979345

chiaramapellimt: Just to confirm, shall I change this to "documentation" instead of "documentation_url"?

WillGibson: This field should stay as is.

dbt_platform_helper/providers/platform_config_schema.py
```diff
@@ -434,6 +436,21 @@ def valid_s3_bucket_name(name: str):
...
+                    Optional("services_to_monitor"): list,
+                }
```
WillGibson, line 450: Are the properties really optional?

Can we do something like this?

```suggestion
                Optional(PlatformConfigSchema.__valid_environment_name()): {
                    "team_name": str,
                    "contact_name": str,
                    "contact_email": str,
                    "documentation_url": str,
                    "services_to_monitor": list,
                }
```

This might apply elsewhere, but it would be good to see if it is possible with this one.

Really, we need at least "*" then other environments are optional, but I think that can wait for the outcome of [DBTP-1708 Should we use Pydantic for schema validation?](https://uktrade.atlassian.net/browse/DBTP-1708).

dbt_platform_helper/providers/platform_config_schema.py
```diff
@@ -28,6 +28,7 @@ def schema() -> Schema:
                         PlatformConfigSchema.__monitoring_schema(),
                         PlatformConfigSchema.__opensearch_schema(),
                         PlatformConfigSchema.__postgres_schema(),
+                        Optional(PlatformConfigSchema.__datadog_schema()),
```
JohnStainsby, line 31: I don't think this needs to be optional the `str: Or(` above means an extension has to be one of these types but not that any of them are required

dbt_platform_helper/providers/platform_config_schema.py
WillGibson: @chiaramapellimt This was resolved without explanation. What's the thinking?

dbt_platform_helper/providers/platform_config_schema.py
WillGibson: This was removed.

WillGibson (approved):
Looks good to me now subject to it having been tested with `terraform-platform-modules`.

dbt_platform_helper/providers/platform_config_schema.py
chiaramapellimt: (I did)

PR #826, opened 2025-03-25
fix: Intermittent failures when calling via requests library (off-ticket)

a-gleeson:
---
## Why:
When calling via `requests` to github we get Intermittent failures within the end to end tests

## What:
- add retry to requests
- on exception log the error and return `None` for the version,  so usage doesn't break.

---
### Title:

### Description:
- ~[ ] Link to ticket included (unless it's a quick out of ticket thing)~

### Tasks:

### DIFF
--- a/dbt_platform_helper/providers/version.py
+++ b/dbt_platform_helper/providers/version.py
@@ -10,6 +10,7 @@ import requests
 
 from dbt_platform_helper.constants import PLATFORM_HELPER_VERSION_FILE
 from dbt_platform_helper.platform_exception import PlatformException
+from dbt_platform_helper.providers.io import ClickIOProvider
 from dbt_platform_helper.providers.semantic_version import SemanticVersion
 from dbt_platform_helper.providers.yaml_file import FileProviderException
 from dbt_platform_helper.providers.yaml_file import YamlFileProvider
@@ -49,6 +50,7 @@ class GithubLatestVersionProvider(VersionProvider):
     def get_semantic_version(repo_name: str, tags: bool = False) -> SemanticVersion:
         if tags:
             tags_list = requests.get(f"https://api.github.com/repos/{repo_name}/tags").json()
+            ClickIOProvider().info(tags_list)
             versions = [SemanticVersion.from_string(v["name"]) for v in tags_list]
             versions.sort(reverse=True)
             return versions[0]
--- a/dbt_platform_helper/providers/version.py
+++ b/dbt_platform_helper/providers/version.py
@@ -21,7 +21,7 @@ from dbt_platform_helper.providers.yaml_file import YamlFileProvider
 
 def set_up_retry():
     session = Session()
-    retries = Retry(total=5, backoff_factor=0.5, status_forcelist=[403, 500, 502, 503, 504])
+    retries = Retry(total=5, backoff_factor=0.1, status_forcelist=[403, 500, 502, 503, 504])
     session.mount("https://", HTTPAdapter(max_retries=retries))
     return session
 
--- a/dbt_platform_helper/providers/version.py
+++ b/dbt_platform_helper/providers/version.py
@@ -6,7 +6,9 @@ from importlib.metadata import PackageNotFoundError
 from importlib.metadata import version
 from pathlib import Path
 
-import requests
+from requests import Session
+from requests.adapters import HTTPAdapter
+from urllib3.util import Retry
 
 from dbt_platform_helper.constants import PLATFORM_HELPER_VERSION_FILE
 from dbt_platform_helper.platform_exception import PlatformException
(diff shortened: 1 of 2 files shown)


dbt_platform_helper/providers/version.py
```diff
@@ -5,16 +5,27 @@
...
 from dbt_platform_helper.platform_exception import PlatformException
+from dbt_platform_helper.providers.io import ClickIOProvider
 from dbt_platform_helper.providers.semantic_version import SemanticVersion
 from dbt_platform_helper.providers.yaml_file import FileProviderException
 from dbt_platform_helper.providers.yaml_file import YamlFileProvider
 
 
+def set_up_retry():
+    session = Session()
+    retries = Retry(total=5, backoff_factor=0.5, status_forcelist=[403, 500, 502, 503, 504])
```
ksugden, line 24: do you expect this to hang for some time if it has to retry 5 times before failing?  Are timeouts built into this session?

ksugden (approved):

dbt_platform_helper/providers/version.py
a-gleeson: reduced backoff to `0.1`

dbt_platform_helper/providers/version.py
```diff
@@ -42,38 +53,53 @@ def get_semantic_version(tool_name: str) -> SemanticVersion:
...
-# TODO Alternatively use the gitpython package?
 class GithubLatestVersionProvider(VersionProvider):
     @staticmethod
-    def get_semantic_version(repo_name: str, tags: bool = False) -> SemanticVersion:
-        if tags:
-            tags_list = requests.get(f"https://api.github.com/repos/{repo_name}/tags").json()
-            versions = [SemanticVersion.from_string(v["name"]) for v in tags_list]
-            versions.sort(reverse=True)
-            return versions[0]
+    def get_semantic_version(
```
WillGibson, line 58: Does this better describe the intent?
```suggestion
    def get_latest_semantic_version(
```

dbt_platform_helper/providers/version.py
```diff
@@ -42,38 +53,53 @@ def get_semantic_version(tool_name: str) -> SemanticVersion:
...
 
 class PyPiLatestVersionProvider(VersionProvider):
     @staticmethod
-    def get_semantic_version(project_name: str) -> SemanticVersion:
-        package_info = requests.get(f"https://pypi.org/pypi/{project_name}/json").json()
-        released_versions = package_info["releases"].keys()
-        parsed_released_versions = [SemanticVersion.from_string(v) for v in released_versions]
-        parsed_released_versions.sort(reverse=True)
-        return parsed_released_versions[0]
+    def get_semantic_version(
```
WillGibson, line 83:
```suggestion
    def get_latest_semantic_version(
```

dbt_platform_helper/providers/version.py
```diff
@@ -42,38 +53,53 @@ def get_semantic_version(tool_name: str) -> SemanticVersion:
...
+        semantic_version = None
+        try:
+            if tags:
+                response = request_session.get(f"https://api.github.com/repos/{repo_name}/tags")
 
-        package_info = requests.get(
-            f"https://api.github.com/repos/{repo_name}/releases/latest"
-        ).json()
-        return SemanticVersion.from_string(package_info["tag_name"])
+                versions = [SemanticVersion.from_string(v["name"]) for v in response.json()]
```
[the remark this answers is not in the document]
a-gleeson: mmm good question. I don't think that is why we are getting an error right now but I can test the class against some repositories with this. 
Currently it will not cause the command to fail but the results would be `None`.
Maybe another ticket

dbt_platform_helper/providers/version.py
WillGibson: Ignore, I see that's covered in the class name.

dbt_platform_helper/providers/version.py
WillGibson: Ignore, I see that's covered in the class name.

dbt_platform_helper/providers/version.py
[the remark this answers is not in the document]
WillGibson: The method is explicitly getting the semantic version. If we need to do something else, I think that would be out of the scope of this pull request.

WillGibson (approved):

dbt_platform_helper/providers/version.py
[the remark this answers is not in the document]
ksugden: This wasn't a change in functionality, as we always expected the request to return something that could be parsed into a semantic version, otherwise we return -1 in the sem-ver positions (wisnae me!)

dbt_platform_helper/providers/version.py
a-gleeson: I think I will leave it as get_semantic_version as this is a common function used between Version providers which also cover installed versions which wouldn't make sense. I think the fact the Provider classes have `Latest` in them should be enough?

dbt_platform_helper/providers/version.py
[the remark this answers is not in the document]
a-gleeson: For the ticket I write, I guess it will cover the case that not all tags contain the semantic version so we need to be able to handle the fact that not all tags are in the same format. Is that understanding correct?

PR #875, opened 2025-05-16
fix: ECR repo tag mismatch (DBTP-2071)

antroy-madetech:
Addresses https://uktrade.atlassian.net/browse/DBTP-2071

Fixes a bug where we were truncating commit hashes when doing a `platform-helper codebase deploy --commit` to 7 characters. For some codebase repos the short sha hash will be longer than 7 characters (e.g. in repos with a lot of commits such as datahub)

Instead we are:
- pulling down a list of all images in the repo
- using an exact match if available
- comparing each image to see if it forma a prefix for the passed in commit hash and using that if there is just one.

---
### Title:
- [ ] Conforms to [our pull request title guidance](https://uktrade.atlassian.net/wiki/spaces/DBTP/pages/4402020487/Git+housekeeping#Pull-request-titles). E.g. `feat: Add new feature (DBTP-1234)` or `chore: Correct typo (off-ticket)`

### Description:

### Tasks:


dbt_platform_helper/providers/ecr.py
```diff
@@ -1,12 +1,18 @@
...
+from dbt_platform_helper.providers.aws.exceptions import AWSException
 from dbt_platform_helper.providers.aws.exceptions import ImageNotFoundException
+from dbt_platform_helper.providers.aws.exceptions import MultipleImagesFoundException
 from dbt_platform_helper.providers.aws.exceptions import RepositoryNotFoundException
 from dbt_platform_helper.providers.io import ClickIOProvider
-from dbt_platform_helper.utils.application import Application
 from dbt_platform_helper.utils.aws import get_aws_session_or_abort
 
+NOT_A_UNIQUE_TAG_INFO = 'INFO: The tag "{image_ref}" is not a unique, commit-specific tag. Deploying the corresponding commit tag "{commit_tag}" instead.'
+NO_ASSOCIATED_COMMIT_TAG_WARNING = 'WARNING: The AWS ECR image "{image_ref}" has no associated commit tag so deploying "{image_ref}". Note this could result in images with unintended or incompatible changes being deployed if new ECS Tasks for your service.'
```
ksugden, line 14:
```suggestion
NO_ASSOCIATED_COMMIT_TAG_WARNING = 'WARNING: The AWS ECR image "{image_ref}" has no associated commit tag so deploying "{image_ref}". Note this could result in images with unintended or incompatible changes being deployed in new ECS Tasks for your service.'
```

tests/platform_helper/providers/test_ecr.py
```diff
@@ -97,150 +113,158 @@ def two_pages_of_describe_repository_data():
...
+            "SomeOtherException",
+            AWSException,
+            "Unexpected error for repo 'test_app/test_codebase' and image reference 'commit-abc123': "
+            "An error occurred (SomeOtherException) when calling the ListImages operation: Unknown",
+        ),
     ],
 )
-def test_find_commit_tag_handles_malformed_image_details(image_details):
-    ecr = ECRProvider(Mock(), Mock())
+def test_get_commit_tag_for_reference_recasts_exceptions_as_platform_exceptions(
```
ksugden, line 251: Just wondering given the test name and intention, should we catch the exception as a `PlatformException` instead of using the 'expected_exception' parameter?

tests/platform_helper/providers/test_ecr.py
```diff
@@ -97,150 +113,158 @@ def two_pages_of_describe_repository_data():
...
+    ecr_provider = ECRProvider(**mocks.params())
 
-    actual_tag = ecr.find_commit_tag(image_details, "tag-1.2.3")
+    with pytest.raises(MultipleImagesFoundException) as ex:
+        ecr_provider.get_commit_tag_for_reference("test_app", "test_codebase", "commit-deadbea7")
 
-    ecr.click_io.warn.assert_called_once_with(
-        'WARNING: The AWS ECR image "tag-1.2.3" has no associated commit tag so deploying "tag-1.2.3". Note this could result in images with unintended or incompatible changes being deployed if new ECS Tasks for your service.'
+    actual_error = str(ex.value)
+    expected_error = MULTIPLE_IMAGES_FOUND_TEMPLATE.format(
```
ksugden, line 228: I'll come back to this after standup, but there's something I'm finding confusing about this error message - it seems back-to-front to say `commit-deadbea7`, matches `commit-dea`, `commit-deadb`,`commit-deadbea`.  I would expect e.g. `commit-dea` matches `commit-deadbea7`, `commit-dea`, `commit-deadb`, `commit-deadbea`.  I will go back and check what the `image_ref` should be though.... could be that the exception message should be taking the `tag` instead of `image_ref`; or could be I need to re-read it and understand it properly!

ksugden: I think your demo clarified this - `commit-deadbeat7` is passed in as the (shortened) commit sha.  Any of the truncated image tags `commit-dea`, `commit-deadb`, `commit-deadbea` could potentially be derived from `commit-deadbeat7` so are considered matches

dbt_platform_helper/providers/aws/exceptions.py
```diff
@@ -12,16 +12,34 @@ def __init__(self, addon_name: str, application_name: str, environment: str):
...
+IMAGE_NOT_FOUND_TEMPLATE = """An image labelled "{image_ref}" could not be found in your image repository. Try the `platform-helper codebase build` command first."""
+
+
 class ImageNotFoundException(AWSException):
     def __init__(self, image_ref: str):
+        super().__init__(IMAGE_NOT_FOUND_TEMPLATE.format(image_ref=image_ref))
+
+
+MULTIPLE_IMAGES_FOUND_TEMPLATE = (
+    'Image reference "{image_ref}" matched the following images: {matching_images}'
```
ksugden, line 24:
```suggestion
    'Image reference "{image_ref}" is matched by the following images: {matching_images}'
```

ksugden: would resolve my "back-to-front" feeling :P

tests/platform_helper/providers/test_ecr.py
antroy-madetech: How about renaming the tests as:
```suggestion
def test_get_commit_tag_for_reference_recasts_exceptions_as_more_specific_exceptions(
```

tests/platform_helper/providers/test_ecr.py
antroy-madetech: Do you want me to implement the opposite as well? The functionality is in there to throw an error if there are multiple matches, so shouldn't be dangerous. i.e. if they pass in commit-abc123 it would match commit-abc1234 as well as commit-ab

dbt_platform_helper/providers/aws/exceptions.py
antroy-madetech: Yes. Or maybe if I implement the opposite behaviour as well, it could be:
```suggestion
    'The following images are possible matches for image reference "{image_ref}": {matching_images}'
```

tests/platform_helper/providers/test_ecr.py
antroy-madetech: I think for me the danger is this scenario:

- Current main branch is on commit sha ab61fa91
- User has got used to the shorthand form, so runs ph codebase deploy --commit ab
- The main branch has not yet been built into an image
- ECR happens to have an image commit-abc123ef
- commit-abc123ef will match and be deployed.

The opposite *could* have this same issue, but I think it's less likely as it would need to match far more characters (at least 7). TBH Adrian is probably right in that we should just name the images with the full sha hash (perhaps in addition to the short hash)

I have the implementation of the shorthand option on a branch, so it's available if we think it is a good idea.

antroy-madetech: Have done this. 7 char minimum, but does the reverse match as well now.

ksugden (approved):

dbt_platform_helper/providers/aws/exceptions.py
ksugden: Yes, that's good

JohnStainsby (approved):

PR #915, opened 2025-07-17
feat: Service generate command (DBTP-2155)

a-gleeson:
Addresses https://uktrade.atlassian.net/browse/DBTP-2155.
---
## What

Adds a `platform-helper service generate` command which generates the service-specific terraform manifest and merged `service-config.yml` (with environment overrides). The latter is used when applying the service terraform and standing up its infrastructure.

- takes an image tag env var or flag to set the image to deploy the service with. Default if neither
- takes a list of environments to generate the service terraform for
- take a list of service to generate the terraform for
- if none are supplied generate for all services and environments
- override the source with env var or set to use local relative path
- use pydantic model to validate the service-config.yml schema 
- use manifest provider to generate terraform manifest

---
### Title:

### Description:

### Tasks:


dbt_platform_helper/utils/deep_merge.py
```diff
@@ -0,0 +1,10 @@
+def deep_merge(base, override):
```
SamBrownMT, line 1: Is this merging two dicts?
If so would [this](https://stackoverflow.com/questions/38987/how-do-i-merge-two-dictionaries-in-a-single-expression-in-python) be a cleaner way?

dragos441: We're merging nested dictionaries, the solution from that article would only apply a shallow merge.

[This article](https://stackoverflow.com/questions/20656135/python-deep-merge-dictionary-data) is the one we used for inspiration.

dbt_platform_helper/constants.py
```diff
@@ -8,13 +10,21 @@
 TERRAFORM_CODEBASE_PIPELINES_MODULE_SOURCE_OVERRIDE_ENV_VAR = (
     "TERRAFORM_CODEBASE_PIPELINES_MODULE_SOURCE_OVERRIDE"
 )
+TERRAFORM_ECS_SERVICE_MODULE_SOURCE_OVERRIDE_ENV_VAR = (
+    "TERRAFORM_ECS_SERVICE_MODULE_SOURCE_OVERRIDE"
+)
+TERRAFORM_MODULE_SOURCE_TYPE_ENV_VAR = "TERRAFORM_MODULE_SOURCE_TYPE"  # "LOCAL", "SSH", "OVERRIDE"
+IMAGE_TAG_ENV_VAR = "IMAGE_TAG"
+IMAGE_TAG_DEFAULT = "tag-latest"
```
JohnStainsby, line 18: I'm not sure about defaulting if a tag isn't provided, at the moment it's a mandatory option on deploy commands. But this will be running in the pipeline where the tag will need to be supplied so it's probably fine

dragos441: Fair enough, might make more sense not to have a default. I've removed it.

If we later choose not to Terraform the ECS task definitions, we won't need to pass image_tag into this command at all.

PR #982, opened 2025-08-08
feat: Configure default providers for pipeline terraform to allow apply only with relevant profile and account_ids (DBPT-2031)

ksugden:
Addresses https://uktrade.atlassian.net/browse/DBTP-2031.

Removes the provider alias so that it is selected as the default provider for

- environment pipeline terraform
- codebase pipeline terraform
- service terraform

The provider is configured with an `allowed_account_id` which is populated with the account being deployed to.  It essentially ensures the account configured in the `platform-config.yml` is consistent with the account configured in the `~/.aws/config` profile and is an extra safeguard to prevent accidentally applying terraform in the wrong account due to misconfigured profiles.

Does not need `shared_credentials_files` argument, since we use the default anyway.  Adding this argument also causes failures in the E2E tests because the codebuild environment cannot find the file and falls back incorrectly to the container credentials, so it is more robust generally without this argument.

Adds tests for the new provider config in the terraform manifests and makes use of a python package `hcl2` which parses terraform and allows us to assert on the content as a dictionary.

---
### Title:

### Description:

### Tasks:


dbt_platform_helper/providers/terraform_manifest.py
```diff
@@ -88,12 +88,13 @@ def generate_codebase_pipeline_config(
         module_source: str,
     ):
         default_account = self._get_account_for_env("*", platform_config)
+        deploy_account_id = self._get_account_id_for_account(default_account, platform_config)
```
ksugden, line 91: I'm wondering about being more specific in the naming and calling this `deploy_to_account_id` (everywhere)

JohnStainsby (approved):

PR #1037, opened 2025-09-05
feat: Internal alb rule (DBTP-2163)

a-gleeson:
Addresses https://uktrade.atlassian.net/browse/DBTP-2163.

---
## What:
- a new command which will create/delete AWS ALB listener rules based on the `service-deployment-mode` parameter
- It will group listener rules into one of the following: platform, maintenance, default, copilot, manual
- The command will fail if there are any "manual" rules
- If the mode is platform (terraform) based then "Platform" rules will be created from priority 10000 for each copilot rule, before creation it will delete any platform rules existing
- If the mode is copilot based then it will delete any platform rules so traffic is directed back to the copilot rules.
- If the commands fail at any point, the operation state is tracked and will attempt to undo any changes made
---
### Title:

### Description:

### Tasks:

### DIFF
--- a/terraform/ecs-service/locals.tf
+++ b/terraform/ecs-service/locals.tf
@@ -1,7 +1,4 @@
 locals {
-
-  secrets      = values(coalesce(var.service_config.secrets, {}))
-  service_name = "${var.application}-${var.environment}-${var.service_config.name}"
   tags = {
     application = var.application
     environment = var.environment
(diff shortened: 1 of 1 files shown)


dbt_platform_helper/domain/update_alb_rules.py
```diff
@@ -0,0 +1,338 @@
...
+        if target_group_arn:
+            tgs = self.load_balancer.get_target_groups_with_tags([target_group_arn])
+
+            # TODO make more robust?
+            for tg in tgs:
+                return tg["Tags"].get("copilot-service", "")
+        else:
+            # TODO add error handling if no target_group_arn
+            self.io.warn("no target group found")
+        return None
```
acodeninja, line 264: Do a bit of tidying here, make sure we either returning a service name or `None`.

dragos441 (approved):

PR #1067, opened 2025-10-07
feat: Custom maintenance mode page for Intranet (off-ticket)

marcelkornblum:
Custom page for intranet to display maintenance mode info. Page only uses embedded content except for a font URL which matches the intranet's URL for that page (so users will mostly have it locally cached in their browsers)


acodeninja (changes requested):
Unfortunately, as the current implementation uses the fixed response action at the ALB, the maximum number of characters in a maintenance page with the current implementation is 1024 (see [API_FixedResponseActionConfig](https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_FixedResponseActionConfig.html)).

 We are looking at S3 origin holding pages, however there are some complexities inherent in this approach, especially in supporting developer bypass of the holding page.

marcelkornblum:
> the current implementation uses the fixed response action at the ALB

@acodeninja does this mean that asset URLs are still available while the page is up? So we could offload SVG and style etc to S3?

(Currently assets are served via the s3proxy ECS service; I guess we'd need a different place to store these?)

adrianclay:
Closing as stale.

PR #1394, opened 2026-04-23
feat: Replacement for copilot job run  (DBTP-2848)

YacquubAdan:
Addresses https://uktrade.atlassian.net/browse/DBTP-2848

---
### Title:
- [X] Conforms to [our pull request title guidance](https://uktrade.atlassian.net/wiki/spaces/DBTP/pages/4402020487/Git+housekeeping#Pull-request-titles). E.g. `feat: Add new feature (DBTP-1234)` or `chore: Correct typo (off-ticket)`

### Description:
- [X] Includes any applicable changes to the documentation in this code base

### Tasks:

## Reviewer Checklist


dbt_platform_helper/domain/job.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class ScheduledJobExecutionFailedException(PlatformException):
+    pass
+
+
+class JobManager:
+
+    SECONDS = 5
+    STATUSES = {"SUCCEEDED", "FAILED", "TIMED_OUT", "ABORTED"}
```
ksugden:
```suggestion
    JOB_FINAL_STATUSES = {"SUCCEEDED", "FAILED", "TIMED_OUT", "ABORTED"}
```

dbt_platform_helper/providers/step_functions.py
```diff
@@ -0,0 +1,88 @@
...
+        if name:
+            kwargs["name"] = name
+
+        try:
+            result = self.sfn_client.start_execution(**kwargs)
+            return result
+
+        except ClientError as err:
+            raise StartExecutionFailedException(
+                state_machine_arn, err.response.get("Error", {}).get("Message", str(err))
```
ksugden, line 23: what's the difference between `str(err)` and `err.response.get("Error", {}).get("Message")`?

dbt_platform_helper/providers/step_functions.py
```diff
@@ -0,0 +1,88 @@
...
+    def _get_status(self, execution_arn: str):
+        response = self.sfn_client.describe_execution(executionArn=execution_arn)
+        return response["status"]
+
+
+class StateMachineNotFoundException(AWSException):
+    def __init__(self, application_name: str, environment: str, job_name: str):
+        super().__init__(
+            f"Scheduled Job '{job_name}' not found in '{environment}' of application '{application_name}'.\n"
+            f"Please check that the combination of application, environment and scheduled job are correct."
```
ksugden:
```suggestion
            f"Please check that the combination of application, environment and name correctly describe a deployed scheduled job."
```

dbt_platform_helper/providers/step_functions.py
```diff
@@ -0,0 +1,88 @@
...
+        )
+
+
+class StartExecutionFailedException(AWSException):
+    def __init__(self, state_machine_arn: str, error: str):
+        super().__init__(
+            f"Failed to start the Scheduled Job execution.\n"
```
ksugden: maybe keep it simple and not provide so much implementation detail -
```suggestion
            f"Failed to start the Scheduled Job execution: {error}"
```

dbt_platform_helper/providers/step_functions.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+class MultipleStateMachinesFoundException(AWSException):
+    def __init__(
+        self, application_name: str, environment: str, job_name: str, state_machine_arns: list[str]
+    ):
+        super().__init__(
+            f"Multiple Jobs {len(state_machine_arns)} with the name '{job_name}' found in '{environment}'\n"
+            f"Expected 1 job\n"
+            f"This usually means that they share the same tags\n"
```
ksugden: Could this actually happen? Does it mean they have somehow deployed the same scheduled job twice?  I feel like this error would have been caught already at an earlier point, like when they tried to deploy the job?

tests/platform_helper/test_command_job.py
```diff
@@ -0,0 +1,88 @@
...
+@patch("dbt_platform_helper.commands.job.load_application")
+def test_job_run_missing_required_arguments(
+    mock_application, mock_step_functions, mock_job_manager_object
+):
+    """Test that job run command requires all arguments."""
+
+    result = CliRunner().invoke(run, [])
+
+    assert result.exit_code != 0
+    assert "Error" in result.output or "Missing option" in result.output
```
ksugden: I'm wondering which one it is here? `"Error" in result.output` or `"Missing option" in result.output`?

dbt_platform_helper/domain/job.py
```diff
@@ -0,0 +1,49 @@
...
+        if follow:
+            self.follow_execution(execution_arn)
+
+    def follow_execution(self, execution_arn: str):
+
+        self.io.info("Waiting for execution to finish...")
+
+        while True:
+            time.sleep(self.SECONDS)
+            status = self.sfn_provider._get_status(execution_arn)
```
ksugden: if we're exposing `_get_status` in the sfn_provider, then it should not be a private method
```suggestion
            status = self.sfn_provider.get_status(execution_arn)
```

dbt_platform_helper/providers/step_functions.py
YacquubAdan: So with `str(err)` were getting the full exception message including AWS boilerplate but with `err.response.get("Error", {}).get("Message")` we're getting just the underlying AWS error message on its own e.g `User is not authorised to perform: states:StartExecution`

dbt_platform_helper/domain/job.py
```diff
@@ -0,0 +1,49 @@
...
+    STATUSES = {"SUCCEEDED", "FAILED", "TIMED_OUT", "ABORTED"}
+
+    def __init__(self, sfn_provider: StepFunctions, io: ClickIOProvider = ClickIOProvider()):
+        self.sfn_provider = sfn_provider
+        self.io = io
+
+    def run(self, app: str, env: str, name: str, follow: bool):
+
+        self.io.info(f"Beginning execution for job '{name}' in {app}/{env}...")
+        state_machine_arn = self.sfn_provider.find_state_machine_arn(name)
```
ksugden: I would prefer if the "state machine" was entirely hidden in this layer.  We could resolve the state machine arn in the sfn_provider and then just expose the `start_execution` function

dbt_platform_helper/domain/job.py
```diff
@@ -0,0 +1,49 @@
...
+
+        response = self.sfn_provider.start_execution(state_machine_arn)
```
ksugden: we could also abstract the AWS response to decouple this layer from AWS

dbt_platform_helper/providers/step_functions.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+class StepFunctions:
+
+    def __init__(self, sfn_client, application_name: str, env: str):
+        self.sfn_client = sfn_client
+        self.application_name = application_name
+        self.env = env
+
+    def find_state_machine_arn(self, job_name: str) -> Optional[str]:
```
ksugden: this should probably be a private function rather than part of the sfn "interface"
```suggestion
    def _find_state_machine_arn(self, job_name: str) -> Optional[str]:
```

dbt_platform_helper/domain/job.py
```diff
@@ -0,0 +1,49 @@
...
+    pass
+
+
+class JobManager:
+
+    SECONDS = 5
+    STATUSES = {"SUCCEEDED", "FAILED", "TIMED_OUT", "ABORTED"}
+
+    def __init__(self, sfn_provider: StepFunctions, io: ClickIOProvider = ClickIOProvider()):
+        self.sfn_provider = sfn_provider
```
ksugden: It might be helpful to consider what this would look like if we were doing our hexagonal architecture "properly" - We would think of this provider as a "JobRunner" port/interface or some such.  Something that can find and run a job, plus check on its status.  The fact we're using an aws state machine and step function shouldn't really be visible in this layer of the code.  The particular implementation that uses the AWS clients should be part of the implementation of the "JobRunner" interface (or an AWS "adapter" in hexagonal architecture speak)

dbt_platform_helper/domain/job.py
```diff
@@ -0,0 +1,49 @@
...
+from dbt_platform_helper.providers.step_functions import StepFunctions
+
+
+class ScheduledJobExecutionFailedException(PlatformException):
+    pass
+
+
+class JobManager:
+
+    SECONDS = 5
```
ksugden:
```suggestion
    JOB_POLL_SECONDS = 5
```

dbt_platform_helper/providers/step_functions.py
ksugden: thinking now we should just leave this exception handling in, just in case - since we can't safely assume one job over another if there are somehow more than one.

ksugden (approved):
better separation of responsibilities now :)

dbt_platform_helper/providers/step_functions.py
```diff
@@ -0,0 +1,47 @@
...
+            )
+        return response["executionArn"]
+
+    def get_status(self, execution_arn: str):
+        response = self.sfn_client.describe_execution(executionArn=execution_arn)
+        return response["status"]
+
+    def _build_state_machine_arn(self, job_name: str) -> str:
+        region = self.sfn_client.meta.region_name
+        state_machine_name = f"{self.application_name}-{self.env}-{job_name}"
```
alexlescionok: Can the returned name include `-sfn` at the end? The new State Machines will have this suffix to distinguish them from the Copilot-created ones.

dbt_platform_helper/providers/step_functions.py
```diff
@@ -0,0 +1,47 @@
...
+            if error_code == "StateMachineDoesNotExist":
+                raise StateMachineNotFoundException(self.application_name, self.env, job_name)
+            raise StartExecutionFailedException(
+                state_machine_arn, err.response.get("Error", {}).get("Message", str(err))
+            )
+        return response["executionArn"]
+
+    def get_status(self, execution_arn: str):
+        response = self.sfn_client.describe_execution(executionArn=execution_arn)
+        return response["status"]
```
alexlescionok, line 34: Can we also return the time of how long has passed since the last `describe_execution` (like a terraform plan)? I think that will be handy to see for longer-running executions

alexlescionok (dismissed):

dbt_platform_helper/commands/job.py
```diff
@@ -0,0 +1,40 @@
...
+@click.group(cls=ClickDocOptGroup)
+def job():
+    """Commands for managing a scheduled job."""
+
+
+@job.command()
+@click.option("--app", "-a", help="Application name", required=True)
+@click.option("--env", "-e", help="Environment name", required=True)
+@click.option("--name", "-n", help="Name of the scheduled job", required=True)
+@click.option("--follow", is_flag=True)
```
tony-griffin: Does `--follow` need a help field?

dbt_platform_helper/domain/job.py
```diff
@@ -0,0 +1,47 @@
...
+        self.job_runner = job_runner
+        self.io = io
+
+    def start_execution(self, app: str, env: str, name: str, follow: bool):
+
+        self.io.info(f"Beginning execution for job '{name}' in {app}/{env}...")
+        execution_id = self.job_runner.run(name)
+        self.io.info(f"Job started: {execution_id}")
+
+        if follow:
```
tony-griffin, line 26: If follow is true, do we need to ensure we have a cut off point so that it cuts out? Standard step functions can execute for up to 1 year I believe: https://docs.aws.amazon.com/step-functions/latest/dg/choosing-workflow-type.html

dbt_platform_helper/domain/job.py
```diff
@@ -0,0 +1,47 @@
...
+            self.follow_execution(execution_id)
+
+    def follow_execution(self, execution_id: str):
+
+        self.io.info("Waiting for execution to finish...")
+        start_time = time.monotonic()
+
+        while True:
+            time.sleep(self.JOB_POLL_SECONDS)
```
tony-griffin, line 37: Do we want this to fire before the first status check prints out?

dbt_platform_helper/providers/step_functions.py
```diff
@@ -0,0 +1,47 @@
...
+            error_code = err.response.get("Error", {}).get("Code")
+            if error_code == "StateMachineDoesNotExist":
+                raise StateMachineNotFoundException(self.application_name, self.env, job_name)
+            raise StartExecutionFailedException(
+                state_machine_arn, err.response.get("Error", {}).get("Message", str(err))
+            )
+        return response["executionArn"]
+
+    def get_status(self, execution_arn: str):
+        response = self.sfn_client.describe_execution(executionArn=execution_arn)
```
tony-griffin: We handle Client error in run(), but not here, curious as to why not?

dbt_platform_helper/providers/step_functions.py
tony-griffin: we could use the returned startDate to get the exact time the step function started and use this as the value to calculate the elapsed time in follow_execution()

dragos441 (approved):

dbt_platform_helper/domain/job.py
YacquubAdan: Good question. I think we could add something like a polling stops after say, 30 minutes. Even without this cutoff the execution will always start so developers can always check AWS console for the status. Or we could add a --timeout option which would work in a similar way where developers can add time in minutes for the execution to run for? We'd need a better idea of how long typical scheduled jobs run for.

dbt_platform_helper/domain/job.py
YacquubAdan: Its a good idea to get the initial status before waiting - good call

dbt_platform_helper/domain/job.py
alexlescionok: We have set the [timeout](https://github.com/uktrade/platform-tools/blob/bd1fe8ee430a38da41f7aa4bad47f64657cfbf5d/terraform/ecs-service/scheduling/locals.tf#L12) to be 24 hours by default, however, service teams can supply their own value. Perhaps this is sufficient?

ksugden (approved):

PR #1397, opened 2026-04-23
fix: Prevent type error if app has no codebase pipeline repos (off-ticket)

kierdavis:
Noticed this on cyber-tenable.
If no codebase pipelines are declared in platform-config.yml, `local.codebase_pipeline_repos` will be null.
Previously this would lead to null being passed as the first argument to `concat()`, which is invalid. We should pass the empty list instead.

---
### Title:

### Description:

### Tasks:

## Reviewer Checklist


tests/platform_helper/providers/test_terraform_manifest.py
```diff
@@ -248,7 +248,7 @@ def test_generate_environment_config_creates_file(
     assert module["environment"] == env
     assert (
         module["repos"]
-        == "${concat(local.codebase_pipeline_repos != null ? (distinct(values(local.codebase_pipeline_repos))) : null, try([local.config.deploy_repository], []))}"
+        == "${concat(local.codebase_pipeline_repos != null ? (distinct(values(local.codebase_pipeline_repos))) : [], try([local.config.deploy_repository], []))}"
```
adrianclay, line 270: Python asserting that terraform code is character for character identical makes me sad.

adrianclay (approved):

adrianclay:
I think the `terraform/extensions/variables.tf` file could be updated too.

This feels wrong now:
```
default = null # Default to null in case the repository isn't provided in platform-config.yml
```

Could also specify `nullable = false`.

kierdavis:
Updated so that the `repos` var defaults to an empty list and can never be null.

adrianclay (approved):

kierdavis:
Tested a `terraform plan` of the environment terraform for all apps/envs; confirmed this doesn't break anything or introduce any changes we don't expect.

PR #1398, opened 2026-04-23
feat: New Terraform module for tracking the version of deployed infra (DBTP-2542)

dragos441:
Addresses https://uktrade.atlassian.net/browse/DBTP-2542.
Related PR for platform-orchestration https://github.com/uktrade/platform-orchestration/pull/566
Related PR for e2e tests https://github.com/uktrade/platform-orchestration/pull/566

A new Terraform module called `version-tracker` that creates an SSM parameter for each one of the 4 Terraform states that comprise the DBT Platform offering. This is intended to track of the version of AWS resources deployed in the wild.

SSM parameter names have the following naming convetion:

| Path                                                                 | Notes                                                                                  |
|----------------------------------------------------------------------|----------------------------------------------------------------------------------------|
| `platform/version/application/<app>/codebase-pipeline`           | One per application in the non-prod AWS account                                                                   |
| `platform/version/application/<app>/environment-pipeline`        | One per application, in both the non-prod and prod AWS accounts                           |
| `platform/version/application/<app>/environment/<env>`             | One for each environment                                                               |
| `platform/version/application/<app>/environment/<env>/service/<svc>` | One for each service in each environment                                               |

A benefit of this implementation would be the resource dependency on the previous Terraform module (e.g. `environment-pipelines` in this example) that contains all other AWS resources. The previous module has to successfully apply first before the SSM parameter in the `version-tracker` module has its value updated with the new platform version.

```terraform
module "environment-pipelines" {
  application         = "test-app"
  pipeline_name       = each.key
  repository          = "uktrade/test-app-deploy"
  <other params...>,

}

module "version-tracker" {
  platform_version = "10.1.0"
  application = "test-app",
  <other params...>,
  depends_on = [
      module.environment-pipelines
  ]
}
```

SSM parameters also provide a log of who & when updated that parameter. The future goal is to have a Platform Helper command that pulls all this data from multiple SSM parameters into a single output output that shows you all the versions of an application's deployed AWS infra.

---
### Title:
- [X] Conforms to [our pull request title guidance](https://uktrade.atlassian.net/wiki/spaces/DBTP/pages/4402020487/Git+housekeeping#Pull-request-titles). E.g. `feat: Add new feature (DBTP-1234)` or `chore: Correct typo (off-ticket)`

### Description:

### Tasks:

## Reviewer Checklist


dbt_platform_helper/domain/pipelines.py
```diff
@@ -196,3 +211,17 @@ def _generate_terraform_environment_pipeline_manifest(
...
+    def _resolve_platform_version_for_version_tracker(
+        self, pinned_version, platform_config_version
+    ):
+        if pinned_version is not None:
+            return pinned_version
+
+        platform_version_override = self.environment_variable_provider.get(
+            PLATFORM_HELPER_VERSION_OVERRIDE_KEY
+        )
+        if pinned_version is None and platform_version_override is not None:
```
ksugden:
```suggestion
        if platform_version_override is not None:
```

ksugden: or moire pythonic
```suggestion
        if platform_version_override:
```

dbt_platform_helper/providers/terraform_manifest.py
```diff
@@ -88,13 +99,55 @@ def _add_service_module(
...
 
+    def _add_version_tracker_module(
+        self,
+        terraform: dict,
+        terraform_state_type: str,
+        platform_helper_version: str,
+        version_tracker_module_source_override: str = None,
+    ):
+
+        source = (
```
ksugden, line 110: I'm finding the idea of using a different module source for the version-tracker-module than the platform_helper_version and version in the ecs_service_module_source_override difficult to get my head around.  Why would we want to use version X of the tracker module to record that the platform is on version Y? Is the tracker module not part of the platform?  Shouldn't it always be aligned? Would we then need an SSM parameter to track which version of platform the SSM parameters are on? XD. Making my head spin...

dbt_platform_helper/domain/pipelines.py
```diff
@@ -174,17 +183,23 @@ def _generate_terraform_environment_pipeline_manifest(
...
                 "aws_account": aws_account,
-                "module_source": module_source,
+                "env_pipeline_module_source": env_pipeline_module_source,
+                "version_tracker_module_source": version_tracker_module_source,
                 "deploy_branch": deploy_branch,
                 "terraform_version": SUPPORTED_TERRAFORM_VERSION,
                 "aws_provider_version": SUPPORTED_AWS_PROVIDER_VERSION,
                 "deploy_account_id": aws_account_id,
                 "pinned_version": pinned_version,
+                "platform_version": platform_version,
```
ksugden, line 202: I'm just asking myself - do we *really* need yet another `version` parameter?  I suppose the difference between `pinned_version` and `platform_version` is that `pinned_version` only gets populated for managed services... so for demodjango these two parameters could differ.  Our versioning logic is so complex.  My feeling is that these parameters could be combined somehow, but probably for another ticket...

ksugden: maybe passing in a single `platform_version` parameter but also passing in an `is_auto` or `is_managed` flag would be clearer, in a future refactor

dbt_platform_helper/domain/pipelines.py
dragos441: Fair enough, I think I wrote them out in full to use this sort of as a pseudocode but agree that's unnecessary!

dbt_platform_helper/domain/pipelines.py
dragos441: I completely agree. The current logic made my head spin but this PR already became bigger than I originally though and I didn't want to scope creep by starting to refactor. Reassuring to hear though it's not just me who thinks that :)

dbt_platform_helper/providers/terraform_manifest.py
dragos441: You're not wrong, I'll change it so they both use the same source. I think I was really focused on working with the existing code but on second thoughts those 2 parameters can be combined into one

dbt_platform_helper/domain/pipelines.py
ksugden: captured in https://uktrade.atlassian.net/browse/DBTP-2978

dbt_platform_helper/providers/terraform_manifest.py
ksugden: we'll do this as part of a refactor: https://uktrade.atlassian.net/browse/DBTP-2978

ksugden (approved):

PR #1420, opened 2026-04-30
feat: Use new restricted roles for ALB/CDN secret rotation (DBTP-2964)

BenFielder:
Addresses https://uktrade.atlassian.net/browse/DBTP-2964.

## Motivation

There's currently a shared IAM role in the dev & live accounts named `dbt_platform_cloudfront_token_rotation`, which has permission to make changes to any CloudFront distro in the account (dev/live). This role can be assumed by any principal in the platform OUs, which obviously means there is no isolation between apps here.

We've [created new roles](https://github.com/uktrade/platform-public-ingress/blob/main/terraform/modules/iam_lambda_secret_rotate/main.tf) to replace this. There's now a role per application and per environment, which only allows manipulating CloudFront distros associated with that app+env. Also, each new role can only be assumed by the corresponding secret rotation lambda function - not any other principal.

## What this PR does

Instructs the lambda function to use the new roles rather than the old one.

## Testing

* Have verified that the secret rotation lambda still works for a demodjango env
* End-to-end tests are passing.

---
### Title:

### Description:

### Tasks:
  - pull-request-end-to-end-tests execution ID: 845ee4cf-b51d-4d88-aafa-ea83e9ce0f4b

## Reviewer Checklist


terraform/application-load-balancer/main.tf
```diff
@@ -439,6 +439,16 @@ data "aws_iam_policy_document" "origin_verify_rotate_policy" {
...
+     condition {
+      test     = "ForAllValues:StringEquals"
+      variable = "aws:ResourceTag/application"
+      values   = [var.application]
+    }
+    condition {
+      test     = "ForAllValues:StringEquals"
+      variable = "aws:ResourceTag/environment"
+      values   = [var.environment]
+    }
```
BenFielder: These may not be needed, however thought it might be worth introducing further restrictions on what it can/cannot perform cloudfront actions against

kierdavis: I think this whole policy statement is irrelevant actually. Remember this policy belongs to the `APP-ENV-origin-secret-rotate-role` which lives in the application account, so this statement is giving permission to get & update CloudFront distros in the application account - but there aren't any. All CloudFront distros are in the dev/live accounts instead.

Let's leave this policy alone in this PR. We can & probably should remove the cloudfront-related stuff from it, but let's do that in a separate PR.

kierdavis (changes requested):

kierdavis (approved):

PR #1445, opened 2026-05-13
feat: Opensearch User creational and config (DBTP-2960)

a-gleeson:
Addresses https://uktrade.atlassian.net/browse/DBTP-2960 .

---
## Why:
We require to set up connections to opensearch from outwith the account. As we use basic auth auth for opensearch we require to create other users for access management into the OS, otherwise we would be sharing master account details accross accounts. To support this we have added the ability to configure external users and a lambda will create users with reduced permissions and their own details.

## What:

- feat: opensearch `ExternalUserAccessEntry` Pydantic model schema for `platform-config.yml`
- pydantic model wrapper functions to integrate with existing schema
- Refactor: Opensearch Schema now Pyndatic
- Tests: test both schema's with same data ensuring expected results
- Terraform: variable schema updated to expect user config
- Terraform: user creation lambda
- Terraform: Set up domain policy based off config
- Terraform: Invoke Lambda with list of users default and configured users
- Terraform: Unit tests for config
- Terraform: Update Secrets manager endpoint SG to take traffic from lambda 

---
`platform-config.yml`
```
  opensearch:
    type: opensearch
    environments:
      dev:
        external_user_access:
          write-access:
            read: false
            write: true
            cyber_sign_off_by: someone@businessandtrade.gov.uk
```
Params created in SSM:
```
/platform/<app>/<env>/secrets/READ_OPENSEARCH_ENDPOINT (default
/platform/<app>/<env>/secrets/WRITE_OPENSEARCH_ENDPOINT (default)
/platform/<app>/<env>/secrets/WRITE_ACCESS_OPENSEARCH_ENDPOINT (via config)
```
---
### Title:

### Description:

### Tasks:

## Reviewer Checklist


terraform/opensearch/manage_users.py
```diff
@@ -0,0 +1,198 @@
...
+
+        print(f"Creating user {username}...")
+        create_or_update_user(
+            host, admin_user, {"username": username, "password": user_password}, roles
+        )
+
+        parsed = urlparse(host)
+        user_secret_string = f"https://{username}:{user_password}@{parsed.hostname}"
+
+        print(f"Updating user {username} in ssm {secret_name}")
```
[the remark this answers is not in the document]
a-gleeson: Just parameter naming not logging a scret value, changed param name now.

terraform/opensearch/main.tf
```diff
@@ -79,6 +79,15 @@ resource "aws_security_group" "opensearch-security-group" {
     ]
   }
 
+  ingress {
+    description = "Ingress from Lambda Functions to Secrets Manager"
```
adrianclay: This confuses me, are you sure this anything to do with Secrets Manager?

a-gleeson: The security group or code? The code uses a function within secrets manager to generate a password. The account already has an end point set up for rds to call secrets manager so I have reused that.

dbt_platform_helper/entities/platform.py
```diff
@@ -0,0 +1,159 @@
...
+class ExternalUserAccessEntry(BaseModel):
+    read: bool
+    write: bool
+    cyber_sign_off_by: str
+
+    @field_validator("cyber_sign_off_by")
+    @classmethod
+    def valid_cyber_sign_off_by(cls, v: str) -> str:
+        if not re.match(r"^[\w.-]+@(businessandtrade\.gov\.uk|digital\.trade\.gov\.uk)$", v):
+            raise ValueError("must contain a valid DBT email address")
```
adrianclay: Looks more like it needs to _end_ or _be_ a valid DBT email address.

a-gleeson: ah as in the error raised, I'll update that

terraform/opensearch/lambda.tf
```diff
@@ -0,0 +1,146 @@
...
+      "ec2:DescribeNetworkInterfaces",
+      "ec2:DeleteNetworkInterface",
+      "ssm:DeleteParameter",
+      "ssm:PutParameter",
+      "ssm:AddTagsToResource",
+      "kms:Decrypt",
+      "secretsmanager:GetRandomPassword",
+
+    ]
+    resources = ["*"]
```
kierdavis: Can we be more precise about which resources the lambda is allowed to act on here?

terraform/opensearch/lambda.tf
```diff
@@ -0,0 +1,146 @@
...
+  output_path = "${path.module}/manage_users.zip"
+  depends_on = [
+    aws_iam_role.lambda-execution-role
+  ]
+}
+
+resource "aws_lambda_function" "lambda" {
+  # checkov:skip=CKV_AWS_272:Code signing is not currently in use
+  # checkov:skip=CKV_AWS_116:Dead letter queue not required due to the nature of this function
+  filename                       = "${path.module}/manage_users.zip"
```
kierdavis: why not this?

```suggestion
  filename                       = data.archive_file.lambda.output_path
```

This would also resolve the "This file needs to exist, but it's not directly used in the Terraform" comment?

terraform/opensearch/lambda.tf
```diff
@@ -0,0 +1,146 @@
...
+
+resource "aws_iam_role_policy" "lambda-execution-role-policy" {
+  name   = "${var.environment}-${local.name}-execution-policy"
+  role   = aws_iam_role.lambda-execution-role.name
+  policy = data.aws_iam_policy_document.lambda-execution-policy.json
+}
+
+data "aws_security_group" "opensearch-endpoint" {
+  name = local.domain_name
+}
```
kierdavis: This is [the same security group that's created in opensearch/main.tf](https://github.com/uktrade/platform-tools/blob/956585c288228f159085a820f75523b63ccc6a58/terraform/opensearch/main.tf#L65), right?
Trying to read it with a data block during the same terraform run that manages it is probably a bad idea - it'll cause problems next time someone creates a new opensearch instance.
Why not refer to `aws_security_group.opensearch-security-group` directly?

terraform/opensearch/lambda.tf
```diff
@@ -0,0 +1,146 @@
...
+          Index    = v.index,
+          Read     = v.read,
+          Write    = v.write
+        }
+      ]
+    ))
+  })
+
+  depends_on = [
+    aws_lambda_function.lambda,
```
kierdavis: explicit dependency not needed because an implicit dependency already exists via `function_name = aws_lambda_function.lambda.function_name`

terraform/opensearch/main.tf
kierdavis: Is this ingress rule for traffic from lambda to the opensearch endpoint, or to secrets manager?
If it's opensearch, the description is misleading.
If it's secrets manager, shouldn't this rule instead belong to [the security group that contains the secrets manager vpc endpoint](https://github.com/uktrade/platform-tools/blob/956585c288228f159085a820f75523b63ccc6a58/terraform/vpc/main.tf#L244-L254)?

terraform/opensearch/manage_users.py
```diff
@@ -0,0 +1,215 @@
...
+
+    user_secret = None
+
+    try:
+        user_secret = ssm.put_parameter(
+            Name=user_secret_name,
+            Description=user_secret_description,
+            Value=json.dumps(user_secret_string),
+            Tags=[
+                {"Key": "managed-by", "Value": "Terraform"},
```
kierdavis:
```suggestion
                {"Key": "managed-by", "Value": "DBT Platform - opensearch-create-users lambda"},
```

terraform/opensearch/lambda.tf
```diff
@@ -0,0 +1,146 @@
...
+  depends_on = [
+    aws_iam_role.lambda-execution-role
+  ]
+}
+
+resource "aws_lambda_function" "lambda" {
+  # checkov:skip=CKV_AWS_272:Code signing is not currently in use
+  # checkov:skip=CKV_AWS_116:Dead letter queue not required due to the nature of this function
+  filename                       = "${path.module}/manage_users.zip"
+  function_name                  = "${var.environment}-${local.name}-opensearch-create-users"
```
kierdavis: We need `var.application` in here somewhere right?
(same applies to `aws_iam_role.lambda-execution-role`)

dbt_platform_helper/entities/platform.py
```diff
@@ -0,0 +1,237 @@
...
+    urlencode_password: Optional[bool] = Field(
+        default=None,
+        description="""Control whether passwords are URL encoded. Defaults to "true".""",
+    )
+    external_user_access: Optional[dict[str, ExternalUserAccessEntry]] = Field(
+        default=None,
+        description="""Allow extra Users to be defined to be used for access from outwith the account.""",
+    )
+
+    # Explicit validator as pythonn 3.10 does not support dyanmically setting Literals in the type Annotations
```
alexlescionok:
```suggestion
    # Explicit validator as python 3.10 does not support dyanmically setting Literals in the type Annotations
```

terraform/opensearch/lambda.tf
```diff
@@ -0,0 +1,146 @@
...
+  reserved_concurrent_executions = -1
+
+  source_code_hash = data.archive_file.lambda.output_base64sha256
+
+  vpc_config {
+    security_group_ids = [data.aws_security_group.opensearch-endpoint.id, data.aws_security_group.rds-endpoint.id]
+    subnet_ids         = data.aws_subnets.private-subnets.ids
+  }
+
+  tracing_config {
```
alexlescionok: Do we need this?

terraform/opensearch/main.tf
alexlescionok: "Ingress from Lambda Functions to OpenSearch" ?

terraform/opensearch/main.tf
```diff
@@ -105,6 +114,52 @@ resource "random_password" "password" {
   override_special = coalesce(var.config.password_special_characters, "-_!.~$&'()*+,;=")
 }
 
+data "aws_iam_policy_document" "opensearch-policy" {
```
alexlescionok, line 117: Are we certain that no one connects / can connect without TLS?

terraform/opensearch/manage_users.py
```diff
@@ -0,0 +1,215 @@
...
+        )
+
+        role_map = {"Read": READ_ROLE, "Write": WRITE_ROLE, "Index": INDEX_ROLE}
+
+        roles = [role_map[key] for key in role_map.keys() if user_config[key]]
+
+        print("generating password ...")
+        user_password = secrets_manager.get_random_password(
+            PasswordLength=16,
+            ExcludeCharacters='[]{}()"@/\\;=?&`><:|#',
```
alexlescionok: Reference OpenSearch config values

tests/platform_helper/fixtures/expected_data/platform/config/extensions/opensearch_bad_data.yml
```diff
@@ -0,0 +1,70 @@
...
+   pydantic: "environments.*dev.*external_user_access.*user-write-access.write.*Input should be a valid boolean"
+my-opensearch-bad-user-cyber: 
+   schema: "environments.*dev.*external_user_access.*user-write-access.cyber_sign_off.*must be a valid DBT email address"
+   pydantic: "environments.*dev.*external_user_access.*user-write-access.cyber_sign_off.*must be a valid DBT email address"
+my-opensearch-invalid-plan: 
+   schema: "environments.*dev.*plan.*does not match 'invalid'"
+   pydantic: "environments.*dev.*plan.*Plan must be one of"
+my-opensearch-bad-user-index:
+   schema: "environments.*dev.*external_user_access.*index.*should be a valid boolean"
+   pydantic: "environments.*dev.*external_user_access.*index.*should be a valid boolean"
```
alexlescionok:
```suggestion
   pydantic: "environments.*dev.*external_user_access.*index.*should be a valid boolean"

```
(To get rid of the red '-' symbol)

tests/platform_helper/entities/test_extensions_config.py
```diff
@@ -0,0 +1,72 @@
...
+        )
+
+        return PlatformConfigSchema.extension_schemas()[extension].validate
+    if param == "pydantic":
+        from dbt_platform_helper.entities.platform import validate_extension_fn
+
+        return validate_extension_fn(extension)
+
+
+def make_validator_fixture(extension, mode):
```
alexlescionok: We can just refer to `_get_validator()`

terraform/opensearch/main.tf
a-gleeson: Yes as the SG only allows traffic over `443`

terraform/opensearch/main.tf
a-gleeson: I can remove this, due to the way I set up I think it works without it

terraform/opensearch/lambda.tf
a-gleeson: updated

terraform/environment-pipelines/iam.tf
```diff
@@ -526,9 +526,10 @@ data "aws_iam_policy_document" "postgres" {
       "lambda:TagResource",
       "lambda:PutFunctionConcurrency"
     ]
-    resources = [for env in local.environment_config :
-      "arn:aws:lambda:${local.account_region}:function:${var.application}-${env.name}-*"
-    ]
+    resources = flatten([for env in local.environment_config : [
+      "arn:aws:lambda:${local.account_region}:function:${var.application}-${env.name}-*",
+      "arn:aws:lambda:${local.account_region}:function:${env.name}-${var.application}-*" # Openserch naming is flipped
```
kierdavis: now that we've change the function name to include var.application, I don't think you need this line anymore

kierdavis (approved):

PR #1476, opened 2026-05-21
fix: CodeBase image builds failing to pull code in rare circumstances (DBTP-2763)

adrianclay:
Reverts uktrade/platform-tools#1466 which reverts uktrade/platform-tools#1446

Manually specify the CodeConnection

This forces all Pipelines and CodeBuilds to use a single consistent codeconnection to pull from.

Currently there is a hidden default option that CodeBuild will use instead if one isn't set explicitly.

Includes a fix for the "User is not authorized to access connection" error we were seeing in the orchestration pipeline last time we tried to release this change.

Signed-off-by: DBT pre-commit check

Addresses https://uktrade.atlassian.net/browse/DBTP-2763


terraform/codebase-pipelines/iam.tf
```diff
@@ -156,7 +156,9 @@ data "aws_iam_policy_document" "codestar_connection_access" {
     effect = "Allow"
     actions = [
       "codestar-connections:GetConnectionToken",
-      "codestar-connections:UseConnection"
+      "codestar-connections:UseConnection",
+      "codeconnections:GetConnectionToken",
+      "codeconnections:GetConnection",
```
adrianclay: According to the support ticket, this is _probably_ the reason @ksugden / @kierdavis

kierdavis:
Tested by [running the platform upgrade pipeline (test tier only) using this branch](https://eu-west-2.console.aws.amazon.com/codesuite/codepipeline/pipelines/platform-upgrade-prod/executions/c8411e2a-d0a5-4247-a303-0c4b498ff4ad/visualization?region=eu-west-2), and it solved the issue we saw last week ("User is not authorized to access connection").

A caveat is that if both the IAM policy and the codebuild project are updated in the same terraform run, the new IAM policy might not have finished replicating out by the time the UpdateProject call is made. This manifests as the "User is not authorized to access connection" error - but it works fine on the second attempt. Not sure if there's value in e.g. adding a terraform `time_sleep` to accommodate for this, or to just let it be.

dragos441 (approved):

PR #1500, opened 2026-06-08
feat: Add ability to disable AWS CodePipelines (DBTP-3072)

dragos441:
Addresses https://uktrade.atlassian.net/browse/DBTP-3072.

Allows disabling AWS CodePipelines at the codebase level via the new `pipeline_mode` property in `platform-config.yml`. The idea behind this is that once you move your ECS deployments to GitHub Actions, there's no point keeping the AWS CodePipelines too.

`pipeline_mode` support 3 values:

- `aws_codepipeline` (default) continue using AWS CodePipeline, no GitHub Actions
- `dual_codepipeline_github` middle ground where GitHub Actions are present, and CodePipelines are still present but with the automatic image build web hook triggers disabled
- `github_actions` CodePipelines and associated resources are deleted, only GitHub Actions are used for deployments

> [!NOTE]  
> Validation has been added to prevent removing AWS CodePipelines if custom pre/post deploy actions are defined. This is because we currently do not provide a like-for-like replacement for them in GitHub Actions. Ticket https://uktrade.atlassian.net/browse/DBTP-3132 raised to address that.

---
### Title:

### Description:

### Tasks:

## Reviewer Checklist


terraform/codebase-pipelines/custom-codebuild.tf
tony-griffin, line 1: Is there any chance these get created then orphaned if they don't get gated off via a codepipeline_enabled for_each?
As in they get created but never invoked?

```
for_each       = toset(local.codepipeline_enabled && local.has_custom_pre_deploy ? [""] : [])
```

dbt_platform_helper/entities/platform_config_schema.py
```diff
@@ -146,7 +146,7 @@ def __codebase_pipelines_schema() -> dict:
                 "repository": str,
                 Optional("slack_channel"): str,
                 Optional("requires_image_build"): bool,
-                Optional("use_github_actions"): bool,
+                Optional("pipeline_mode"): str,
```
tony-griffin, line 149: Should we enforce/validate the expected strings here like we do in the Terraform code? A typo here might pass schema validation and fail later in the pipleline

Something like:

```
Optional("pipeline_mode"): Or(
                      "aws_codepipeline", "dual_codepipeline_github", "github_actions"
                  ),
```

tony-griffin: is there a chance that if the **use_github_actions** field is silently dropped, meaning a service that thought it had GitHub Actions enabled could revert to CodePipeline behaviour unnoticed?

terraform/codebase-pipelines/custom-codebuild.tf
dragos441: Ah you are right, good spot! There currently isn't an alternative in GHA for these custom pre/post deploy pipeline actions ...

If you have them, you basically can't get off AWS CodePipeline. Could either add a like-for-like replacement in GHA, or add validation to prevent them being orphaned.

dbt_platform_helper/entities/platform_config_schema.py
dragos441: Only 1 test tier service mentions `use_github_actions`, was going to address that plus any BYOD repos we've migrated so far once this is released.

terraform/codebase-pipelines/custom-codebuild.tf
dragos441: So that I don't scope creep myself I think I'll add some logic to avoid removing CodePipelines if pre/post deploy  jobs are present ... even if `pipeline_mode` is set to `github_actions`.  

And I'll write up a new ticket to decide what to do about pre/post deploy in GHA.

tony-griffin (approved):

tony-griffin (approved):

PR #1504, opened 2026-06-09
feat: Switch the source of truth for cdn_domain_list (DBTP-2701)

andiempett:
Addresses https://uktrade.atlassian.net/browse/DBTP-2701.
linked to https://github.com/uktrade/platform-tools/tree/DBTP-2702-deprecate-cdn-options.

- switch to using the [parameter populated by platform-public-ingress](https://github.com/uktrade/platform-public-ingress/blob/886bceda9a362aa3f80d45c4add9ac1e4e692c75/terraform/main.tf#L88) for the list of cdns
- remove the option to disable cdn secret, secret rotation lambda and associated resources.

This change ensures that an ALB is set up with The WAF header rule and the secrets rotation lambda.
We have checked there is only [one use case where an ALB is not protected in this way.](https://github.com/uktrade/platform-documentation-deploy/blob/72af0346e34b3e24e1c9a4322e27a441c6e2fd98/platform-config.yml#L66)
This change enforces the protection with no option to have an 'open' ALB

---
### Title:

### Description:

### Tasks:

## Reviewer Checklist


kierdavis (commented):
Looks good so far!

Some other things I think we should do:

- [x] Remove [`cdn_domains_list`](https://github.com/uktrade/platform-tools/blob/e3157e51b3f7b650d53db069ef4e524835603d7c/terraform/application-load-balancer/variables.tf#L20) and [its associated validation](https://github.com/uktrade/platform-tools/blob/e3157e51b3f7b650d53db069ef4e524835603d7c/terraform/application-load-balancer/variables.tf#L27-L32) from variables.tf.
- [x] Remove references to the `cdn_domains_list` variable from the unit tests.
- [x] Remove reference to `cdn_domains_list` from the [managed_ingress validation](https://github.com/uktrade/platform-tools/blob/e3157e51b3f7b650d53db069ef4e524835603d7c/terraform/extensions/main.tf#L177). We can just delete this line, so that the logic becomes "assert that all alb extensions have managed_ingress: true" (regardless of whether they have cdn_domains_list or not).
- [x] Get feedback from wider platform team about the decision to enable the origin verify secret for all ALBs even if they don't have CDNs - let's make sure we drop this PR into dbt-platform-pull-requests

kierdavis (approved):

a-gleeson (commented):
Changes look good.

Just considering a few things, if my understanding is correct.

- We will always set up the secrets variable
- We will always set up the rotation lambda
- This will run every 7 days

I am wondering if:
- What happens if there's no cdn set up for the account?
- Will there be a failing lambda
- Is it ideal to have this little cost when there's no cdn set up for the account or not needed

I guess there won't actually be a service without a cdn and alb set up right now, although it is not impossible

andiempett:
> Changes look good.
> 
> Just considering a few things, if my understanding is correct.
> 
> * We will always set up the secrets variable
> * We will always set up the rotation lambda
> * This will run every 7 days
> 
> I am wondering if:
> 
> * What happens if there's no cdn set up for the account?
> * Will there be a failing lambda
> * Is it ideal to have this little cost when there's no cdn set up for the account or not needed
> 
> I guess there won't actually be a service without a cdn and alb set up right now, although it is not impossible

@a-gleeson @kierdavis 
changed the behaviour: 
if no `cdn_domains_list`, don't create the lambda and secret etc. 70fa99ca68e4520d1bc02633b95cffe01c15054e

kierdavis (approved):

PR #1525, opened 2026-06-26
feat: Privatelink Service Infra (DBTP-2925)

a-gleeson:
Addresses https://uktrade.atlassian.net/browse/DBTP-2925.
---
## What:
- Adds `Load Balanced Internal Service`
- Sets up target group
- Sets up NLB listerner
- Validates pending certs for endpoint domain
---
### Title:

### Description:

### Tasks:

## Reviewer Checklist

- [x] I have reviewed the PR and ensured no secret values are present


terraform/ecs-service/main.tf
```diff
@@ -161,7 +171,10 @@ resource "aws_ecs_service" "service" {
     ignore_changes = [task_definition, desired_count] # See reasoning for this above
   }
 
-  depends_on = [aws_lambda_invocation.dummy_listener_rule]
+  depends_on = [
+    aws_lambda_invocation.dummy_listener_rule,
+    aws_lb_listener.nlb
```
dragos441, line 176: Should only depend on `aws_lb_listener.nlb` when it's a load balanced internal service, otherwise might error when trying to reference a resource that doesn't exist

terraform/ecs-service/main.tf
```diff
@@ -482,3 +495,69 @@ module "scheduling" {
   tags                = local.tags
   log_group_arn       = aws_cloudwatch_log_group.ecs_service_logs.arn
 }
+
+data "aws_acm_certificate" "acm" {
+  count = local.internal_service_required == 1 ? length(var.service_config.http.alias) : 0
+
+  domain   = var.service_config.http.alias[count.index]
```
dragos441: Might be cleaner to use `for_each` instead of `count` so that Terraform tracks each instance by the alias rather than a numeric index.

terraform/ecs-service/main.tf
```diff
@@ -482,3 +495,69 @@ module "scheduling" {
...
+
+resource "aws_lb_listener" "nlb" {
+  for_each = local.internal_service_required == 1 ? {
+    for idx, cert in data.aws_acm_certificate.acm : cert.domain => cert
+  } : {}
+
+  load_balancer_arn = data.aws_lb.nlb[0].arn
+  protocol          = "TLS"
+  port              = 443
+  ssl_policy        = "ELBSecurityPolicy-TLS13-1-2-Res-PQ-2025-09"
```
dragos441, line 549: Post-quantum, very nice! 

However, that policy only supports TLS 1.2 and 1.3, which means the clients connecting to NLB also need to support those versions.

terraform/ecs-service/main.tf
```diff
@@ -482,3 +495,69 @@ module "scheduling" {
...
+    interval            = var.service_config.http.healthcheck.interval
+    timeout             = var.service_config.http.healthcheck.timeout
+  }
+
+}
+
+resource "aws_lb_listener" "nlb" {
+  for_each = local.internal_service_required == 1 ? {
+    for idx, cert in data.aws_acm_certificate.acm : cert.domain => cert
+  } : {}
```
dragos441, line 544: Should this be one listener on port 443 with multiple certs attached, rather than one listener per cert? 

If multiple services run this separately, would they all try to create listeners on the same NLB port 443 and collide?

terraform/privatelink/producer/main.tf
```diff
@@ -0,0 +1,138 @@
+data "aws_vpc" "vpc" {
```
dragos441, line 1: Is the entire `producer` module duplicated from [your other PR](https://github.com/uktrade/platform-tools/pull/1524/changes#diff-70eb89097de2af5e3654f5796f6fb82621dc7de962135268b2681be13acc6b72)?

terraform/ecs-service/main.tf
a-gleeson: This wont error when it doesn't exist. I will test it too though. Research does indicate it will be fine although I could be wrong still :D

terraform/ecs-service/main.tf
a-gleeson: do I have to map the list into a map for a for_each to work?

terraform/ecs-service/main.tf
a-gleeson: That's fair, I think for now we leave it and enforce it. So the connecting services will require it.

terraform/ecs-service/main.tf
a-gleeson: You are correct, each port can only have one listener. So support does not exist for this currently. 
I was trying to kind of add support for multiple alias's or services through privatelink. 

Currently you can support multiple services by setting up a unique privatelink for each but to add support and stop this from erroring on multiple alia's we would need to create each alias on a different listener. This is because NLB does not do path based routing

terraform/extensions/main.tf
```diff
@@ -168,3 +168,10 @@ resource "aws_ssm_parameter" "environment_data" {
 data "aws_ssm_parameter" "log_destination_arn" {
   name = "/copilot/tools/central_log_groups"
 }
+
+module "privatelink-accept-certs" {
```
dragos441: Would there be any way to place this under a `for_each` like the others rather than to run that module every time?

terraform/ecs-service/main.tf
dragos441: Looks good, having one NLB per internal service and only allowing one `http.alias` should mitigate this.

terraform/extensions/main.tf
a-gleeson: yes, I thought it would be better to live with the `pirvatelink` modules

terraform/ecs-service/main.tf
```diff
@@ -484,3 +497,61 @@ module "scheduling" {
...
+
+resource "aws_lb_target_group" "nlb_to_ecs" {
+  count = local.internal_service_required
+
+  name                 = substr("${var.service_config.name}-tg-${random_string.tg_suffix[count.index].result}", 0, 32)
+  port                 = 443
+  protocol             = "TLS"
+  target_type          = "ip"
+  vpc_id               = data.aws_vpc.vpc.id
+  deregistration_delay = 30
```
dragos441: Without hardcoding?

```suggestion
  deregistration_delay = var.service_config.http.deregistration_delay
```

terraform/extensions/main.tf
dragos441: The module’s intent is not very obvious, it appears always required, but only applies to environments with an internal service.

How would you feel about calling the `privatelink/accept-certificates` module from `ecs-cluster`? I think that would make the intent clearer.

a-gleeson: It's specific for services so maybe that makes sense

dragos441 (approved):

PR #1582, opened 2026-07-27
feat: Upgrade all remaining Redis caches to Valkey (DBTP-3081)

ddwilliams74:
Addresses https://uktrade.atlassian.net/browse/DBTP-3081.

---
### Title:

### Description:

### Tasks:

## Reviewer Checklist

- [x] I have reviewed the PR and ensured no secret values are present

### DIFF
--- a/terraform/elasticache-redis/locals.tf
+++ b/terraform/elasticache-redis/locals.tf
@@ -14,10 +14,8 @@ locals {
     copilot-environment = var.environment
   }
 
-  # Don't switch to Valkey for great prod account
-  switch_to_valkey = !(var.application == "great" && (var.environment == "hotfix" || var.environment == "prod"))
-  engine           = var.config.engine == "7.1" && local.switch_to_valkey ? "valkey" : "redis"
-  engine_version   = var.config.engine == "7.1" && local.switch_to_valkey ? "7.2" : var.config.engine
+  engine           = (var.config.engine == "7.1" || var.config.engine == "7.2") ? "valkey" : "redis"
+  engine_version   = (var.config.engine == "7.1" || var.config.engine == "7.2") ? "7.2" : var.config.engine
 
   central_log_group_arns        = jsondecode(data.aws_ssm_parameter.log-destination-arn.value)
   central_log_group_destination = var.environment == "prod" ? local.central_log_group_arns["prod"] : local.central_log_group_arns["dev"]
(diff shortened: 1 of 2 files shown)


terraform/elasticache-redis/tests/unit.tftest.hcl
```diff
@@ -296,14 +296,15 @@ run "aws_elasticache_replication_group_unit_7_1_prod" {
   }
 }
 
-run "aws_elasticache_replication_group_unit_7_1_dev_great" {
+# Ensure that setting engine_version to 7.2 is respected in dev
+run "aws_elasticache_replication_group_unit_7_2_dev" {
```
adrianclay, line 300: Minor feedback would be that the code doesn't care about the dev/prod split anymore. I like having the test to show that 7.2 works, but having both is unnecessary.

terraform/elasticache-redis/locals.tf
```diff
@@ -14,10 +14,8 @@ locals {
     copilot-environment = var.environment
   }
 
-  # Don't switch to Valkey for great prod account
-  switch_to_valkey = !(var.application == "great" && (var.environment == "hotfix" || var.environment == "prod"))
-  engine           = var.config.engine == "7.1" && local.switch_to_valkey ? "valkey" : "redis"
-  engine_version   = var.config.engine == "7.1" && local.switch_to_valkey ? "7.2" : var.config.engine
+  engine           = (var.config.engine == "7.1" || var.config.engine == "7.2") ? "valkey" : "redis"
+  engine_version   = (var.config.engine == "7.1" || var.config.engine == "7.2") ? "7.2" : var.config.engine
```
adrianclay, line 18: I hope that we have a plan for what we're doing about future Valkey versions. Would hate to see this ticket closed without doing the cleanup of this debt, however this issue isn't a blocker to the merging of this PR.

I see two options:

- Platform takes responsibility for the version, and teams can no longer specify the version in their config
- Teams can specify the version used, and we go round all platform-config.yml files and update them to 7.2

kierdavis (approved):

terraform/elasticache-redis/locals.tf
ddwilliams74: Agree. I've been thinking about this, particularly around how we may want to rollout in the future. 

We could include non-prod and prod variables that we'd just change e.g., 7.2 -> 7.3, if we wanted to retain control. We should then remove the version from platform-config. Alternatively as you say, allow service teams to control their version but push up the lower limit over time.

terraform/elasticache-redis/tests/unit.tftest.hcl
ddwilliams74: We'd probably want the split in there for future upgrades though where we'd presumably do non-prod first. You could definitely argue that the tests should match the current config in locals though, and we don't have the non-prod/prod distinction in there at the moment.

